PR #4, opened 2015-05-16
Update JUnit to 4.12

mkordas:


adamdubiel:
Pull requests are welcome, but please put some more effort into them - bit more meaningful code change. I see nothing in 4.12 release that would be critical for us.

PR #12, opened 2015-05-18
Change in docker management

kdebski:
Change in docker management configuration to support multiple kafka clusters.


adamdubiel:
This will be fixed in #20, closing this PR.

Issue #10, opened 2015-05-18
Additional metrics for producer with ack=ALL


pawelszymczyk:
Hermes uses same metric to meter latency for producer with ack=-1 and ack=1. Due to that we have one chart with latency for 99 precentile which show us always value 65ms. We could visualize latency more precisely by separating latency metric for both producers.  

AC:
* additional metric for producer latency with ack=-1
* two charts with latency. First for ack = 1, second for ack=-1

PR #23, opened 2015-05-19
#10 new metrics for ack producers

pawelszymczyk:
Additional metrics and major refactor.


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/PathsCompiler.java
```diff
@@ -0,0 +1,37 @@
+package pl.allegro.tech.hermes.common.metric;
+
+import org.apache.commons.lang.text.StrBuilder;
+
+public class PathsCompiler {
+
+    public static final String $HOSTNAME = "$hostname";
```
adamdubiel: i don't like the syntax -> can we rmove $ from constant names?

adamdubiel:
Just one remark, other than that :+1:

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/PathsCompiler.java
pawelszymczyk: sure

Issue #31, opened 2015-05-27
support for custom Jersey package scanning


adamdubiel:
As Hermes user i want to be able to specify custom packages that Jersey will scan on startup (i.e. to attach custom filters).

PR #32, opened 2015-05-27
fixing #31 - management custom package scanning

adamdubiel:


adamdubiel:
New configuration in management `application.yaml`:

```
jersey:
  packagesToScan:
    - com.my.package
```

Issue #38, opened 2015-05-29
Thread pools queues monitoring


rzukow:
When using thread pool it is always difficult to set correct number of threads.

To make it easier we want to measure a time each task is waiting for execution in thread pool, and store it together with other metrics in graphite.

First pool we want to measure this way is jetty pool in consumers module.

PR #61, opened 2015-06-08
#38 allowing using http client from hk2 context when adding message s…

rzukow:
…ender providers

Just to discuss possible solution...


adamdubiel:
:+1:

pawelszymczyk:
:+1:

PR #86, opened 2015-06-17
Alternative latency timer fix

rzukow:
Possible omptimisation and fix for timer.
This is an alternative and possibly better version of my previous PR.
What do you think?

[Review on Reviewable](https://reviewable.io/reviews/allegro/hermes/86)


pawelszymczyk:
this PR should be merged into rc branch

PR #96, opened 2015-06-24
Fix int tests

pawelszymczyk:
First approach to run integration tests on Travis


pawelszymczyk:
need  additional fixes

Issue #72, opened 2015-06-11
Hermes support for publishing events using HTTP/2


wojtkiewicz:
As a hermes user I want to be able to publish events using http2 protocol.

PR #109, opened 2015-06-29
#72 separate ssl/http2 switch, secure random fix

dankraw:
+ separate SSL switch from HTTP/2 
+ moved to faster (system dependent) SecureRandom implementation


rzukow:
:+1:

adamdubiel:
:+1:

PR #116, opened 2015-07-03
docker build fix

dankraw:
Fixes docker images build by conditionally applying gradle application/distribution plugin.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -2,7 +2,7 @@
 
 rm -rf build
 mkdir -p build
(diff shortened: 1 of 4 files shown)


docker/build.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+export HERMES_DISTRIBUTION=1
```
adamdubiel: change it to gradle argument, why play iwht sys variables when flags are available in build system?

Issue #128, opened 2015-07-20
Running listeners from kafka-network-thread


wojtkiewicz:
Listeners should be run from undertow worker pool.

PR #129, opened 2015-07-20
#128 thread jump from kafka-network-thread to x-nio worker pool

wojtkiewicz:
#128 thread jump from kafka-network-thread to x-nio worker pool for listener execution


adamdubiel:
:+1:

pawelszymczyk:
:+1:

PR #132, opened 2015-07-23
consumer 4xx output rate fix

dankraw:
Fix for: consumers should treat 4xx response from "subscriptions without 4xx retry" as success


dankraw:
wrong target branch

PR #157, opened 2015-08-05
156 consumer sender backoff

pawelszymczyk:


adamdubiel:
I guess metrics we talked about (inflight lifetime) will be added in separate PR?

PR #165, opened 2015-08-11
156 consumer sender backoff

pawelszymczyk:


hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
```diff
@@ -9,21 +9,29 @@
...
 
     @JsonCreator
-    public SubscriptionPolicy(@JsonProperty("rate") int rate, @JsonProperty("messageTtl") int messageTtl,  @JsonProperty("retryClientErrors") boolean retryClientErrors) {
+    public SubscriptionPolicy(@JsonProperty("rate") int rate, @JsonProperty("messageTtl") int messageTtl,
+                              @JsonProperty("retryClientErrors") boolean retryClientErrors,
+                              @JsonProperty("messageBackoff") Integer messageBackoff) {
         this.rate = rate;
         this.messageTtl = messageTtl;
         this.retryClientErrors = retryClientErrors;
+        this.messageBackoff = messageBackoff != null ? messageBackoff : DEFAULT_MESSAGE_BACKOFF;
```
druminski: default values are set in the builder. Why this one is an exception?

pawelszymczyk: After some discuss we decided to try tihis approach

pjagielski: We read this from zookeeper and it's needed here for backwards compatibility

PR #169, opened 2015-08-11
168 event schema metadata

druminski:


pbetkier:
JsonMessageContentWrapper still converts the message into {"metadata": ..., "message": ...} structure instead of adding __metadata map to the existing structure like AvroMessageContentWrapper now does. We should keep the wrapping structure consistent across formats.

Other than that, nice code :)

adamdubiel:
I don't think we should change existing structure, at least not without some in-depth analysis of potential side effects (like how to deploy new version without breaking compatibility).

pbetkier:
Good point, let's have such analysis and make a migration plan. I just want to make sure the structures in these formats converge at some point in the near future. I believe that's what our users (i.e. hermes adopters) would expect, not caring about the story of hermes deployment at allegro.

pbetkier:
:+1:

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageContentWrapperDispatcher.java
```diff
@@ -0,0 +1,44 @@
+package pl.allegro.tech.hermes.common.message.wrapper;
+
+import org.apache.avro.Schema;
+import pl.allegro.tech.hermes.api.Topic;
+import pl.allegro.tech.hermes.domain.topic.schema.SchemaRepository;
+
+import javax.inject.Inject;
+
+public class MessageContentWrapperDispatcher {
```
pawelszymczyk: this class does not dispatch any MessageContenetWrapper, why it isn't MessageContentWrapper?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageMetadata.java
```diff
@@ -7,11 +7,11 @@
 
 public class MessageMetadata {
 
-    private final long timestamp;
+    private final Long timestamp;
```
pawelszymczyk: it's necessary to enter this change?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/metadata/MetadataAddingMessageConverter.java
```diff
@@ -0,0 +1,23 @@
+package pl.allegro.tech.hermes.frontend.publishing.metadata;
+
+import pl.allegro.tech.hermes.api.Topic;
+import pl.allegro.tech.hermes.common.message.wrapper.MessageContentWrapperDispatcher;
+import pl.allegro.tech.hermes.frontend.publishing.message.Message;
+
+import javax.inject.Inject;
+
+public class MetadataAddingMessageConverter {
```
pawelszymczyk: in my opinion this class is redundant, over engineered

hermes-test-helper/src/main/java/pl/allegro/tech/hermes/test/helper/avro/RecordToBytesConverter.java
```diff
@@ -13,9 +13,9 @@
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 
-public class RecordToBytesConverter {
+public interface RecordToBytesConverter {
```
pawelszymczyk: AvroRecordToBytesConverter is very similiar with this class

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageContentWrapperDispatcher.java
druminski: ok, class renamed

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageMetadata.java
druminski: it's not necessary, change reverted

pawelszymczyk:
:+1:

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/metadata/MetadataAddingMessageConverter.java
druminski: We need to invoke this code, question is where? I don't want to move it to PublishingServlet so I decided to not change it.

hermes-test-helper/src/main/java/pl/allegro/tech/hermes/test/helper/avro/RecordToBytesConverter.java
druminski: Yes, they are the same but in different scopes so we need small duplication.

PR #175, opened 2015-08-19
Fix for always-passing awaitility assertions in Waiter

pbetkier:
Fixes such assertions:
```java
(…).until(() -> {
	x.equals(y);
});
```

Such lambdas were treated as a `Runnable`, so since no exception was thrown the assertions did always pass. I replaced such lambdas with `() -> x.equals(y)`. Also simplified lambdas of this sort: `() -> { return x.equals(y); }` for brevity.

Also, did some minor refactorings for code reuse and removed `@Unreliable` annotations, since the tests do pass on Travis right now and assertions are now reliable :)


adamdubiel:
:+1:

druminski:
:+1:

Issue #177, opened 2015-08-20
Timeout for graphite client


druminski:
Today our hermes-management process stopped responding after 8 days of uptime. It turned out that all of worker threads where waiting for response from graphite. We should add some timeout for graphite client.

PR #178, opened 2015-08-20
Configuring timeout for http client in hermes-management. Issue #177

druminski:


pjagielski:
:+1:

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/graphite/GraphiteClient.java
```diff
@@ -32,11 +32,10 @@ public GraphiteClient(WebTarget webTarget) {
     public GraphiteMetrics readMetrics(String... metrics) {
         try {
             GraphiteMetrics response = new GraphiteMetrics();
-            queryGraphite(metrics).stream()
-                    .forEach(metric -> response.addMetricValue(metric.getTarget(), getFirstValue(metric)));
+            queryGraphite(metrics).stream().forEach(metric -> response.addMetricValue(metric.getTarget(), getFirstValue(metric)));
             return response;
         } catch (GraphiteConnectionException exception) {
-            logger.warn("Unable to read from Graphite", exception);
+            logger.warn("Unable to read from Graphite", exception.getMessage());
```
pbetkier: Why do we remove stacktrace information from our logs? GraphiteConnectionException messages are not very informative, e.g. "Error connecting to Graphite". Now we won't know the cause: timeout, corrupted data, host unreachable or something else?

druminski: I decided to remove whole stack trace because it's spamming our logs lately ;) But you have right that last exception cause message doesn't say much, so I decided to log root cause message which always tells you something informative. Right now we will be see something like this:
```
p.a.t.h.m.i.graphite.GraphiteClient - Unable to read from Graphite. Connection refused
```

pbetkier: Ok, fair enough :)

Issue #99, opened 2015-06-24
When creating hermes Consumer fails, kafka Consumer is still active


pjagielski:
ConsumersSupervisor should close kafka Consumer when creating hermes Consumer fails.

PR #190, opened 2015-08-25
Issue #99. Handling consumers exceptions.

druminski:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Consumer.java
```diff
@@ -36,6 +37,7 @@
     private Subscription subscription;
 
     private volatile boolean consuming = true;
+    private volatile CountDownLatch stoppedLatch = new CountDownLatch(1);
```
pbetkier: Volatile keyword is not needed here, since you never reassign stoppedLatch reference and CountDownLatch.countDown() operation is visible to other threads, as it *happens-before* CountDownLatch.await(). You may make this field final by the way.

Source: http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/SubscriptionLocks.java
```diff
@@ -5,27 +5,44 @@
...
 
-    public SubscriptionLock getLock(Subscription subscription) {
-        return getLock(subscription.getId());
+    public CloseableSubscriptionLock lock(Subscription subscription) {
+        return getLock(subscription.getId()).autoLock();
     }
 }
 
-final class SubscriptionLock {
+final class CloseableSubscriptionLock extends ReentrantLock implements AutoCloseable {
```
pbetkier: I like it!

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Consumer.java
druminski: Good point, done

pbetkier:
:+1:

dankraw:
:+1:

PR #198, opened 2015-08-27
196 rc broker metrics actual ids

wojtkiewicz:
Added latency-per-broker metrics that are based on actual cluster information from working producer.

**Warning**
There is a couple of issues with this implementation: 
- there should be a way to retrieve producer-node-metrics without knowing exact node-ids
- producer metadata should be accessible for the user (I should not be forced to use reflection API to do that)
- there should be a way to register listener that triggers when actual producer metadata changes

I will prepare PR to KafkaProducer as soon as possible to make such functionality easier to implement.


pjagielski:
:+1:

druminski:
:+1:

PR #209, opened 2015-09-17
204 initial version of new consumer runtime model

wojtkiewicz:
Initial version of new consumer runtime model. Designed to be extended by introducing more advanced workload spread algorithms.

Two basic implementations are provided:
- **legacy.mirror** - consumers are mirroring each other without using new runtime model (legacy behaviour)
- **mirror** - consumers are mirroring each other using new runtime model, the difference between *legacy.mirror* is another level of indirection between change in *SubscriptionCache* and spawning new consumers

New runtime model is based on a structure in zk reflecting which consumer supervisor is assigned to which subscription. Structure looks as follows:

/hermes/consumers/**runtime**/*qualified_topic_name$subscription_name*/*(supervisor_id)+*

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionName.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionName.java
@@ -1,5 +1,7 @@
 package pl.allegro.tech.hermes.api;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 public class SubscriptionName {
 
     private String name;
@@ -19,5 +21,16 @@ public class SubscriptionName {
 
     public TopicName getTopicName() {
         return topicName;
+    }
+
+    public static SubscriptionName fromString(String string) {
+        String[] tokens = string.split("\\$");
+        checkArgument(tokens.length > 1, "Incorrect string format. Expected 'topic$subscription'. Found:'%s'", string);
+        return new SubscriptionName(tokens[1], TopicName.fromQualifiedName(tokens[0]));
+    }
+
+    @Override
+    public String toString() {
+        return topicName.qualifiedName() + "$" + name;
     }
 }
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
@@ -1,5 +1,10 @@
 package pl.allegro.tech.hermes.common.config;
 
+import pl.allegro.tech.hermes.common.util.InetAddressHostnameResolver;
+
+import static java.lang.Math.abs;
+import static java.util.UUID.randomUUID;
+
 @SuppressWarnings("PMD.AvoidUsingHardCodedIP")
 public enum Configs {
 
@@ -101,6 +106,9 @@ public enum Configs {
     CONSUMER_OFFSET_COMMIT_QUEUE_ALERT_MINIMAL_IDLE_PERIOD("consumer.offset.commit.queue.alert.minimal.idle.period", 3600),
     CONSUMER_OFFSET_COMMIT_QUEUE_ALERT_SIZE("consumer.offset.commit.queue.alert.size", 20_000),
     CONSUMER_HEALTH_CHECK_PORT("consumer.status.health.port", 8000),
+    CONSUMER_WORKLOAD_ALGORITHM("consumer.workload.algorithm", "legacy.mirror"),
+    CONSUMER_WORKLOAD_ID("consumer.workload.id",
+            new InetAddressHostnameResolver().resolve().replaceAll("\\.", "_") + "$" + abs(randomUUID().getMostSignificantBits())),
 
     GRAPHITE_HOST("graphite.host", "localhost"),
     GRAPHITE_PORT("graphite.port", 2003),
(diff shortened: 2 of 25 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -101,6 +104,8 @@
     CONSUMER_OFFSET_COMMIT_QUEUE_ALERT_MINIMAL_IDLE_PERIOD("consumer.offset.commit.queue.alert.minimal.idle.period", 3600),
     CONSUMER_OFFSET_COMMIT_QUEUE_ALERT_SIZE("consumer.offset.commit.queue.alert.size", 20_000),
     CONSUMER_HEALTH_CHECK_PORT("consumer.status.health.port", 8000),
+    CONSUMER_WORKLOAD_ALG("consumer.workload.alg", "legacy.mirror"),
```
adamdubiel: we have plenty of space, why not use "algorithm" instead of "alg"?

wojtkiewicz: done

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionName.java
```diff
@@ -20,4 +22,15 @@ public String getName() {
...
+
+    public static SubscriptionName fromString(String string) {
+        String[] tokens = string.split("\\$");
+        checkArgument(tokens.length > 1, "Incorrect string format. Expected 'topic$subscription'. Found:'%s'", string);
+        return new SubscriptionName(tokens[1], TopicName.fromQualifiedName(tokens[0]));
+    }
+
+    @Override
+    public String toString() {
+        return topicName.qualifiedName() + "$" + name;
```
dankraw: maybe we should stick to the good 'ol consumer group id `{group}_{topic}_{subscription}`, as in kafka's `/consumers` zookeeper path, instead of reinventing a new abstraction :)

dankraw: https://cwiki.apache.org/confluence/display/KAFKA/Kafka+data+structures+in+Zookeeper (point 6.)

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/ConsumersSupervisor.java
```diff
@@ -133,15 +125,22 @@ public void onSubscriptionChanged(Subscription modifiedSubscription) {
         }
     }
 
+    public void updateSubscription(Subscription modifiedSubscription) {
+        consumerHolder.get(modifiedSubscription.getTopicName(), modifiedSubscription.getName()).
```
druminski: This method should be synchronized over subscription lock.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workTracking/WorkTracker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public void dropAssignment(Subscription subscription) {
+        askCuratorPolitely(() ->
+                curatorClient.delete().guaranteed().forPath(pathSerializer.serialize(subscription)));
+    }
+
+    private void askCuratorPolitely(CuratorTask task) {
+        try {
+            task.run();
+        } catch (KeeperException.NodeExistsException | KeeperException.NoNodeException ex) {
```
druminski: Do we need this catch? If yes then we should add here logging at warn level instead of writing ignore comment.

wojtkiewicz: That's actually deliberate behaviour to call this method and not check if node actually exists before both node deletion and node creation. I don't want to warn anybody about this because everything is fine.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionName.java
wojtkiewicz: I tried using this format but it's an absolute nightmare to deserilize and has plenty of corner cases when group, topic or subscription starts (or ends) with an underscore. I managed to cover simple cases with `regexp="(?<!_)_(?!_)"` though...

The advantage of using something other than an underscore to separate topic from subscription is that we use an underscore to escape dots in zookeeper paths so parsing such a string is significantly easier. If you don't mind I would rather add few extra rules to topic and subscription validation to not contain dollar sign which should be pretty sensible.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/ConsumersSupervisor.java
wojtkiewicz: done

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionName.java
dankraw: :ok_hand:

druminski:
:+1:

PR #236, opened 2015-11-02
235 - Propagate traceId

jmnarloch:
Makes the hermes aware of Trace-Id.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
@@ -4,7 +4,7 @@ public enum MessageMetadataHeaders {
 
     MESSAGE_ID("Hermes-Message-Id"),
     TOPIC_NAME("Hermes-Topic-Name"),
-    TRACE_ID("Hermes-Trace-Id");
+    TRACE_ID("Trace-Id");
 
     private final String headerName;
 
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
@@ -3,7 +3,8 @@ package pl.allegro.tech.hermes.common.http;
 public enum MessageMetadataHeaders {
 
     MESSAGE_ID("Hermes-Message-Id"),
-    TOPIC_NAME("Hermes-Topic-Name");
+    TOPIC_NAME("Hermes-Topic-Name"),
+    TRACE_ID("Hermes-Trace-Id");
 
     private final String headerName;
 
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
@@ -4,6 +4,7 @@ import com.google.common.collect.Maps;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.util.Utf8;
+import pl.allegro.tech.hermes.api.TraceInfo;
 
 import java.util.Map;
 
@@ -33,13 +34,13 @@ public class AvroMessageContentWrapper {
         }
     }
 
-    byte[] wrapContent(byte[] message, String id, String traceId, long timestamp, Schema schema) {
+    byte[] wrapContent(byte[] message, String id, TraceInfo traceInfo, long timestamp, Schema schema) {
         try {
             GenericRecord genericRecord = bytesToRecord(message, schema);
             Map<Utf8, String> metadata = Maps.newHashMap();
             metadata.put(METADATA_TIMESTAMP_KEY, Long.toString(timestamp));
             metadata.put(METADATA_MESSAGE_ID_KEY, id);
-            ofNullable(traceId).ifPresent((val) -> metadata.put(METADATA_TRACE_ID_KEY, val));
+            ofNullable(traceInfo.getTraceId()).ifPresent((val) -> metadata.put(METADATA_TRACE_ID_KEY, val));
 
             genericRecord.put(METADATA_MARKER, metadata);
             return recordToBytes(genericRecord, schema);
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
@@ -1,17 +1,19 @@
 package pl.allegro.tech.hermes.common.message.wrapper;
 
+import com.google.common.collect.Maps;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.util.Utf8;
 
 import java.util.Map;
 
-import static com.google.common.collect.ImmutableMap.of;
+import static java.util.Optional.ofNullable;
 import static pl.allegro.tech.hermes.common.message.converter.AvroRecordToBytesConverter.bytesToRecord;
 import static pl.allegro.tech.hermes.common.message.converter.AvroRecordToBytesConverter.recordToBytes;
 import static pl.allegro.tech.hermes.common.message.wrapper.AvroMetadataMarker.METADATA_MARKER;
 import static pl.allegro.tech.hermes.common.message.wrapper.AvroMetadataMarker.METADATA_MESSAGE_ID_KEY;
 import static pl.allegro.tech.hermes.common.message.wrapper.AvroMetadataMarker.METADATA_TIMESTAMP_KEY;
+import static pl.allegro.tech.hermes.common.message.wrapper.AvroMetadataMarker.METADATA_TRACE_ID_KEY;
 
 public class AvroMessageContentWrapper {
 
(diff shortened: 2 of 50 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
```diff
@@ -3,7 +3,8 @@
 public enum MessageMetadataHeaders {
 
     MESSAGE_ID("Hermes-Message-Id"),
-    TOPIC_NAME("Hermes-Topic-Name");
+    TOPIC_NAME("Hermes-Topic-Name"),
+    TRACE_ID("Hermes-Trace-Id");
```
adamdubiel: since it uses some custom header, we should also change hermes-client to easily pass it on in request

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageMetadata.java
```diff
@@ -38,6 +46,8 @@ public boolean equals(Object obj) {
             return false;
         }
         final MessageMetadata other = (MessageMetadata) obj;
-        return Objects.equals(this.id, other.id) && Objects.equals(this.timestamp, other.timestamp);
+        return Objects.equals(this.id, other.id)
+                && Objects.equals(this.timestamp, other.timestamp)
+                && Objects.equals(this.traceId, other.traceId);
```
adamdubiel: do we need traceId to identify the object? i think the previous implementation was fine

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingTest.java
```diff
@@ -291,4 +294,27 @@ public void shouldPublishInvalidJsonMessageOnValidationDryRun() {
         assertThat(response).hasStatus(CREATED);
     }
 
+    @Test
+    public void shouldPublishMessageWithTraceId() {
```
adamdubiel: i don't know whats the purpose of this test, the body does hardly anything with trace-id?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/HttpMetaHeaders.java
```diff
@@ -0,0 +1,6 @@
+package pl.allegro.tech.hermes.frontend.publishing;
+
+public interface HttpMetaHeaders {
```
adamdubiel: hmm.. so this is a source Header for trace? it's a bit confusing to see this and then `Hermes-Trace-Id` in MessageMetadataHeaders

hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
druminski: As far as I know services uses 'Trace-Id' header. Since it is the same value I think we should also use same name for the header.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/HttpMetaHeaders.java
jmnarloch: So we need either mapping between the HTTP header and Hermes header or use the same name for both.

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingTest.java
jmnarloch: Ups, I've added the header to the request.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageMetadata.java
jmnarloch: Revoked the change.

jmnarloch:
Turns out that the consumer side was easier then I though it's going to be, here is the second commit.

jmnarloch:
@adamdubiel if you agree I would like to refactor all of the creation of Message in tests and introduce some convenient builder with some default values set.

adamdubiel:
@jmnarloch go ahead :)

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
```diff
@@ -23,20 +25,23 @@ UnwrappedMessageContent unwrapContent(byte[] data, Schema schema) {
...
         try {
             GenericRecord genericRecord = bytesToRecord(message, schema);
-            genericRecord.put(METADATA_MARKER, of(
-                METADATA_TIMESTAMP_KEY, Long.toString(timestamp),
-                METADATA_MESSAGE_ID_KEY, id
-            ));
+            Map<Utf8, String> metadata = Maps.newHashMap();
+            metadata.put(METADATA_TIMESTAMP_KEY, Long.toString(timestamp));
+            metadata.put(METADATA_MESSAGE_ID_KEY, id);
+            ofNullable(traceId).ifPresent((val) -> metadata.put(METADATA_TRACE_ID_KEY, val));
```
wojtkiewicz: is there any reason why trace-id is optional when wrapping message content?

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaMessageReceiverTest.java
```diff
@@ -41,7 +42,8 @@
     private static final Topic TOPIC = topic().withContentType(Topic.ContentType.JSON).withName("group.topic1").build();
     private static final Integer KAFKA_STREAM_COUNT = 1;
     private static final String CONTENT = "{\"test\":\"a\"}";
-    private static final MessageMetadata METADATA = new MessageMetadata(1L, "unique");
+    private static final String TRACE_ID = UUID.randomUUID().toString();
```
wojtkiewicz: this constant isn't really used in any of the test cases

pbetkier:
I would consider if we really want to extend the surface area of Hermes by supporting correlation ids this way. There are many ways of tracing requests and our assumption, a header called Trace-Id, is likely to miss the target. E.g. currently in allegro we would also need to pass span id, parent span id and trace reported flag to be compatible with our tracing infrastructure with Zipkin etc.

So the best we can do in my opinion is to make room for hooking one's own mechanisms instead. I suggest introducing interfaces to implement in order to store additional message metadata in frontends when receiving a request and to enrich the outgoing request using this metadata in consumers. Our internal Hermes setup would then implement them to match our internal tracing infrastructure.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
jmnarloch: Yes it might be absent, and at this point we do not generate any value instead, which make sense since the trace id is a notion of some external id.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/http/MessageMetadataHeaders.java
jmnarloch: Alright I've unified the header naming, for producers and consumers.

jmnarloch:
So my plan for now is to do the fallowing:

* Update the client code - to allow to pass the additional header
* Refactor the tests, so the creation of Message wouldn't require to make so many changes across different tests
* Add true integration tests for this functionality - including sending and receiving both JSON and AVRO messages

* Finally, we can discuss and consider generalization of this idea, so that extra headers/metadata could be passed with the messsage.

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/offset/OffsetCommitQueueTest.java
```diff
@@ -155,8 +155,8 @@ public void shouldSetLastToCommitToNullWhenAllMessagesAreDelivered() {
 
     private void addOffsets(long ... offsets) {
         for (Long offset : offsets) {
-            queue.put(new Message("id", "topic", new byte[0], Topic.ContentType.JSON, 12091212L, 120912123L,
-                    new PartitionOffset(KafkaTopicName.valueOf("kafka_topic"), offset, 0)).getOffset());
+            queue.put(new Message("id", "topic", "traceId", new byte[0], Topic.ContentType.JSON,
```
jmnarloch: Why do we bother with creation of the messages if only what is done here is adding the offset value which looks like will be the same as loop offset variable.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
wojtkiewicz: ok, I understand

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/interpolation/MessageBodyInterpolatorTest.java
```diff
@@ -102,8 +115,16 @@ public void willInterpolateMultipleJsonPathsFromTemplateInReverseOrder() throws
         EndpointAddress endpoint = EndpointAddress.of("http://some.endpoint.com/{some.object}?test={some.test}");
         URI expectedEndpoint = URI.create("http://some.endpoint.com/100?test=hello");
         String jsonMessage = "{\"some\": {\"test\": \"hello\", \"object\": 100}}";
-        Message msg = new Message("id", "some.topic", jsonMessage.getBytes(), Topic.ContentType.JSON, 1232443L, 12324434L,
-                new PartitionOffset(KAFKA_TOPIC, 0, 0));
+//        Message msg = withTestMessage()
```
druminski: Here are some leftovers to clean up. I like that you wrote the builder, maybe we can also use it here.

integration/src/integration/java/pl/allegro/tech/hermes/integration/test/HttpResponseAssertion.java
```diff
@@ -1,6 +1,8 @@
 package pl.allegro.tech.hermes.integration.test;
 
 import javax.ws.rs.core.Response;
+
+import io.netty.handler.codec.http.HttpResponse;
```
druminski: unnecessary import

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyTraceIdAppender.java
```diff
@@ -0,0 +1,15 @@
+package pl.allegro.tech.hermes.consumers.consumer.sender.http;
+
+import org.eclipse.jetty.client.api.Request;
+import pl.allegro.tech.hermes.common.http.MessageMetadataHeaders;
+import pl.allegro.tech.hermes.consumers.consumer.Message;
+import pl.allegro.tech.hermes.consumers.consumer.trace.TraceIdAppender;
+
+public class JettyTraceIdAppender implements TraceIdAppender<Request> {
```
adamdubiel: i think this is not `Jetty` specific implementation, but rather `Header`

jmnarloch:
The pull request has been closed since it was suppose to be targeted to rc branch - new pull request https://github.com/allegro/hermes/pull/238

PR #238, opened 2015-11-04
Hermes 235

jmnarloch:
Adds awareness of trace information (like trace id).


druminski:
:+1:

adamdubiel:
:+1:

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/di/FrontendBinder.java
```diff
@@ -36,6 +38,7 @@ protected void configure() {
 
         bindSingleton(SchemaPrefetchService.class);
         bindSingleton(HealthCheckService.class);
+        bind(HeaderTraceIdExtractor.class).to(TraceExtractor.class).in(Singleton.class);
```
adamdubiel: could we expose this in `FrontendBuilder` and `ConsumersBuilder`? now you need to know about this secret setting

jmnarloch: Ok, both from what I can understand you can provide your own binding directly through the builder: https://github.com/allegro/hermes/blob/master/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/HermesFrontend.java#L159

So is there a need to explicitly provide method for that?

Issue #241, opened 2015-11-09
Expand the supported trace headers


jmnarloch:
We should expand the propagated hermes header with additional Span-Id, Parent-Span-Id, Trace-Sampled, Trace-Reported

PR #243, opened 2015-11-09
241 - Expand the supported trace headers

jmnarloch:
Related hermes issue: https://github.com/allegro/hermes/issues/241


hermes-api/src/main/java/pl/allegro/tech/hermes/api/TraceInfo.java
```diff
@@ -25,12 +57,64 @@ public boolean equals(Object o) {
 
         TraceInfo traceInfo = (TraceInfo) o;
 
-        return !(traceId != null ? !traceId.equals(traceInfo.traceId) : traceInfo.traceId != null);
+        if (traceId != null ? !traceId.equals(traceInfo.traceId) : traceInfo.traceId != null) return false;
```
druminski: Those statements are not easy to read - too many negations. Maybe we could use here java.util.Objects.equals()

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/MessageTrace.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        MessageTrace that = (MessageTrace) o;
+
+        if (traceId != null ? !traceId.equals(that.traceId) : that.traceId != null) return false;
```
druminski: Same here

jmnarloch:
This pull request tries to generalize the header extraction and propagation through hermes, this is way the trace specific types has been removed.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/MessageMetadata.java
```diff
@@ -28,8 +34,8 @@ public long getTimestamp() {
         return timestamp;
     }
 
-    public String getTraceId() {
-        return traceId;
+    public Map<String, String> getExternalMetadata() {
+        return externalMetadata;
```
wojtkiewicz: It would be nice if we didn't return mutable copy of metadata here

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Message.java
```diff
@@ -80,6 +84,10 @@ public String getId() {
         return id;
     }
 
+    public Map<String, String> getExternalMetadata() {
+        return externalMetadata;
```
wojtkiewicz: It would be nice if we didn't return mutable copy of metadata here

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyHttpMessageSenderProvider.java
```diff
@@ -17,19 +17,19 @@
     private final HttpClient httpClient;
     private final EndpointAddressResolver endpointAddressResolver;
     private final int requestTimeout;
-    private final DefaultHttpTraceIdAppender traceIdAppender;
+    private final DefaultHttpMetadataAppender metadataAppender;
```
wojtkiewicz: I think it would be nice if we use an interface ```MetadataAppender<Request>``` here instead of concrete (and default) implementation

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyMessageSender.java
```diff
@@ -22,13 +22,13 @@
     private final HttpClient client;
     private final ResolvableEndpointAddress endpoint;
     private final long timeout;
-    private final DefaultHttpTraceIdAppender traceIdAppender;
+    private final DefaultHttpMetadataAppender metadataAppender;
```
wojtkiewicz: I think it would be nice if we use an interface MetadataAppender<Request> here instead of concrete (and default) implementation

druminski:
:+1:

integration/src/integration/java/pl/allegro/tech/hermes/integration/JmsConsumingTest.java
```diff
@@ -64,6 +65,43 @@ public void shouldPublishAndConsumeJMSMessageWithTraceId() throws Exception {
         assertThat(jmsEndpoint.waitAndGetLastMessage()).assertStringProperty("TraceId", traceId);
     }
 
+    @Test
+    public void shouldPublishAndConsumeJmsMessageWithTraceHeaders() throws Exception {
```
wojtkiewicz: This test could be a little more concise by using instance of ```Invocation.Builder request``` object in ```// given``` block and then in ```// when``` it could be just posted in one line. Then it would be possible to just iterate over all the headers from request and compare them with response if they are the same.

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingTest.java
```diff
@@ -319,4 +324,41 @@ public void shouldPublishAndConsumeMessageWithTraceId() {
         assertThat(response).hasStatus(Response.Status.CREATED);
         assertThat(remoteService.waitAndGetLastRequest()).hasHeaderValue("Trace-Id", traceId);
     }
+
+    @Test
+    public void shouldPublishAndConsumeMessageWithTraceAndSpanHeaders() {
```
wojtkiewicz: This test is very similar to ```JmsConsumingTest.shouldPublishAndConsumeJmsMessageWithTraceHeaders```. If possible it would be nice to make both test cases more concise to make duplication less obvious. If not then maybe we should share some code between these tests...

integration/src/integration/java/pl/allegro/tech/hermes/integration/JmsConsumingTest.java
```diff
@@ -64,6 +65,43 @@ public void shouldPublishAndConsumeJMSMessageWithTraceId() throws Exception {
...
+                .post(Entity.entity(message, MediaType.APPLICATION_JSON));
+
+        // then
+        assertThat(response).hasStatus(Response.Status.CREATED);
+        Message lastMessage = jmsEndpoint.waitAndGetLastMessage();
+        assertThat(lastMessage).assertStringProperty("TraceId", traceId);
+        assertThat(lastMessage).assertStringProperty("SpanId", spanId);
+        assertThat(lastMessage).assertStringProperty("ParentSpanId", parentSpanId);
+        assertThat(lastMessage).assertStringProperty("TraceSampled", sampled);
+        assertThat(lastMessage).assertStringProperty("TraceReported", reported);
```
wojtkiewicz: why do we change header names to CamelCase?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyHttpMessageSenderProvider.java
jmnarloch: I would like to do this, but the binder complains about the generic types:
```
Caused by:
        org.glassfish.hk2.api.UnsatisfiedDependencyException: There was no object available for injection at SystemInjecteeImpl(requiredType=MetadataAppender<Request>,parent=JettyHttpMessageSenderProvider,qualifiers={},position=3,optional=false,self=false,unqualified=null,693596959)
```

wojtkiewicz: Try to use ```TypeLiteral<>``` thing. Something like this should work:
```bindFactory(JsonSchemaRepositoryFactory.class).in(Singleton.class).to(new TypeLiteral<SchemaRepository<JsonSchema>>() {});```

File: integration/src/integration/java/pl/allegro/tech/hermes/integration/JmsConsumingTest.java (hunk 9 above)
Comment (jmnarloch) [inline integration/src/integration/java/pl/allegro/tech/hermes/integration/JmsConsumingTest.java]: That's an actually very good question, since according to the specs: https://docs.oracle.com/javaee/6/api/javax/jms/Message.html - the underscore is valid indentifier also see http://www-01.ibm.com/support/knowledgecenter/SSDKJ8_6.2.0/com.ibm.wpg.entadv.doc/config/integration/creating_jms_properties_bo.html

But due to the existing convention of other header handling I've implemented this in the same manner.

Comment (wojtkiewicz) [inline integration/src/integration/java/pl/allegro/tech/hermes/integration/JmsConsumingTest.java]: Ok, I understand. From what I remember we have very little JMS subscribers. If we check how many of them adopted these tracing features then maybe it would be possible to use underscores as you suggested.

File: hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyHttpMessageSenderProvider.java [hunk 10]
```diff
@@ -17,19 +19,19 @@
...
-    private final DefaultHttpTraceIdAppender traceIdAppender;
+    private final MetadataAppender<Request> metadataAppender;
 
     @Inject
     public JettyHttpMessageSenderProvider(
             HttpClient httpClient,
             ConfigFactory configFactory,
             EndpointAddressResolver endpointAddressResolver,
-            DefaultHttpTraceIdAppender traceIdAppender) {
+            DefaultHttpMetadataAppender metadataAppender) {
```
wojtkiewicz: I think it is possible to inject ```MetadataAppender<Request> metadataAppender``` here using some hk2 TypeLiteral magic

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingTest.java
```diff
@@ -319,4 +329,36 @@ public void shouldPublishAndConsumeMessageWithTraceId() {
...
+
+        // given
+        String message = "{\"id\": 101}";
+        TraceContext trace = new TraceContext(
+                UUID.randomUUID().toString(),
+                UUID.randomUUID().toString(),
+                UUID.randomUUID().toString(),
+                "1",
+                "0"
+        );
```
wojtkiewicz: ...but my point was to actually have less text to read, especially if values are random or arbitrary :-)

For instance something like this would be more concise but still readable:
```
// given
TraceHeaders headers = TraceHeaders.random();

// when
Response response = headers.addAll(client.request()).post()

// then
assertThat(response.headers().entrySet()).containsAll(headers.entrySet())
```

jmnarloch: That's done.

jmnarloch:
@wojtkiewicz , please take a look if this is what you wanted?

wojtkiewicz:
yeah, good enough :-)

:+1:

PR #278, opened 2015-12-21
277 elasticsearch tracker memory leak fix

pbetkier:
Keeps already constructed Elasticsearch documents on the tracker queue instead of unclosed document builders. Fixes memory leaks when exceptions are thrown from `ElasticsearchQueueCommitter#processBatch`, e.g. due to Elasticsearch unavailability, which caused builders not to get closed.

Also, decreases memory footprint of the tracker queue elements from 16 kB to 1 kB by overriding the default size of Elasticsearch's `BytesStreamOutput`. This means that 100k elements now take ~100 MB instead of ~1.6 GB heap memory.


adamdubiel:
:+1: remember about squash

pbetkier:
Actually, I split into commits on purpose. They are both green and solve separate issues.

dankraw:
:+1: nice :)

PR #280, opened 2015-12-22
Added necessary parameter in 'create subscription' request

kretes:


kretes:
Have you considered making a living documentation e.g. with Concordion - we did it for some rest APIs and it was quite nice and transparent - e.g. of non-branded report here: http://pragmatists.github.io/OpenTrapp/concordion/projects/Projects.html

PR #284, opened 2015-12-27
276 moved hostname to reporter prefix

wojtkiewicz:
I refactored hermes metrics slightly and moved hostname to reporter prefix. This allows metric names to be free from host specific information and be queried from tag based metric repositories (like kairosDB).


adamdubiel:
:+1: +squash

pawelszymczyk:
:+1:

PR #289, opened 2016-01-04
231 ability to change subscription endpoint

pawelszymczyk:


pawelszymczyk:
It need additiona changes in consumers.

PR #293, opened 2016-01-11
292 Support for multi content types in subscryption

pawelszymczyk:


pawelszymczyk:
Support for reading JSON from Kafka and sending it as Avro is not introduced in this PR. We need to discuss whether we want implements this scenario.

pawelszymczyk:
Closed due to lack of possibility to dynamic subscription content-type changing.

Issue #310, opened 2016-01-19
HQL with no Java Beans


druminski:
Currently Hermes Query Language works only for Java Beans. It means that if we have field type of boolean and access to it through method wasXXX() then HQL will fail throw exception. We should fix it.

PR #311, opened 2016-01-19
Issue #310 | HQL works with no java beans

druminski:


wojtkiewicz:
:+1:

jmnarloch:
By this ocassion there is one more thing that you could do in JsonQueryParser: https://github.com/allegro/hermes/blob/310_hql_with_no_java_beans/hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/query/parser/json/JsonQueryParser.java#L72

is to branch to handle boolean values, something like: 

```
if(!node.isValueNode()) {
            throw new ParseException("The node value wasn't present");
        }
        if (node.isTextual()) {
            return node.asText();
        } else if (node.isBoolean()) {
            return node.asBoolean();
        } if (node.isInt()) {
            return node.asInt();
        } else if(node.isDouble()) {
            return node.asDouble();
        }
        return null;
```

Thanks

druminski:
@jmnarloch good idea, done

adamdubiel:
:+1:

Issue #239, opened 2015-11-04
Zero subscriptions in metric endpoint


druminski:
Endpoint with path /topics/{topicName}/metrics returns json with 'subscriptions' field which always have value 0. The field is nowhere filled. We should remove it or fix it.

PR #322, opened 2016-01-23
#239 Fill #subscription in topic metrics

alberskib:


adamdubiel:
I don't know if this counts as a metric we had in mind, any usecase maybe?

alberskib:
To be honest I don't see any metric that would be reasonable fit into this field (int) - perhaps better option is to remove subscriptions, but (probably) it should go with new version of the API.

adamdubiel:
Oh, wait, i didn't notice that this field was already in the model, just not used. I will dig around in the history to try to find what happened to it and then just merge the pull request :)

alberskib:
Feel free to decline PR if you think that it is not good fit for this field :)

PR #324, opened 2016-01-24
fixing bug in frotnend counter reporting

adamdubiel:
both 'publish' and 'unpublish' counters have the same prefix,
since hosntame was removed, sometimes (data race) unpublished
was reported as published, which casuses us to lose topic
published messages metrics (they are no longer monotonic)


dankraw:
:+1:

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/CounterMatcher.java
```diff
@@ -27,19 +25,19 @@ private void parseCounter(String counterName) {
     }
 
     public boolean isTopicPublished() {
-        return counterName.contains("published.");
```
dankraw: ouch! :neutral_face:

druminski:
:+1:

PR #332, opened 2016-01-29
configuring circle ci

adamdubiel:
let's try other CI, since Travis had a lot of problems with building Hermes lately (false positives) and is virtually undebuggable


druminski:
:+1:

Issue #313, opened 2016-01-19
Support https protocol


sslavic:
Hermes currently doesn't allow subscribers/consumers to be https endpoints, only http and jms - probably this https://github.com/allegro/hermes/blob/master/hermes-api/src/main/java/pl/allegro/tech/hermes/api/constraints/EndpointAddressValidator.java validation prevents https subscribers to register. Please consider allowing https endpoints as well, even if as a feature flag which is off by default.

adamdubiel:
It's bit more complicated than just validation, but not too hard either. I suppose you would want to have a way to specify truststore (so we don't accept selfsigns)?

sslavic:
Initially no custom truststore support needed, only default Java truststore (jssecacerts if it exists, otherwise cacerts), so no self-signed certs supported. Subscriber server/endpoint has to have trusted certificate to establish https and has to trust any client, so client (hermes-consumers) does not have to have certificate.

PR #343, opened 2016-02-08
HTTPS support in message senders

dankraw:
Tested on MacOS X and Ubuntu 15 VM with [Mockbin](https://mockbin.org) used as a valid HTTPS subscriber endpoint.

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientFactory.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientFactory.java
@@ -2,29 +2,39 @@ package pl.allegro.tech.hermes.consumers.consumer.sender.http;
 
 import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.util.HttpCookieStore;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.glassfish.hk2.api.Factory;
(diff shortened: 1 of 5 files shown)


dankraw:
resolves #313

adamdubiel:
:+1:

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientFactory.java
```diff
@@ -2,29 +2,39 @@
...
 
     private final ConfigFactory configFactory;
     private final InstrumentedExecutorServiceFactory executorFactory;
+    private final SslContextFactory sslContextFactory;
 
     @Inject
     public HttpClientFactory(ConfigFactory configFactory, InstrumentedExecutorServiceFactory executorFactory) {
         this.configFactory = configFactory;
         this.executorFactory = executorFactory;
+        this.sslContextFactory = sslContextFactory();
```
pawelszymczyk: why SslContextFactory isn't injected?

wojtkiewicz:
:+1:

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientFactory.java
dankraw: right, this could be injected, but anyway it has a very narrow scope as it's used only by the http client.

PR #350, opened 2016-02-10
342 adaptive consumer backoff

chemicL:


docs/docs/user/subscribing.md
```diff
@@ -105,6 +105,19 @@ failure the rescue team has one hour to fix the problem before any event will be
 
 By default inflight TTL is set to 3600 seconds (an hour) and retry backoff is set to 100ms.
 
+## Adaptive backoff
+
+The client is able to signal it can't handle the message at the moment and Hermes Consumer will retry delivery after
+minimum of given delay.
+
+The endpoint can return **Retry-After** header, with the amount of seconds to backoff, combined with status **503**.
```
adamdubiel: so the maximum is inflight ttl?

docs/docs/user/subscribing.md
```diff
@@ -105,6 +105,19 @@ failure the rescue team has one hour to fix the problem before any event will be
 
 By default inflight TTL is set to 3600 seconds (an hour) and retry backoff is set to 100ms.
 
+## Adaptive backoff
```
adamdubiel: i think we shouldn't call it adaptive :) it;s just forced backoff, we don't really adapt, just follow the orders

docs/docs/user/subscribing.md
chemicL: Correct.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/MessageSendingResult.java
```diff
@@ -56,14 +62,26 @@ public boolean isClientError() {
...
-        if (!isInFamily(SUCCESSFUL)) {
+        if (this.failure == null && !isInFamily(SUCCESSFUL)) {
             this.failure = new InternalProcessingException("Message sending failed with status code:" + statusCode);
         }
     }
 
+    private long extractRetryAfterMillis(Result result) {
+        HttpFields headers = result.getResponse().getHeaders();
+        if (headers.contains(HttpHeader.RETRY_AFTER)) {
+            return Long.parseLong(headers.get(HttpHeader.RETRY_AFTER)) * 1000L;
```
dankraw: Maybe some try/catch for NumberFormatException would be nice here.

docs/docs/user/subscribing.md
chemicL: Perhaps back pressure would be more meaningful?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/MessageSendingResult.java
```diff
@@ -56,14 +62,26 @@ public boolean isClientError() {
...
+        }
+        return 0L;
```
wojtkiewicz: I think that better idea would be to retry after backoff from subscription policy rather than hammer again service that is currently not working...

docs/docs/user/subscribing.md
chemicL: Actually, it's not a maximum in which we'll retry. Just to keep it simple: the TTL will still expire messages regardless of the Retry-After logic. It's up to the user to ensure the Retry-After delay is less than Inflight TTL. I'll update the docs.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -142,13 +142,19 @@ public void accept(MessageSendingResult result) {
...
                 } else {
                     handleMessageDiscarding(message, result);
                 }
             }
         }
 
+        private long extractRetryDelay(MessageSendingResult result) {
+            return result.isRetryLater()
+                    ? result.getRetryAfterMillis()
+                    : subscription.getSubscriptionPolicy().getMessageBackoff();
```
wojtkiewicz: I think that if header Retry-After is absent then we should respect message backoff from policy...

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -142,13 +142,19 @@ public void accept(MessageSendingResult result) {
                 handleFailedSending(message, result);
                 if (!isTtlExceeded(message) && shouldRetrySending(result)) {
                     retrySingleThreadExecutor.schedule(() -> retrySending(result),
-                            subscription.getSubscriptionPolicy().getMessageBackoff(), TimeUnit.MILLISECONDS);
+                            extractRetryDelay(result), TimeUnit.MILLISECONDS);
```
wojtkiewicz: this is not ok to schedule tasks using only time from received header:
- we should validate if value is within reasonable boundries (0, maxMessageBackoff)
- we should discard message if retry-after is after message TTL because we will never send this message anyway

wojtkiewicz: Remember that if you schedule retrySending then actual attempt will happen because TTL is checked when processing response... We don't have any logic that would clear jobs from retry executor...

wojtkiewicz: Putting jobs with delay equal to Integer.MAX will cause them to stay for... Integer.MAX milliseconds in our queue...

hermes-consumers/src/main/resources/config.properties
```diff
@@ -0,0 +1,3 @@
+kafka.broker.list=10.10.10.10:9092
```
adamdubiel: why should we keep it in main as default config?

adamdubiel:
i think we could use some kind of integration test (for example wait for 2 seconds with retry-after 1.5second and make sure it was retried once only)

chemicL:
Sure, an integration test will be useful. I'll make it wait 1.5 seconds with Retry-After set to 1 second, as it needs to remain integer. Specs:

> this field can be either an HTTP-date or an integer number of seconds (in decimal)

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSenderTest.java
```diff
@@ -274,6 +276,38 @@ public void shouldBackoffRetriesWhenEndpointFails() throws InterruptedException
         verifyErrorHandlerHandleFailed(message, subscriptionWithBackoff, 1 + executionTime / senderBackoffTime);
     }
 
+    @Test
+    public void shouldBackoffRetriesOnRetryAfter() throws InterruptedException {
+        // given
+        int retrySeconds = 1;
```
wojtkiewicz: can we define retry-after in ms? 50 ms for instance

chemicL: I understand it is not ideal for testing time, but I'd rather not invest in refactors driven by tests, while the core implementation accepts only seconds per [specification](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html):
> The value of this field can be either an HTTP-date or an integer number of seconds (in decimal)

wojtkiewicz: I was not aware of that. Weird choice. We could add custom header Retry-After-Millis and accept milliseconds but we have to test original Retry-After anyway so your point is valid.

adamdubiel:
:+1:

wojtkiewicz:
:+1:

Issue #320, opened 2016-01-22
Slow consumer assignments rebalance when a leader drops his assignments


dankraw:
Related to the **selective** consumers workload algorithm implementation.

After performing rebalance and writing changes do zookeeper, all nodes (including the leader itself) receive events about their assignments (see `SubscriptionAssignmentRegistry`). When the leader consumer node has to release some amount his work i.e. stop some of his kafka consumers, due to the limit of threads used this may take time, zk-events get queued and delivered **later** than another rebalance starts. This makes the leader node to have **stale cache** state and repeat the rebalance even if there are no changes in the cluster.

We should consider running the release-work job separately from actually accepting the assignments change events from zookeeper.

PR #368, opened 2016-02-18
320 nonblocking subscription assignment processing for selective consumer workload

wojtkiewicz:
This PR addresses the issue #320 by separating handling zk events from actual assignment processing. For instance handling onAssignmentRemoved will no longer block until actual consumer is killed.


chemicL:
:+1:

dankraw:
:+1: :rooster:

PR #378, opened 2016-02-26
374 separated handling of zk events from processing consumer commands

wojtkiewicz:
1. Introduced handling of consumer commands on separate ThreadPoolExecutor that has a property of providing sequence of unique tasks (via LinkedHashSetBlockingQueue).
2. Consumer suspension doesn't wait for message TTL to expire anymore.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/cache/queue/QueueTask.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/cache/queue/QueueTask.java
@@ -1,0 +1,31 @@
+package pl.allegro.tech.hermes.common.cache.queue;
+
+import java.util.Arrays;
+
+public class QueueTask implements Runnable {
+    private final Runnable operation;
+    private final Object[] payload;
+
+    public QueueTask(Runnable operation, Object... payload) {
+        this.operation = operation;
+        this.payload = payload;
+    }
+
+    @Override
+    public void run() {
+        operation.run();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        QueueTask that = (QueueTask) o;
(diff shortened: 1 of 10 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/cache/queue/QueueTask.java
```diff
@@ -0,0 +1,31 @@
+package pl.allegro.tech.hermes.common.cache.queue;
+
+import java.util.Arrays;
+
+public class QueueTask implements Runnable {
+    private Runnable operation;
```
adamdubiel: both are finals

adamdubiel:
So `Queue` task payload is there only to be able to group them and eliminate duplicates?

wojtkiewicz:
QueueTask.payload is an array of arbitrary objects that has to be equal to another QueueTask.payload to be considered duplicate...

hermes-common/src/main/java/pl/allegro/tech/hermes/common/cache/queue/QueueTask.java
wojtkiewicz: done

druminski:
So if we have few updates on same subscription then only one of them can be executed because others are treated as duplicates?

wojtkiewicz:
Updates which are identical are considered duplicates. We take into consideration Subscription, Subscription.state and zkEvent.name. If someone emits sequence of active/suspend/activate/suspend then we should discard such events if we have them already enqueued.

druminski:
Ok, that make sense but how about this type of sequence: update/update/update? first update can change for example ttl, second backoff, third retries-client-errors, it would be nice to apply all of them instead of only the first one.

wojtkiewicz:
These updates are different so they are going to be enqueued.

druminski:
:+1:

adamdubiel:
:+1:

Issue #356, opened 2016-02-12
Hermes is subscription-monitoring friendly


adamdubiel:
We could make Hermes subscription-monitoring friendly by introducing few changes. By subscription-monitoring friendly i mean, that it is easy to apply automated monitoring for each subscription.

## Expose subscription health status via management

Currently Console subscription view shows information about subscription health status. This is calculated by Console based on topic and subscription metrics. We should expose this kind of data in management - calculate in on backed side so it is available via API. Format would be:

* overall status (healthy/unhealthy)
* reasons (like on console)

## Add severity and monitoring integration for subscription

We should enrich the model of subscription and add "Monitoring" section including:

* severity: critical/important/non-important which would serve as an information for monitoring how to react when subscriptions becomes unhealthy
* reaction: generic field that could contain for example team name or Pager Duty ID - anything that could be useful to call the team responsible for subscription

Adding this features is easy, low-cost operation that would give us great benefits - automated subscription monitoring for all clients.

PR #381, opened 2016-03-02
356: Hermes is subscription-monitoring friendly

PiotrGoralczyk:
First part of #356.

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/MonitoringDetails.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/MonitoringDetails.java
@@ -1,0 +1,58 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotNull;
+import java.util.Objects;
+
+public final class MonitoringDetails {
+    public static final MonitoringDetails EMPTY = new MonitoringDetails(Severity.NON_IMPORTANT, "");
+
+    @NotNull
+    private final Severity severity;
+
+    @NotNull
+    private final String reaction;
+
+    @JsonCreator
+    public MonitoringDetails(@JsonProperty("severity") Severity severity,
+                             @JsonProperty("reaction") String reaction) {
+        this.severity = severity;
+        this.reaction = reaction;
+    }
+
+    public Severity getSeverity() {
+        return severity;
+    }
+
+    public String getReaction() {
+        return reaction;
+    }
+
+    public enum Severity {
+        CRITICAL, IMPORTANT, NON_IMPORTANT
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof MonitoringDetails)) return false;
+        MonitoringDetails that = (MonitoringDetails) o;
+        return severity == that.severity &&
+                Objects.equals(reaction, that.reaction);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(severity, reaction);
+    }
+
+    @Override
+    public String toString() {
+        return "MonitoringDetails{" +
+                "severity=" + severity +
+                ", reaction='" + reaction + '\'' +
+                '}';
+    }
+}
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
@@ -51,6 +51,9 @@ public class Subscription {
     private String contact;
 
     @NotNull
+    private final MonitoringDetails monitoringDetails;
+
+    @NotNull
     private DeliveryType deliveryType = DeliveryType.SERIAL;
 
     public enum State {
@@ -58,15 +61,17 @@ public class Subscription {
     }
 
     private Subscription(TopicName topicName,
-                        String name,
-                        EndpointAddress endpoint,
-                        State state,
-                        String description,
-                        Object subscriptionPolicy,
-                        boolean trackingEnabled,
-                        String supportTeam,
-                        String contact,
-                        ContentType contentType, DeliveryType deliveryType) {
+                         String name,
+                         EndpointAddress endpoint,
+                         State state,
+                         String description,
+                         Object subscriptionPolicy,
+                         boolean trackingEnabled,
+                         String supportTeam,
+                         String contact,
+                         MonitoringDetails monitoringDetails,
+                         ContentType contentType,
+                         DeliveryType deliveryType) {
         this.topicName = topicName;
         this.name = name;
         this.endpoint = endpoint;
@@ -75,6 +80,7 @@ public class Subscription {
         this.trackingEnabled = trackingEnabled;
         this.supportTeam = supportTeam;
         this.contact = contact;
+        this.monitoringDetails = monitoringDetails == null ? MonitoringDetails.EMPTY : monitoringDetails;
         this.contentType = contentType == null ? ContentType.JSON : contentType;
         this.deliveryType = deliveryType;
         this.batchSubscriptionPolicy = this.deliveryType == DeliveryType.BATCH ? (BatchSubscriptionPolicy) subscriptionPolicy : null;
@@ -82,31 +88,33 @@ public class Subscription {
     }
 
     public static Subscription createSerialSubscription(TopicName topicName,
-                        String name,
-                        EndpointAddress endpoint,
-                        State state,
-                        String description,
-                        SubscriptionPolicy subscriptionPolicy,
-                        boolean trackingEnabled,
-                        String supportTeam,
-                        String contact,
-                        ContentType contentType) {
-        return new Subscription(topicName, name, endpoint, state, description, subscriptionPolicy, trackingEnabled, supportTeam,
-                contact, contentType, DeliveryType.SERIAL);
-    }
-
-    public static Subscription createBatchSubscription(TopicName topicName,
                                                         String name,
                                                         EndpointAddress endpoint,
                                                         State state,
                                                         String description,
-                                                        BatchSubscriptionPolicy subscriptionPolicy,
+                                                        SubscriptionPolicy subscriptionPolicy,
                                                         boolean trackingEnabled,
                                                         String supportTeam,
                                                         String contact,
+                                                        MonitoringDetails monitoringDetails,
                                                         ContentType contentType) {
         return new Subscription(topicName, name, endpoint, state, description, subscriptionPolicy, trackingEnabled, supportTeam,
-                contact, contentType, DeliveryType.BATCH);
+                contact, monitoringDetails, contentType, DeliveryType.SERIAL);
+    }
+
+    public static Subscription createBatchSubscription(TopicName topicName,
+                                                       String name,
+                                                       EndpointAddress endpoint,
+                                                       State state,
+                                                       String description,
+                                                       BatchSubscriptionPolicy subscriptionPolicy,
+                                                       boolean trackingEnabled,
+                                                       String supportTeam,
+                                                       String contact,
+                                                       MonitoringDetails monitoringDetails,
+                                                       ContentType contentType) {
+        return new Subscription(topicName, name, endpoint, state, description, subscriptionPolicy, trackingEnabled, supportTeam,
+                contact, monitoringDetails, contentType, DeliveryType.BATCH);
     }
 
     @JsonCreator
@@ -119,6 +127,7 @@ public class Subscription {
                                       @JsonProperty("trackingEnabled") boolean trackingEnabled,
                                       @JsonProperty("supportTeam") String supportTeam,
                                       @JsonProperty("contact") String contact,
+                                      @JsonProperty("monitoringDetails") MonitoringDetails monitoringDetails,
                                       @JsonProperty("contentType") ContentType contentType,
                                       @JsonProperty("deliveryType") DeliveryType deliveryType) {
         DeliveryType validDeliveryType = deliveryType == null ? DeliveryType.SERIAL : deliveryType;
@@ -135,6 +144,7 @@ public class Subscription {
                 trackingEnabled,
                 supportTeam,
                 contact,
+                monitoringDetails,
                 contentType,
                 validDeliveryType
         );
@@ -142,7 +152,7 @@ public class Subscription {
 
     @Override
     public int hashCode() {
-        return Objects.hash(endpoint, topicName, name, description, serialSubscriptionPolicy, batchSubscriptionPolicy, contentType);
+        return Objects.hash(endpoint, topicName, name, description, serialSubscriptionPolicy, batchSubscriptionPolicy, trackingEnabled, supportTeam, contact, monitoringDetails, contentType);
     }
 
     @Override
@@ -164,6 +174,7 @@ public class Subscription {
                 && Objects.equals(this.trackingEnabled, other.trackingEnabled)
                 && Objects.equals(this.supportTeam, other.supportTeam)
                 && Objects.equals(this.contact, other.contact)
+                && Objects.equals(this.monitoringDetails, other.monitoringDetails)
                 && Objects.equals(this.contentType, other.contentType);
     }
 
@@ -231,6 +242,10 @@ public class Subscription {
         return contact;
     }
 
+    public MonitoringDetails getMonitoringDetails() {
+        return monitoringDetails;
+    }
+
     public ContentType getContentType() {
         return contentType;
     }
@@ -271,6 +286,7 @@ public class Subscription {
                     trackingEnabled,
                     supportTeam,
                     contact,
+                    monitoringDetails,
                     contentType,
                     deliveryType
             );
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealth.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealth.java
@@ -1,0 +1,74 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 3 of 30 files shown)


hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionHealthCalculator.java
```diff
@@ -0,0 +1,139 @@
...
+            EnumSet<Problem> healthProblems = calculateHealthProblems(subscription, internalTopicMetrics, internalSubscriptionMetrics);
+            return SubscriptionHealth.of(healthProblems);
+        } catch (NumberFormatException e) {
+            return SubscriptionHealth.NO_DATA;
+        }
+    }
+
+    private EnumSet<Problem> calculateHealthProblems(Subscription subscription, TopicMetrics topicMetrics, SubscriptionMetrics subscriptionMetrics) {
+        EnumSet<Problem> healthProblems = noneOf(Problem.class);
+        if (isLagging(topicMetrics, subscriptionMetrics)) {
```
PiotrGoralczyk: I thought about refactoring these "ifs" into strategies, but isnt't it overkill?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionHealthCalculator.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private boolean isCode4xxErrorRateHigh(SubscriptionMetrics subscriptionMetrics) {
+        return subscriptionMetrics.code4xxErrorsRate > MAX_4XX_ERRORS_RATIO * subscriptionMetrics.rate;
+    }
+
+    private static final class TopicMetrics {
+        private final double rate;
+
+        private TopicMetrics(pl.allegro.tech.hermes.api.TopicMetrics that) {
+            this.rate = parseDouble(that.getRate());
```
PiotrGoralczyk: There is a null check in JS: https://github.com/allegro/hermes/blob/master/hermes-console/static/js/console/subscription/SubscriptionHealth.js#L6 - I omitted it, is it necessary?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionHealthCalculator.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private SubscriptionHealth calculateActiveSubscriptionHealth(Subscription subscription, pl.allegro.tech.hermes.api.TopicMetrics topicMetrics,
+                                                                 pl.allegro.tech.hermes.api.SubscriptionMetrics subscriptionMetrics) {
+        try {
+            TopicMetrics internalTopicMetrics = new TopicMetrics(topicMetrics);
+            SubscriptionMetrics internalSubscriptionMetrics = new SubscriptionMetrics(subscriptionMetrics);
+            EnumSet<Problem> healthProblems = calculateHealthProblems(subscription, internalTopicMetrics, internalSubscriptionMetrics);
+            return SubscriptionHealth.of(healthProblems);
+        } catch (NumberFormatException e) {
```
PiotrGoralczyk: FYI: In the previous JS implementation, we would return a partial result in the case where subscription metrics are available but topic metrics aren't. But here we'll return NO_DATA (without problems).

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
```diff
@@ -120,6 +126,13 @@ public SubscriptionMetrics getSubscriptionMetrics(TopicName topicName, String su
         return metricsRepository.loadMetrics(topicName, subscriptionName);
     }
 
+    public SubscriptionHealth getSubscriptionHealth(TopicName topicName, String subscriptionName) {
+        Subscription subscription = getSubscriptionDetails(topicName, subscriptionName);
+        TopicMetrics topicMetrics = topicService.getTopicMetrics(topicName);
```
PiotrGoralczyk: Out of curiosity: why doesn't TopicService.getTopicMetrics() ensure that the topic exists?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionHealthCalculator.java
wojtkiewicz: You could add more behaviour to Problem instances so that each Problem can detect itself based on given Subscription (since this is SubscriptionHealthCalculator). Then you could iterate over entire class extent and make those problems raise themselves :-)

PiotrGoralczyk: So, Strategy pattern implemented on top of an enum? :)

wojtkiewicz: you basically implemented it already, all what is left is to move 'problem detectors' to actual problem definitions making them more useful...

PiotrGoralczyk: OK, I just wanted to check if I understand you correctly. :)

PiotrGoralczyk:
I did the refactoring I discussed with @wojtkiewicz. If you can, please refer to my other comments :)

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
wojtkiewicz: This is very counter intuitive. We have certain metrics available via zookeeper so if we try to get metrics for non existing topic then zookeeper would create empty nodes for us which is a problem because for instance listing available topics is dependent on zookeeper tree.

wojtkiewicz:
:+1:

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
PiotrGoralczyk: OK, thanks for the explanation :)

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MonitoringDetails.java
```diff
@@ -0,0 +1,58 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.constraints.NotNull;
+import java.util.Objects;
+
+public final class MonitoringDetails {
+    public static final MonitoringDetails EMPTY = new MonitoringDetails(Severity.NON_IMPORTANT, "");
```
adamdubiel: package scope?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -51,6 +51,9 @@
     private String contact;
 
     @NotNull
+    private MonitoringDetails monitoringDetails = MonitoringDetails.EMPTY;
```
adamdubiel: hm, looks like could be final (see code in constructor)

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealth.java
```diff
@@ -0,0 +1,70 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Sets.immutableEnumSet;
+import static java.util.Collections.emptySet;
+
+public final class SubscriptionHealth {
+    public static final SubscriptionHealth HEALTHY = new SubscriptionHealth(Status.HEALTHY, emptySet());
+    public static final SubscriptionHealth NO_DATA = new SubscriptionHealth(Status.NO_DATA, emptySet());
+
+    private final Status status;
+    private final ImmutableSet<SubscriptionHealthProblem> problems;
```
adamdubiel: this is implementation detail, Set would be ok

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
```diff
@@ -0,0 +1,116 @@
+package pl.allegro.tech.hermes.api;
+
+public enum SubscriptionHealthProblem {
```
adamdubiel: those constants work poorly for subscriptions with little traffic (like 1rps) - maybe they need to be adjusted somehow?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public static final class TopicMetrics {
+        private final double rate;
+
+        public TopicMetrics(double rate) {
+            this.rate = rate;
+        }
+    }
+
+    public static final class SubscriptionMetrics {
```
adamdubiel: why are Subscription/Topic Metrics defined under `SubscriptionHealthProblem` ?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
adamdubiel: this class is misplaced - API should contain only simple enum which is used in API. hide calculations in some better place, this way you wont have to create some strange inner classes.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealth.java
PiotrGoralczyk: Yeah, but actually, this *is* implementation (a private field). :) The getter returns a Set. I've left ImmutableSet here just for the sake of clarity that SubscriptionHealth is immutable.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MonitoringDetails.java
PiotrGoralczyk: I need it in pl.allegro.tech.hermes.test.helper.builder.SubscriptionBuilder :(

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
```diff
@@ -120,6 +126,13 @@ public SubscriptionMetrics getSubscriptionMetrics(TopicName topicName, String su
...
+        SubscriptionMetrics subscriptionMetrics = getSubscriptionMetrics(topicName, subscriptionName);
```
adamdubiel: maybe we should invest in some subscriptionmetrics cache now, those values won't change per request, but in time

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
PiotrGoralczyk: OK

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
PiotrGoralczyk: I changed the implementation to this (Strategy implemented using an enum) after our earlier discussion with @wojtkiewicz 's in the comments. Please see: https://github.com/allegro/hermes/pull/381#discussion-diff-54714790

Another solution would be to leave plain enum constants and move problem calculation to strategies corresponding to those constants...

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
PiotrGoralczyk: Because their single use case is to pass metrics to SubscriptionHealthProblem.occurs(). I created these classes because the top-level SubscriptionMetrics and TopicMetrics contain String values which have to be parsed (and can be unparsable). I thought that it is the responsibility of SubscriptionHealthChecker to decide whether the subscription health is HEALTHY, UNHEALTHY or NO_DATA and the SubscriptionHealthProblems responsibility is to calculate if they occur.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
adamdubiel: oh, i see. still, i don't like how domain is mixed with API, this is perfect example of tight coupling, especially that calculations will probably be corrected (or parameterized), which would force reafactoring in API

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
adamdubiel: i think this is all because of misplaced calculations

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionHealthProblem.java
wojtkiewicz: I think that the actual problem is that Subscription is used both as an API and as a domain object. From a domain object I would expect some useful behaviour and API I would expect to be stable. But the thing is that our API is JSON over HTTP and not Java-RPC.

PiotrGoralczyk: @wojtkiewicz +1000000

wojtkiewicz: I'm sure we can work something out but we have to map label to implementation somewhere. I like to keep these two close if possible.

adamdubiel: enums are not flexbile/extensible at all. if i want to make those parameters accessible via configuration, i need to refactor the whole thing to accept configuraion somehow, i don't care if enum is close to the health problem calculator

wojtkiewicz: because if something is parameterized then it should come as an input to an algorithm which can be kept close to label (but doesn't have to)

adamdubiel: i would rather decouple label and calculation in this case

wojtkiewicz: I know :-), I added followup to a previous comment after I submitted that...

PiotrGoralczyk: OK, so Map<HealthProblem, HealthProblemCalculator> in HealthProblemChecker, filled by Spring DI? This decouples HealthProblems fromHealthProblemCalculators and vice versa.

I didn't really get the fragment about configuration parameters :P

wojtkiewicz: there is an ongoing trend that we put stuff into production and then we have to tweak stuff from configuration because we don't know good values upfront, extracting to configuration helps to shorten feedback loop

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
PiotrGoralczyk: Implemented in #394

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/SubscriptionHealthContextCreator.java
```diff
@@ -0,0 +1,31 @@
+package pl.allegro.tech.hermes.management.domain.subscription.health;
+
+import pl.allegro.tech.hermes.api.Subscription;
+
+import static java.lang.Double.parseDouble;
+
+class SubscriptionHealthContextCreator {
+    public SubscriptionHealthContext createContext(Subscription subscription, pl.allegro.tech.hermes.api.TopicMetrics topicMetrics,
+                                                   pl.allegro.tech.hermes.api.SubscriptionMetrics subscriptionMetrics) {
```
druminski: qualified name can be replaced by import

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/SubscriptionHealthContextCreator.java
```diff
@@ -0,0 +1,31 @@
+package pl.allegro.tech.hermes.management.domain.subscription.health;
+
+import pl.allegro.tech.hermes.api.Subscription;
+
+import static java.lang.Double.parseDouble;
+
+class SubscriptionHealthContextCreator {
```
druminski: Is in this package any class name conflict? if no then all qualified names from it can by replaced by imports.

druminski:
:+1:

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/SubscriptionHealthContextCreator.java
PiotrGoralczyk: Unfortunately pl.allegro.tech.hermes.api.SubscriptionMetrics conflicts with pl.allegro.tech.hermes.management.domain.subscription.health.SubscriptionMetrics (created below in convertSubscriptionMetrics()). I created the second SubscriptionMetrics class in domain because the one in API can contain arbitrary strings, e.g. "unavailable" and it would complicate SubscriptionHealthProblemIndicator implementations.

PR #387, opened 2016-03-09
384 handling subscription endpoint change in consumers instead of adm…

chemicL:
…in tool thread

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
@@ -9,6 +9,7 @@ import pl.allegro.tech.hermes.consumers.consumer.rate.ConsumerRateLimiter;
 import pl.allegro.tech.hermes.consumers.consumer.result.ErrorHandler;
 import pl.allegro.tech.hermes.consumers.consumer.result.SuccessHandler;
 import pl.allegro.tech.hermes.consumers.consumer.sender.MessageSender;
+import pl.allegro.tech.hermes.consumers.consumer.sender.MessageSenderFactory;
 import pl.allegro.tech.hermes.consumers.consumer.sender.MessageSendingResult;
 import pl.allegro.tech.hermes.consumers.consumer.sender.timeout.FutureAsyncTimeout;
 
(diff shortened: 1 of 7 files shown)


adamdubiel:
how does this work? looks like you just removed code responsible for restarting consumer on demand and moved the restart to each update (i.e. rate limit), which was not the point.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -80,8 +83,9 @@ public void sendMessage(final Message message) {
         }
     }
 
-    public void updateSubscription(Subscription newSubscription) {
+    public synchronized void updateSubscription(Subscription newSubscription) {
         this.subscription = newSubscription;
+        this.messageSender = messageSenderFactory.create(newSubscription);
```
wojtkiewicz: I think this is a good idea to rebuilt sending component but we need some logic that determines whether this is actually necessary. What you did will work but I think is a little bit too simplistic. I think there is a value in not disrupting delivery flaw if possible.

chemicL: Sure, I'll add the check whether the endpoint actually changed.

chemicL:
@adamdubiel so essentially no actual restart is happening anymore, just the endpoint information helper object is replaced. The check to determine whether the subscription's endpoint has actually changed is the missing part, I'll add it.
Regarding the other part of admin operations - handling retransmissions, that will require another PR.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
wojtkiewicz: You could also stop sender from consuming instead of just replacing...

wojtkiewicz: I think my comments made no sense because component that you replaced are responsible for single message delivery

wojtkiewicz:
:+1:

pbetkier:
:+1: 

While reviewing this code I noticed that we never call `MessageSender.close()`, both in the previous and in the new solution. I wonder if it doesn't cause resource leaks when updating subscriptions with JMS endpoints. JMS senders, unlike Jetty senders, do have some resources to clean up. We can either dive into this as part of this task or do it later, up to you :)

chemicL:
@pbetkier https://github.com/allegro/hermes/issues/399

Issue #194, opened 2015-08-26
Create a suspended subscription


Kornel:
I'd like to create a suspended subscription. Currently each created subscription is active by default.

mtrojanowski:
The PR for this one is ready ⬆️

druminski:
Thanks, although we don’t need this feature at the moment. We don’t have real use case for it but we can back to it in the future. At the moment closing this issue.

PR #406, opened 2016-03-16
Implementation of #194 Create a suspended subscription.

konrad-kaminski:


adamdubiel:
Looks simple, but i'm not sure it's what we need. An integration test would be nice.

konrad-kaminski:
The REST API already supports creating subscription with initial state set. We just needed to make to available on UI. Therefore I'd stick with this simple change in console.

hermes-console/static/partials/modal/editSubscription.html
```diff
@@ -45,6 +45,17 @@ <h3 class="modal-title" ng-show="operation === 'EDIT'"><small>Edit:</small> {{su
             </div>
 
             <div class="form-group">
+                <label for="subscriptionState" class="col-md-3 control-label">Initial state</label>
+                <div class="col-md-9">
+                    <select required class="form-control" id="subscriptionState" name="subscriptionState" ng-model="subscription.state">
+                        <option value="PENDING">PENDING</option>
```
wojtkiewicz: PENDING subscription means it exists but consumer is not spawned yet. Choosing such an option doesn't make much sense. I think choice should be either ACTIVE or SUSPENDED.

konrad-kaminski: Well, this is actually the default state of subscription upon creation right now. It is then automatically changed to ACTIVE (asynchronously).

adamdubiel: Small change, lot's of discussion. I think the only thing we should allow (or promote via Console) is a simple checkbox "create this as suspended". This way we don't need to have all this discussion and we don't make our client to think about this choice too much.

wojtkiewicz: Yes but up until now user had no choice. Default was to activate subscription and until it was done subscription had intermediate state: pending. Right now you ask user if he wants his subscription to have intermediate state... Thats why I think that choice should be to have either SUSPENDED (do nothing until activated) or ACTIVE (PENDING until activated) state... You ask user for a desired state in my opinion...

wojtkiewicz: Also please rebase your commit instead of adding merges...

adamdubiel:
This PR needs some serious rebasing against develop, not merging until it happens.

adamdubiel:
Closing for now - not resolved, not rebased etc. Better to close than to keep it hanging longer :)

PR #414, opened 2016-03-18
413 fix for batch receiver shutdown

wojtkiewicz:
Batch receiver is not shutdown properly which results in threads leaking


pawelszymczyk:
:+1:

chemicL:
:+1:

PR #418, opened 2016-03-18
417 shutting down client with a timeout to process inflight published…

chemicL:
… messages


hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/HermesClientTest.groovy
```diff
@@ -120,6 +123,42 @@ class HermesClientTest extends Specification {
...
+
+        when:
+        client.publish(TOPIC, CONTENT_TYPE, CONTENT)
+
+        and:
+        client.shutdown()
+        client.awaitShutdown(50, 5, TimeUnit.SECONDS)
+
+        then:
+        latch.await()
```
pbetkier: This test is going to hang indefinitely if some messages are not sent, right? How about calling `await` with some reasonable timeout? If it returns false then the test fails and we know it's because the condition wasn't met.

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/HermesClientTest.groovy
```diff
@@ -120,6 +123,42 @@ class HermesClientTest extends Specification {
...
+    }
+
```
pbetkier: A test specifying what happens if sending messages after shutdown fails would be nice.

pbetkier:
:+1:

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/HermesClientTest.groovy
chemicL: line 150 :-)

pbetkier:
In the previous comment regarding a new test I meant what happens if, after shutdown has started, sending of the trailing messages failed with exceptions instead of returning responses.

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/HermesClientTest.groovy
```diff
@@ -120,6 +125,58 @@ class HermesClientTest extends Specification {
...
+        when:
+        client.shutdown()
+        def future = client.publish(TOPIC, CONTENT_TYPE, CONTENT)
+
+        then:
+        future.completedExceptionally
+    }
+
+    def "should keep retrying on sender exception after shutdown"() {
```
chemicL: @pbetkier here it is :-)

chemicL:
@pbetkier Right, sorry I didn't get what you meant initially - I added a test case for that.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -67,16 +71,41 @@ private String createUri(URI uri) {
...
         return sender.send(URI.create(uri + message.getTopic()), message).exceptionally(HermesResponseBuilder::hermesFailureResponse);
     }
 
+    private CompletableFuture<HermesResponse> completedWithShutdownException() {
+        CompletableFuture<HermesResponse> alreadyShutdown = new CompletableFuture<>();
+        alreadyShutdown.completeExceptionally(new HermesClientShutdownException());
+        return alreadyShutdown;
+    }
+
+    public void shutdown() {
```
adamdubiel: shouldn't we make it `Closeable` instead of some new way to close this resource?

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClientTermination.java
```diff
@@ -0,0 +1,44 @@
...
+            while (!condition.evaluate()) {
+                Thread.sleep(pollInterval);
+            }
+        } catch (InterruptedException e) {
+        } finally {
+            latch.countDown();
+        }
+    }
+
+    interface Condition {
```
adamdubiel: can't we use some interfaces from Java 8?

adamdubiel:
i see no docs update

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -67,16 +71,41 @@ private String createUri(URI uri) {
...
+        shutdown = true;
+    }
+
+    /**
+     * @return whether all inflight published messages have completed before timeout
+     */
+    public boolean awaitShutdown(long pollIntervalMillis, long timeout, TimeUnit timeoutTimeUnit) {
```
adamdubiel: what if they are not? why are we returning `false` when not in shutdown mode - does it mean that they have not been sent? i don't get how to use this to be sure, that hermes client will gracefully shutdown

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClientTermination.java
chemicL: we could use `BooleanSupplier`, yep.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
chemicL: I thought of a combination of `shutdown` and `awaitShutdown`, as simply `Closeable` would imply that we're done after calling `close` - while in fact we're sending a command to shutdown.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
chemicL: Thanks for pointing this out, I'll remove thiss condition and simply await until the client is shutdown and the inflight counter is 0.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
adamdubiel: maybe make it closeable then:

```
close() # this will block until it closes with some default timeout
close(int timeout) # this will close in some timeout
CompletableFuture<> closeAsync(int timeout) # this will return future you can block on to close
```

adamdubiel:
:+1:

pbetkier:
:+1:

PR #419, opened 2016-03-19
support for Http Basic Auth in Consumers

adamdubiel:
BasicAuth support in Consumers module - based on credentials passed in URI, e.x `http://username:password@allegrogroup.com`.

Credentials are stored in plaintext in ZK for now, though they are never printed to user after saving them.


druminski:
:+1:

dankraw:
:+1: :sparkles:

PR #431, opened 2016-04-04
430 slow starting consumers monitoring

pawelszymczyk:


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/Timers.java
```diff
@@ -35,5 +35,7 @@
             CONSUMER_WORKLOAD_REBALANCE_DURATION = "consumers-workload." + KAFKA_CLUSTER + ".selective.rebalance-duration",
 
             EXECUTOR_DURATION = "executors." + EXECUTOR_NAME + ".duration",
-            EXECUTOR_WAITING = "executors." + EXECUTOR_NAME + ".waiting";
+            EXECUTOR_WAITING = "executors." + EXECUTOR_NAME + ".waiting",
+
+            CREATING_CONSUMER_LATENCY = "creating-latency";
```
adamdubiel: create-time? i would not call it latency

adamdubiel:
shouldn't we target some specific issue? like see if the problem is with with schema repo or sth, we can get this measurement by reading logs and it will have no value for us once we solve the *real* problem

pawelszymczyk:
Yes after we solve this problem it woldn't have real value. Main puropose of this timer is to see what is realy going on. It is some degradation, or just few outstanding slow cases?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/Timers.java
pawelszymczyk: ok

PR #435, opened 2016-04-08
312 message filtering

wojtkiewicz:
Initial version of message filtering for hermes-consumers.

**Scope**:
* registering subscription specific filters
* registering global filters
* registering custom filters
* monitoring filtered events

**Notable features**:
* support for AVRO (via [avpath](https://github.com/wandoulabs/avpath))
* support for JSON (via [jsonpath](https://github.com/jayway/JsonPath))
* filter chaining

**Choosing appropriate filter**:
This mainly concerns message content type. Filtering is done *before* any conversion takes place so all messages have the same content type as topic on which they were published. Not every filter has to dig into message content though.

| topic content-type  | filter type  |
|---|---|
| avro  | avropath  |
| json  | jsonpath  |

**Feature activation**:
Filtering component has to be enabled by setting:
`consumer.filtering.enabled = true`

Setting this to `false` will disable both global and subscription specific filters.

**Order of filtering**:
Filters run in the order that they are declared. Global filters will run before subscription specific.

**Global filters**:
Are enabled on application bootstrap and will run for every message for every subscription. Global filters cannot be enabled/disabled or inspected by subscribers. Global filters cannot be viewed via REST api.

**Subscription configuration**:
```
{  "name": "mySubscription", 
   "endpoint": "http://my-service", 
   "supportTeam": "My Team",
   "filters": [
      {"type": "jsonpath", path: "$.user.name", matcher: "^abc.*"},
      {"type": "jsonpath", path: "$.offer.type", matcher: "new"}  
   ]
}
```

**Subscription filter model**:
Filters don't have to conform to any schema. We only require `type` field to be declared. Rest of properties are filter specific (means each filter specification can have different set of properties).

**Docs**:
in progress

**CAUTION**:
this will be squashed soon


adamdubiel:
@sslavic @miguelpuyol i guess you might be interested :)

sslavic:
Will look more in code changes to find answers for following, but maybe you can answer or point to relevant commits/tests/examples/docs faster:

- How easy is it to configure and provision hermes-consumers with custom global filter logic, and what is the process to store/access filter specific subscription metadata?
- How can filter signal that message should neither be delivered nor skipped completely, but filter couldn't decide at that moment and delivery with filtering should be retried again e.g. applying subscription specific re-delivery strategy? How can this be monitored/observed?
- How does filtering work for new batch consumer, will entire batch be delivered/skipped/retried, or is it evaluated per message in batch?

Btw, while testing 0.8.6 we encountered few NPEs, in hermes management while managing groups, topics and subscriptions, resulting in stack traces in error responses but sometimes at least parts of requests would actually be successful; not sure if those have been covered in recent hot fixes; we'll retest and if we reproduce issues we'll report back.

wojtkiewicz:
@sslavic 
1) `HermesConsumersBuilder` has new method `withGlobalFilter(MessageFilter filter)` that you can use to plugin compiled filter. Subscription specific filters are stored the same way as subscription policy for instance.
2) You cannot do that. Filter either passes a message or not. If you need more complicated logic then perhaps you need more complex filter that has multiple stages hidden inside. Filter itself has the same logic a simple predicate.
3) Filtering is applied before batching. It is evaluated per message just after we read it from broker.

sslavic:
Thanks for heads up and all the hard work!

1) was hoping for non programatic, through configuration way to provision global filter class, but that should work too, thanks!
2) this is kind of limiting for us; as Adam was mentioning in earlier discussions one typical use case would be for filter to access a remote system to decide whether to skip or deliver, and if remote system is temporarily not accessible, what to do then? Blocking and retrying in filter itself could be one option, but that would be ugly, might result in inefficient use of resources (thread blocked waiting for filter to respond), and might get inconsistent with subscription delivery policy (e.g. filter could block longer than subscription allows message retention); what happens if filter throws unexpected runtime exception and is there a test for that?
3) sounds good, thanks!

adamdubiel:
@sslavic i think we need to do some work on just adding jars on classpath which would "click" with the rest of Hermes. However this requires some serious work, so don't hold your breath just yet :) I just wanted to let you know that i notice how inconvenient current setup is, as we run 3 clusters of Hermes internally already and creating separate project for each is a pain.

As to the 2nd point: i was aware of the use case (contacting remote), but i didn't think you would want to "skip" messages and come back to them later. Kafka offsets are linear, meaning even if we kept some messages for further filtering, messages already acknowledged by the consumers could not be committed. This is however just a tradeoff you might be willing to take. The solution would be to keep messages with some *filter later* status in memory buffer and every now and then try to run them through filters again and again.

I think we could compromise by adding a *filter later* response to `MessageFilter` interface (making it three-state instead of just boolean) so that you could create a PR with some pluggable handling of those messages. How about that?

adamdubiel:
As to the NPEs you observed - could you give us some specifics? `0.8.6-hotfix1` fixes some management related stuff, but it might not be all.

wojtkiewicz:
@sslavic @adamdubiel 
2) I think that solution really depends on what problem are you trying to solve. If you don't want any message to come out of your system without running your security checks then I'm afraid that buffering messages while external service is unavailable solves absolutely nothing. In my opinion what you should be doing in case of failure is to lower your consumption rate to 0 (perhaps only on topic that requires that external service) and increase consumption rate when service is available again. This is very similar to how we handle failures of our subscribers. We are lowering consumption rate automatically and we detect when they are fully operational (via heartbeat mode).

wojtkiewicz:
Filtering is implemented via FilteringMessageReceiver that has all the tools required for controlling consumption rate so this possible to implement but in the first iteration of filtering implementation we focused on addressing requirements that concern filtering by analysing message content and not via external calls to some remote service.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MessageFilterSpecification.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class MessageFilterSpecification {
+    private String type;
+    private Map<String, Object> spec;
+
+    public MessageFilterSpecification() {
+    }
+
+    @JsonCreator
+    public MessageFilterSpecification(Map<String, Object> spec) {
```
druminski: Since you are using @JsonCreator for this constructor, the default one can be deleted, also declared fields in this class can be final as it is immutable class.

hermes-client/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     testCompile group: 'org.spockframework', name: 'spock-core', version: versions.spock
     testCompile group: 'com.github.tomakehurst', name: 'wiremock', version: versions.wiremock
     testCompile group: 'javax.servlet', name: 'javax.servlet-api', version: '3.1.0'
+    testCompile group: 'com.jayway.jsonpath', name: 'json-path', version: '2.2.0'
```
druminski: you don't need this dependency here

hermes-consumers/build.gradle
```diff
@@ -10,12 +10,14 @@ dependencies {
     compile group: 'org.eclipse.jetty', name: 'jetty-client', version: '9.3.6.v20151106'
 
     compile group: 'javax.jms', name: 'javax.jms-api', version: '2.0'
-    compile group: 'com.jayway.jsonpath', name: 'json-path', version: '0.9.1'
+    compile group: 'com.jayway.jsonpath', name: 'json-path', version: '2.2.0'
```
druminski: you have already defined this dependency in hermes-common, therefore you don't need to add it here again

hermes-consumers/build.gradle
```diff
@@ -10,12 +10,14 @@ dependencies {
...
     compile group: 'joda-time', name: 'joda-time', version: '2.5'
 
     compile group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.1'
     compile group: 'com.github.rholder', name: 'guava-retrying', version: '2.0.0'
 
+    compile group: 'com.wandoulabs.avro', name: 'wandou-avpath_2.11', version: '0.1.0'
```
druminski: I am wondering why you have added wandou-avpath_2.11 here but json-path to hermes-common module, maybe json-path also should be added only here

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/FilteringMessageReceiver.java
```diff
@@ -0,0 +1,65 @@
...
+        FilterResult result = filterChain.apply(message);
+        filteredMessageHandler.handle(result, message, subscription);
+        return result.filtered;
+    }
+
+    @Override
+    public Message next() {
+        Message message;
+        do {
+            updateFilter();
```
druminski: I don't understand why you verify if filterChain needs to be updated with every message read. Maybe filterChain should be volatile.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/chain/FilterChain.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.List;
+
+public class FilterChain {
+
+    public FilterChain(final List<MessageFilter> messageFilters) {
+        this.messageFilters = messageFilters;
+    }
+
+    private List<MessageFilter> messageFilters;
```
druminski: All immutable classes in hermes have got fields declared at the class beginning with a final word. Imho it would be nice to keep that consistency.

integration/src/integration/java/pl/allegro/tech/hermes/integration/MessageBufferLoadingTest.java
```diff
@@ -70,7 +70,8 @@ public void tearDown() throws IOException {
         FileUtils.deleteDirectory(tempDir);
     }
 
-    @Test
+
+    @Test(enabled = false)
     public void shouldBackupMessage() throws Exception {
```
druminski: Name of this test suggests that it is rather important. I think it would be nice to have it enabled.

druminski:
@Kornel @wojtkiewicz I have added few small comments to the code but in general nice PR.

sslavic:
In FilterResult class would you consider replacing message with full Exception reference? message can be null in Exception while Exception could have caused message to be filtered out and it would be impossible to differentiate that case from case where message was successfully filtered out without exception thrown.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MessageFilterSpecification.java
wojtkiewicz: done

hermes-client/build.gradle
wojtkiewicz: wiremock has dependency on json-path and by default that version cause some issues on our classpath

hermes-consumers/build.gradle
wojtkiewicz: done

hermes-consumers/build.gradle
wojtkiewicz: not sure, I'm checking what is going on

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/FilteringMessageReceiver.java
wojtkiewicz: I will work on this code cause I think it can be more readable but general idea is not to update with every message read (there is a condition hidden in updateFilter) but if we have to update then we're doing that from the same thread that is going to read actual value later.

hermes-consumers/build.gradle
wojtkiewicz: removed from common cause it made no sense

hermes-consumers/build.gradle
wojtkiewicz: leaving dependency in consumer, jsonpth from hermes-test-helper was causing all sorts of trouble in all modules

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/chain/FilterChain.java
wojtkiewicz: done

integration/src/integration/java/pl/allegro/tech/hermes/integration/MessageBufferLoadingTest.java
wojtkiewicz: sure

wojtkiewicz:
@sslavic done

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/FilteringMessageReceiver.java
wojtkiewicz: I simplified this algorithm cause I think we're better off with writing from one thread only.

druminski:
:+1: @wojtkiewicz please enable shouldBackupMessage() test at the end of your work in this pr.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/SerialConsumer.java
```diff
@@ -141,6 +142,9 @@ public Subscription getSubscription() {
     public void updateSubscription(Subscription newSubscription) {
         rateLimiter.updateSubscription(newSubscription);
         sender.updateSubscription(newSubscription);
+        if (messageReceiver != null) {
```
chemicL: Perhaps for cleaner code, we could check whether this consumer is consuming from the outside when updating subscription and avoid null checks. Also, the private method `startConsuming` can use the field instead of a parameter.

wojtkiewicz: Heh this was some dirty and unexpected fix after rebasing on develop. I'll fix this right away.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/FilteredMessageHandler.java
```diff
@@ -0,0 +1,44 @@
...
+            updateMetrics(message, subscription);
+            trackers.get(subscription).logFiltered(toMessageMetadata(message, subscription), result.getFilterType().get());
+        }
+    }
+
+    protected void updateMetrics(Message message, Subscription subscription) {
+        metrics.meter(Meters.FILTERED_METER).mark();
+        metrics.counter(Counters.FILTERED, subscription.getTopicName(), subscription.getName()).inc();
+        metrics.decrementInflightCounter(subscription);
+        metrics.inflightTimeHistogram(subscription).update(System.currentTimeMillis() - message.getReadingTimestamp());
```
chemicL: Should filtered out messages influence the inflight time? In case of filtering most of the messages this will greatly lower the expected inflight time.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/FilteringException.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public FilteringException(String message) {
+        super(message);
+    }
+
+    public FilteringException(String message, Throwable throwable) {
+        super(message, throwable);
+    }
+
+    public static void assertTrue(boolean condition, String message) {
```
chemicL: Would you consider calling this `throwIfNot`? I got confused what would happen when I saw this assert used with static import.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/chain/FilterChain.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    public FilterResult apply(final Message message) {
+        for (MessageFilter filter : messageFilters) {
+            try {
+                if (!filter.test(message)) {
+                    return FilterResult.failed(filter.getType(), "logical");
+                }
+            } catch (Exception ex) {
+                return FilterResult.failed(filter.getType(), ex.getMessage());
```
chemicL: The full exception could be passed, as @sslavic suggested IIRC

chemicL:
👍 but I'd have a look at the inflight time metric to exclude filtered out time.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/SerialConsumer.java
wojtkiewicz: I worked on subscription updates a little bit.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/FilteredMessageHandler.java
wojtkiewicz: I think it's ok and that's representing what's going on on this particular subscription. If we filter out 90% of messages then we would expect to spend very little time sending messages to this subscriber.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/FilteringException.java
wojtkiewicz: done

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/chain/FilterChain.java
wojtkiewicz: done

wojtkiewicz:
@chemicL @druminski @sslavic thanks for your comments, take a look and if everything is all right I'll merge this tomorrow

sslavic:
Looks good to me, ship it!

Is there anything else planned for 0.8.7 but still not finished, i.e. when can we expect 0.8.7 to be released?

wojtkiewicz:
@sslavic we've encountered few bugs with 0.8.6 on our production and we're focused on fixing those first (these bugs has nothing to do with message filtering).

PR #439, opened 2016-04-14
438 request timeout lock

pawelszymczyk:


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/MessagePublisher.java
```diff
@@ -22,9 +23,11 @@ public MessagePublisher(BrokerMessageProducer brokerMessageProducer) {
...
     }
 
-    public void publish(Message message, Topic topic, MessageState messageState, BrokerListeners listeners, PublishingCallback callback) {
-        messageState.setState(SENDING_TO_KAFKA_PRODUCER_QUEUE);
-        brokerMessageProducer.send(message, topic, callback);
-        messageState.setState(SENDING_TO_KAFKA);
+    public void publish(Message message, Topic topic, MessageState messageState, RequestTimeoutLock requestTimeoutLock, BrokerListeners listeners, PublishingCallback callback) {
+        if (requestTimeoutLock.tryLock()) {
+            messageState.setState(SENDING_TO_KAFKA_PRODUCER_QUEUE);
+            brokerMessageProducer.send(message, topic, callback);
```
druminski: if send() throws an exception (for example could not fetch topic metadata) after 65ms then client receives 202 but message will not be send to kafka, only persisted in backup storage.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/MessagePublisher.java
```diff
@@ -22,9 +23,12 @@ public MessagePublisher(BrokerMessageProducer brokerMessageProducer) {
...
+            messageState.setState(SENDING_TO_KAFKA);
+            requestTimeoutLock.unlock();
```
chemicL: the unlock should be called in a finally block

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/message/RequestTimeoutLock.java
```diff
@@ -0,0 +1,21 @@
+package pl.allegro.tech.hermes.frontend.publishing.message;
+
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class RequestTimeoutLock {
```
chemicL: Not sure if this class is necessary, it just delegates everything to the ReentrantLock - a meaningful variable name should be sufficient.

druminski:
Ok, so this pr can break 65ms barrier and can return 202 response much later instead of 408 response at 65ms. Although, Delayed 202 should be very rare, for example when kafka topic metadata are fetched during first event on topic, therefore I like where this pr is going. However, we should test it very well and verify how this change impacts on performance, because I am not sure about locking in undertow listeners. Also I think this change should go to develop branch as normal issue, not as hotfix.

pawelszymczyk:
It could be at most 65ms +  KAFKA_CONSUMER_METADATA_READ_TIMEOUT which is 5000ms by default, then Kafka connector throws exception and lock will be released.

druminski:
@pawelszymczyk actually, KAFKA_PRODUCER_METADATA_FETCH_TIMEOUT_MS is taken, its default value is 500ms.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/TimeoutAsyncListener.java
```diff
@@ -26,9 +29,15 @@ public void onComplete(AsyncEvent event) throws IOException {
 
     @Override
     public void onTimeout(AsyncEvent event) throws IOException {
-        if (!messageState.wasDelegatedToKafka()) {
+        requestTimeoutLock.lock();
+        if (messageState.equals(MessageState.State.WAITING_FOR_FIRST_PORTION_OF_DATA) ||
+                messageState.equals(MessageState.State.PARSING) ||
+                messageState.equals(MessageState.State.PARSED)) {
             httpResponder.timeout(event.getThrowable());
+        } else if (messageState.equals(MessageState.State.SENDING_TO_KAFKA)) {
```
druminski: This else if block is duplicated in BrokerTimeoutAsyncListener but without lock usage. This need to be fixed as we don't want to end of exchange in two places.

adamdubiel:
waiting for performance tests - i will close this PR since it's just hanging

PR #450, opened 2016-04-16
Kafka consumers rebalance

druminski:
Hermes does not configure max number of rebalance retries for kafka consumers. Default value is 4. I have noticed that when `kafka.common.ConsumerRebalanceFailedException: ... can't rebalance after 4 retries` is thrown then some topic partitions can be blocked and messages from them are not read. 

This is one of the reason why we should migrate to 0.9 kafka-consumer. In the meantime I propose this PR which increase max number of rebalance retries and decrease backoff time between retries from 20s to 4s, consequently it should speed up kafka-consumer startup when first rebalance fails.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
@@ -13,7 +13,6 @@ public enum Configs {
     ZOOKEEPER_BASE_SLEEP_TIME("zookeeper.base.sleep.time", 1000),
     ZOOKEEPER_CONNECTION_TIMEOUT("zookeeper.connection.timeout", 10000),
     ZOOKEEPER_SESSION_TIMEOUT("zookeeper.session.timeout", 10000),
-    ZOOKEEPER_SYNC_TIME("zookeeper.sync.time", 20000),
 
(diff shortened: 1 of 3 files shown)


dankraw:
👍

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -13,7 +13,6 @@
     ZOOKEEPER_BASE_SLEEP_TIME("zookeeper.base.sleep.time", 1000),
     ZOOKEEPER_CONNECTION_TIMEOUT("zookeeper.connection.timeout", 10000),
     ZOOKEEPER_SESSION_TIMEOUT("zookeeper.session.timeout", 10000),
-    ZOOKEEPER_SYNC_TIME("zookeeper.sync.time", 20000),
```
chemicL: The default is 2000 - won't this affect performance?

chemicL:
👍

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
druminski: Previous value was set to 20k, current is 4k but there will be more retries. We will see how it performs, finally we can change those values.

Issue #441, opened 2016-04-14
Hermes does not support Avro messages in BackupMessagesLoader.


tmmi:

tmmi:
It works, but code is a little misleading.

PR #454, opened 2016-04-19
#441 | Support Avro in backupLoader

tmmi:


druminski:
Hermes already supports loading avro messages from backup storage, however JsonMessage class name is misleading in BackupMessageLoader because it holds also avro blobs. Maybe we should change the name of the class to something different, but we can easily do this with another occasion in normal release flow, not as hotfix. Closing the issue.

PR #455, opened 2016-04-19
449 schema fetch with try-harding

wojtkiewicz:
This is initial version of schema fetching with try-harding algorithm that first checks for offline candidates and after giving up is trying to find them online. We should probably rate limit online calls but please take a look on the initial draft.


druminski:
👍 LGTM

pbetkier:
👍

tmmi:
👍

chemicL:
👍

Issue #372, opened 2016-02-20
Broadcast mode in subscription


adamdubiel:
Allow `EndpointResolver` to return multiple endpoints. If subscription has **broadcast** mode on, the messages will be broadcasted to all resolved addresses. Otherwise only 1st address is used.

We need to figure out retransmissions, either:

* no retransmissions (probably easier)
* retransmit until all instances ack

rafal-glowinski:
Yes yes! Pretty please :) I vote for 2nd one.

adamdubiel:
This change will involve changes in [SerialConsumer](https://github.com/allegro/hermes/blob/d7dba18c3e5f21a07cdff493d278e376f5ece4fe/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/SerialConsumer.java) and [EndpointAddressResolver](https://github.com/allegro/hermes/blob/28ec77ad48e05dd9f75662d74f1f9630ec11b253/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/resolver/EndpointAddressResolver.java).

The easy part is making `EndpointAddressResolver` return list of addresses instead of just one (although remember to keep the backwards compatibility of this interface, probably some default method will be helpful). The hard part is making one-to-many relationship work in sender and retries. Message can't be committed unless it has been ackd by all consuming nodes.

PR #461, opened 2016-04-26
#372 | Broadcast mode in subscription

sbolimowski:


hermes-console/static/partials/subscription.html
```diff
@@ -111,6 +111,9 @@ <h3 class="panel-title">Properties</h3>
                         <strong>Delivery type:</strong> {{subscription.deliveryType}}
                         <span uib-popover='Hermes can deliver messages in SERIAL (one message at a time) or in BATCH (group of messages at a time).' popover-trigger="mouseenter" class="fa helpme pull-right">&#xf128;</span>
                     </p>
+                    <p>
+                        <strong>Mode:</strong> {{subscription.mode}}
```
adamdubiel: can we add a popover with explaination of mode?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -131,8 +133,12 @@ private boolean shouldReduceSendingRate(MessageSendingResult result) {
         return !result.isRetryLater() && subscriptionAllowsResending(result);
     }
 
+    private boolean messageSentSucceeded(MessageSendingResult result) {
+        return result.succeeded() || (result.isClientError() && !subscription.getSerialSubscriptionPolicy().isRetryClientErrors());
+    }
+
     private boolean subscriptionAllowsResending(MessageSendingResult result) {
```
adamdubiel: i dont quite get the naming here: so subscription allows resending if sending failed?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Message.java
```diff
@@ -82,8 +86,9 @@ public boolean isTtlExceeded(long ttlMillis) {
         return currentTimestamp > readingTimestamp + ttlMillis;
     }
 
-    public void incrementRetryCounter() {
+    public void incrementRetryCounter(Collection<String> succeededUris) {
         this.retryCounter++;
+        this.succeededUris = ImmutableSet.copyOf(succeededUris);
```
adamdubiel: shouldn't this be an append operation? and i don't know if we want to create new collection for each retry, think about our poor allocation rate :) maybe appending to the set would be enough, sacrificing the internal immutability?

hermes-console/static/partials/subscription.html
sbolimowski: done

adamdubiel:
👍

wojtkiewicz:
👍

PR #474, opened 2016-05-04
Removing inactive subscription assignment entry from zk

### LINKED ISSUE #471: Subscription zk consumer-assignment parent entry not removed after subscription removal
This leaves a lot of garbage and makes assignment logs misleading.

dankraw:
fixes #471


adamdubiel:
👍

wojtkiewicz:
👍

PR #483, opened 2016-05-12
Stale backup files

druminski:
PR contains two unrelated commits with separate functionality:

1. loading backup messages from multiple stale backup files
2. waiting on topic metadata before backup messages are send to kafka-producer. There is no point to send them when kafka nodes are not available in frontend startup phase, because they will not get to internal kafka producer buffer and they will not be send to kafka. Therefore safer way is to wait on broker topic metadata and after that send backup messages.

### DIFF
--- a/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/buffer/BackupMessagesLoader.java
+++ b/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/buffer/BackupMessagesLoader.java
@@ -25,8 +25,10 @@ import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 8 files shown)


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/buffer/BackupMessagesLoader.java
```diff
@@ -141,6 +160,30 @@ private boolean sendMessageIfNeeded(Message message, Optional<Topic> topic, Stri
         }
     }
 
+    private void waitOnBrokerTopicMetadata(Topic topic) {
```
adamdubiel: can we get an infinite loop here? should we monitor this somehow (metrics etc)?

druminski: This code is run as the last startup hook after frontend is started. It is also run from the main thread, so it will be blocked until kafka is alive. I don't see any strong args right now why we should do this in another thread but I am open for them.

As for monitoring, this is not the way how we would like to monitor whether kafka is alive. When it is down then we will see logs from here in every 5s (default value) which informs that backup message loader waits for kafka to be responsive.

adamdubiel: i think we could at least add some log messages like "Topic {} is not available, checked {} times" i "Topic loaded"? this only happens on startup after all and might give us some hint of what is going on in case we see frontend hanging

druminski: Ok, I have added counter to log message.

integration/src/integration/java/pl/allegro/tech/hermes/integration/MessageBufferLoadingTest.java
```diff
@@ -71,7 +71,7 @@ public void tearDown() throws IOException {
     }
 
 
-    @Test
+    @Test(enabled = false)
```
adamdubiel: :(

adamdubiel:
👍

integration/src/integration/java/pl/allegro/tech/hermes/integration/MessageBufferLoadingTest.java
druminski: enabled again

pawelszymczyk:
👍

wojtkiewicz:
👍

PR #484, opened 2016-05-12
359 oauth2

Saal89:
Czekam jeszcze na https://github.com/scribejava/scribejava/pull/659. Jak na razie spreparowany jar jest w folderze hermes-consumers


hermes-api/src/main/java/pl/allegro/tech/hermes/api/EndpointAddress.java
```diff
@@ -17,28 +17,14 @@
 @JsonSerialize(using = EndpointAddressSerializer.class)
 public class EndpointAddress {
 
-    private static final String ANONYMIZED_PASSWORD = "*****";
```
adamdubiel: we can't lose backwards compatibility - we need to support both url-like credentials and this separate sturcture you created

wojtkiewicz:
Please rebase your commits and remove explicit merges

hermes-api/src/main/java/pl/allegro/tech/hermes/api/EndpointAddress.java
Saal89: Basic authorization is still there, but it is not written inside url. I have added select option "Authorization type" (3 options: NONE, BASIC, OAUTH2). When you choose BASIC, the username and password fields are displayed. So functionality remains.
Still you want to have it in the url?

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/BasicAuthProviderTest.java
```diff
@@ -0,0 +1,36 @@
...
+import pl.allegro.tech.hermes.api.DeliveryType;
+import pl.allegro.tech.hermes.api.EndpointAddress;
+import pl.allegro.tech.hermes.api.Subscription;
+import pl.allegro.tech.hermes.consumers.consumer.Message;
+import pl.allegro.tech.hermes.consumers.test.MessageBuilder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class BasicAuthProviderTest {
```
adamdubiel: why did you rewrite it to java? we try to write new stuff in groovy (per contributors guide)

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyMessageSenderTest.java
```diff
@@ -152,7 +163,14 @@ public void shouldSendRetryCounterInHeader() {
     @Test
     public void shouldSendAuthorizationHeaderIfAuthorizationProviderAttached() {
         // given
-        JettyMessageSender messageSender = new JettyMessageSender(client, address, Optional.of(m -> "Basic Hello!"), 1000, new DefaultHttpMetadataAppender());
+        HashMap<String,String> authData = new HashMap<String,String>();
```
adamdubiel: i don't need to create subscription here -> why not stay with simple function that just returns fixed text? looks like you try to test some other stuff here

hermes-api/src/main/java/pl/allegro/tech/hermes/api/EndpointAddress.java
adamdubiel: i know the functionality remains as a whole, but i want to deploy this on prod (and probably other clients as well) without breaking existing stuff, and we have some basic-auth subs

Saal89: Ok, I understand.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        service = new ServiceBuilder()
+                .apiKey(authData.getConsumerKey())
+                .apiSecret(authData.getConsumerSecret())
+                .build(new OAuthApi(authData.getAccessTokenEndpoint()));
+        obtainAccessToken();
+    }
+
+    @Override
+    public String authorizationToken(Message message) {
+        if (message.getNumberOfUnauthorizedRequests() == 1) {
```
adamdubiel: i think setting this to 1 might be dangerous - why 1? is it not to repeat authorization if it fails once? what if i have 10 msgs and all came back unathorized (in parallel) - do you want to send separate requests for all of them?

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyMessageSenderTest.java
Saal89: Yes, I want to ensure that provider access token is equal to authorization header

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import com.github.scribejava.core.oauth.OAuth20Service;
+import com.github.scribejava.core.model.OAuthConfig;
+
+public class OAuth2AuthProvider implements HttpAuthorizationProvider {
+
+    private final OAuth2AuthenticationData authData;
+
+    private final OAuth20Service service;
+
+    private String accessToken;
```
adamdubiel: i think we lack some thread safety here, if we get 10 requests for new token out, all of them will be modifying this field, while others still use it to send messages

hermes-api/src/main/java/pl/allegro/tech/hermes/api/AuthenticationData.java
```diff
@@ -0,0 +1,9 @@
+package pl.allegro.tech.hermes.api;
+
+public interface AuthenticationData {
+
+    public String getUsername();
```
adamdubiel: no need for `public` in interface, also shouldn't `anoymize()` be a part of this interface?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/AuthenticationType.java
```diff
@@ -0,0 +1,7 @@
+package pl.allegro.tech.hermes.api;
+
+public enum AuthenticationType {
```
adamdubiel: i don't think it's authentication, but authorization in general (in all class names)

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
Saal89: I' ll think about some better solution

adamdubiel:
Some remarks:

* I think we lack any E2E test (in integration module). 
* Did you run some traffic through it? Like at least 40-50rps to see how it behaves in corner cases (need to reauthenticate etc)?
* Do we need to add refresh token support?

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyMessageSenderTest.java
adamdubiel: but it was already tested - provider returned "Hello world" and that made it to the header

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import pl.allegro.tech.hermes.consumers.consumer.Message;
+
+import com.github.scribejava.core.builder.api.DefaultApi20;
+
+import com.github.scribejava.core.builder.ServiceBuilder;
+import com.github.scribejava.core.model.OAuth2AccessToken;
+import com.github.scribejava.core.oauth.OAuth20Service;
+import com.github.scribejava.core.model.OAuthConfig;
+
+public class OAuth2AuthProvider implements HttpAuthorizationProvider {
```
adamdubiel: i think we could use some metrics here - so we know how does the auth service behave, what are the response times etc, might come in handy :)

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -35,6 +35,15 @@
     private EndpointAddress endpoint;
 
     @NotNull
+    private AuthenticationType authenticationType = AuthenticationType.NONE;
```
wojtkiewicz: I think it would be a good idea to merge all authentication related fields into `AuthenticationDetails` or something similar...

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
```diff
@@ -0,0 +1,65 @@
...
+    @Override
+    public String authorizationToken(Message message) {
+        if (message.getNumberOfUnauthorizedRequests() == 1) {
+            obtainAccessToken();
+        }
+        return accessToken;
+    }
+
+    public String obtainAccessToken() {
+        OAuth2AccessToken token = service.getAccessTokenPasswordGrant(authData.getUsername(), authData.getPassword());
```
adamdubiel: what are fail scenarios here? how to deal with them? we need at least some explicit logs

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
adamdubiel: i wonder if it is a good idea to run this blocking code in Consumer thread.. this is also in regard to "what to do when all messages need to authorize" comment below. I think we could go with sth like separate thread pool to execute oauth requests, isolating it form main consumer loop, and this would "communicate" via some thread safe collection (ConcurrentHashMap?) with the sender itself. Also we need to think about some rate limiting for OAuth service - we deal with some pretty serious traffic here, i don't think OAuth gateways are prepared to get hit by hundreds of rps.

adamdubiel:
Thanks for the PR :) There are some things to fix, but looks like we might be getting oauth2 support!

@sslavic you might be interested as well - could you take a look at this?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -157,14 +157,22 @@ public void accept(MessageSendingResult result) {
                 rateLimiter.registerSuccessfulSending();
                 handleMessageSendingSuccess(message, result);
             } else {
-                handleFailedSending(message, result);
-                message.incrementRetryCounter();
+                if (result.isUnauthorized()) {
+                    message.incrementNumberOfUnauthorizedRequests();
+                }
```
wojtkiewicz: 1) move this stuff to `handleFailedSending`
2) don't send messages from the same thread that handles responses

wojtkiewicz: This code is really weird... What is your approach for handling unauthorized requests? Should we retry or not? Should we discard messages? Should we stop and reduce outgoing rate?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
wojtkiewicz: Generally speaking if someone requires check for every request then OAuth gateway has to handle our outgoing traffic. What I'm concerned with is how we can handle failure of such service. I think that good idea is to reduce consumer rate to heartbeat and wait till its available again.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        @Override
+        public String getAccessTokenEndpoint() {
+            return accessTokenEndpoint;
+        }
+
+        @Override
+        public String getAuthorizationUrl(OAuthConfig config) {
+            return null;
```
wojtkiewicz: ?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
wojtkiewicz: Consumer is single-threaded and is responsible for reading message and sending payload without waiting for response. Anything that goes between reading message and sending payload can be blocking. There is not much to do until token is authorised. We could have separate connection pool per OAuth gateway though. Not sure what is going on in ScribeJava. We have to read what that library is doing before merging this.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
wojtkiewicz: We have to read what goes underneath in this library before we go any further cause code in this layer looks a little naive. Exceptions from scribe would flood whatever layer this code is called on.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/OAuth2AuthProvider.java
```diff
@@ -0,0 +1,65 @@
...
+    private String accessToken;
+
+    public OAuth2AuthProvider(Subscription subscription) {
+        this.authData = subscription.getOAuth2AuthenticationData();
+
+        service = new ServiceBuilder()
+                .apiKey(authData.getConsumerKey())
+                .apiSecret(authData.getConsumerSecret())
+                .build(new OAuthApi(authData.getAccessTokenEndpoint()));
+        obtainAccessToken();
```
wojtkiewicz: You can't make online calls from this constructor. Have you checked which thread is running this code and when? You can only make online and blocking calls on consumer thread (if any blocking calls at all).

wojtkiewicz: If you want to do such things then do it when consumer is bootstrapping in its own thread. Then we have some chance to come up reliable solution how to handle gateway errors.

wojtkiewicz:
@adamdubiel @Saal89 There is some wiring going on with scribe (which is useful) but what we lack here is some general strategy how to handle gateway failures and unauthorised messages. We need also some strategy how to manage resource because we are potentially calling different gateways per subscription (or at least we are generating tokens per subscription). Right now we do whatever scribe does (whether it makes sense for us or not).

adamdubiel:
Closing after consultation with @Saal89 - we might get back to this at some later time.

PR #486, opened 2016-05-13
Backup messages only for 202

### LINKED ISSUE #485: Prevent OutOfMemory when broker is unavailable

druminski:
This is a minimalistic PR fixing memory leak problem when kafka is down.

Currently when kafka is down frontend persists all messages in backup storage. Backup storage allocates memory off heap. So it quickly can lead to no free memory on the machine. This fix does not persist messages in backup storage when there is no free space in internal kafka producer buffer. For this situation 5xx status code is returned. 

With this fix message will be only persisted in backup storage when 202 status code is returned.


adamdubiel:
👍 i would feel safer having some e2e test, but it understand it could be very costly

pawelszymczyk:
👍

wojtkiewicz:
👍

PR #490, opened 2016-05-25
489 replaced avropath with simple avro filter

wojtkiewicz:
This PR removed usage of avpath library because it seems to no longer be maintained. As a replacement we have simple selector that understands paths like ".foo.bar". Matching over collections and any sort of complex querying is not supported at the moment. With that being said this filter supports all cases that we encountered in production at the moment.

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/avro/AvroPathPredicate.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/avro/AvroPathPredicate.java
@@ -1,30 +1,33 @@
 package pl.allegro.tech.hermes.consumers.consumer.filtering.avro;
 
 import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.lang.StringUtils;
 import pl.allegro.tech.hermes.api.ContentType;
 import pl.allegro.tech.hermes.consumers.consumer.Message;
 import pl.allegro.tech.hermes.consumers.consumer.filtering.FilteringException;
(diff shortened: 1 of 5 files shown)


pbetkier:
👍

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/avro/AvroPathPredicate.java
```diff
@@ -1,62 +1,65 @@
...
-    private List<Evaluator.Ctx> selectNodes(final Message message) throws IOException {
+    private Optional<Object> select(final Message message) throws IOException {
         CompiledSchema<Schema> compiledSchema = message.<Schema>getSchema().get();
-        return Evaluator.select(bytesToRecord(message.getData(), compiledSchema.getSchema()), pathSyntax);
+        return select(bytesToRecord(message.getData(), compiledSchema.getSchema()));
     }
 
-    private boolean matches(final List<Evaluator.Ctx> nodes) {
-        scala.collection.Iterator<Evaluator.Ctx> iter = nodes.iterator();
+    public Optional<Object> select(GenericRecord record) {
```
dankraw: does this has to be public?

dankraw:
👍

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/avro/AvroPathPredicate.java
wojtkiewicz: no :-), sry

wojtkiewicz: fixed

PR #492, opened 2016-05-27
Subscription message-independent headers

### LINKED ISSUE #491: Add custom message-independent headers for subscribers
Hermes should be able to pass some headers that are message-independent to subscribers and are configured per subscription.

dankraw:
resolves #491


dankraw:
I'll move this to develop branch.

PR #507, opened 2016-06-14
Update publishing docs: creating groups and topics

pbetkier:
Creating groups and topics was a bit outdated for a couple of months.


pbetkier:
This update presents how things work right now, not necessarily how things *should* work. Is it alright that the user has to provide retention time to create a topic? Was it introduced by mistake @adamdubiel?

chemicL:
👍

adamdubiel:
it should not work like this, but 👍 for doc change

PR #510, opened 2016-06-16
508 Abnormal delivery do not log throwable reason

pszymczyk:


adamdubiel:
👍

adamdubiel:
I added a way to say that i don't want my exception to cause rate drop when resolving hosts.

druminski:
👍

PR #521, opened 2016-07-01
Subscription metadata visibility

dankraw:
Fixing subscription metadata visibility in the console: available metadata entries will always be visible, even when not set for subscription.


adamdubiel:
👍

pbetkier:
👍

PR #550, opened 2016-08-01
488 added expire_after_access to CachedCompiledSchemaRepository

wojtkiewicz:
Added expire-after-access to second level schema repository cache. Main idea is that cache is self maintaining and invalidates entries that were not used for *n* minutes/hours/days. This should reduce memory footprint for our dev/test environment and also partially support workflow where user wants to  delete entire topic with all its schemas and replace it with topic with the same name. In this solution on  *topic reboot* it is required to wait until all caches invalidate. Previously we had to reset entire module to clear caches with compiled schemas.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
@@ -159,6 +159,7 @@ public enum Configs {
 
     SCHEMA_CACHE_REFRESH_AFTER_WRITE_MINUTES("schema.cache.refresh.after.write.minutes", 10),
     SCHEMA_CACHE_EXPIRE_AFTER_WRITE_MINUTES("schema.cache.expire.after.write.minutes", 60 * 24),
+    SCHEMA_CACHE_COMPILED_EXPIRE_AFTER_ACCESS_MINUTES("schema.cache.compiled.expire.after.access.minutes", 60 * 48),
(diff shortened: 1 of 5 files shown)


druminski:
👍

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -159,6 +159,7 @@
 
     SCHEMA_CACHE_REFRESH_AFTER_WRITE_MINUTES("schema.cache.refresh.after.write.minutes", 10),
     SCHEMA_CACHE_EXPIRE_AFTER_WRITE_MINUTES("schema.cache.expire.after.write.minutes", 60 * 24),
+    SCHEMA_CACHE_EXPIRE_AFTER_ACCESS_MINUTES("schema.cache.expire.after.access.minutes", 60 * 48),
```
pbetkier: I'd say it should be `SCHEMA_CACHE_COMPILED_EXPIRE_AFTER_ACCESS_MINUTES` (near existing `SCHEMA_CACHE_COMPILED_MAXIMUM_SIZE `), because it's a different cache than the one configured by upper properties.

pbetkier:
👍

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
wojtkiewicz: ok, good idea

wojtkiewicz: done

PR #554, opened 2016-08-04
fix batch messages double wrapping

adamdubiel:
This is a bug with double wrapping of batched messages if they happen to end up in inflight queue.


wojtkiewicz:
👍

pbetkier:
👍

PR #565, opened 2016-08-29
Subsciption request timeout max value increased to 5 min

dankraw:
Consumers request timeout max value increased from 1 to 5 mins.


wojtkiewicz:
👍

druminski:
👍

PR #582, opened 2016-09-21
566 anonymizing data in auditor

aihnat:
Auditing OAuthProvider operations also added.


hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -17,7 +17,7 @@
 
 import static pl.allegro.tech.hermes.api.constraints.Names.ALLOWED_NAME_REGEX;
 
-public class Subscription {
+public class Subscription implements Anonymizable {
```
sputnikci: \[Checkstyle\] ERROR: Class Subscription should be declared as final\.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Anonymizable.java
```diff
@@ -0,0 +1,5 @@
+package pl.allegro.tech.hermes.api;
+
+public interface Anonymizable {
+    Anonymizable anonymize();
```
adamdubiel: i don't like this interface, feels like it enters the domain to fix this single problem in this single context and adds no additional value - it returns an `Anonymizable` instance, which probably won't be useful. When i anonymize `Subscription` i want to get back a subscription, not a very generic interface.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/Auditor.java
```diff
@@ -1,13 +1,13 @@
 package pl.allegro.tech.hermes.management.domain;
 
-public interface Auditor {
-    default void objectCreated(String username, Object createdObject) {
+public interface Auditor<T extends Object> {
```
adamdubiel: this is now a generic interface, yet you lose generic information in all places where it is used - maybe we don't need it to be generic after all? I mean Auditor is here to audit, not to mingle with objects too much

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/oauth/OAuthProviderService.java
```diff
@@ -1,26 +1,30 @@
...
 
     private final OAuthProviderRepository repository;
-
     private final ApiPreconditions preconditions;
+    private final Auditor auditor;
 
     @Autowired
-    public OAuthProviderService(OAuthProviderRepository repository, ApiPreconditions preconditions) {
+    public OAuthProviderService(OAuthProviderRepository repository, ApiPreconditions preconditions,
+                                @Qualifier("anonymizingAuditor") Auditor auditor) {
```
adamdubiel: example of losing generics i mentioned in previous comment

adamdubiel:
I don't like how complex this is compared to the problem - maybe the solution is too general, while we have literally 4 places to change (use anonimize)?

On the other hand i like the idea of `AnonimizingAuditor` and how changing one dependency in `SubscriptionService` makes it hide all passwords that i can think of. We can think on how to get the best of two words tomorrow.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
sputnikci: \[Checkstyle\] ERROR: Class Subscription should be declared as final\.

PR #622, opened 2016-11-10
615 bugfix for occasional null pointer when reading assignements

wojtkiewicz:


sslavic:
We're affected by this too, especially in tests where there is single of everything (hermes-consumers, subscription, ...)

Seems CI job failed for unrelated reason - failed to download JDK. Maybe that can be fixed too to be better handled, or just restarting of the job will pass.

PR #658, opened 2016-12-05
655 max message size per topic

wojtkiewicz:
General idea is that configuring KafkaConsumer the same for every subscription is too inefficient memory wise. This PR introduces way to specify max message size per topic which is needed to  correctly configure **maxPartitionFetchBytes** which basically represents how heavy the buffer is per partition for receiving messages from Kafka. Goal was not to exceed 1 mb by default per subscription.

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Topic.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Topic.java
@@ -5,6 +5,8 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 import javax.validation.Valid;
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 import java.util.Objects;
 
@@ -25,6 +27,14 @@ public class Topic {
     @NotNull
     private ContentType contentType;
 
+    @Min(MIN_MESSAGE_SIZE)
+    @Max(MAX_MESSAGE_SIZE)
+    private int maxMessageSize;
+
+    public static final int MIN_MESSAGE_SIZE = 1024;
+    public static final int MAX_MESSAGE_SIZE = 2 * 1024 * 1024;
+    private static final int DEFAULT_MAX_MESSAGE_SIZE = 50 * 1024;
+
     public enum Ack {
         NONE, LEADER, ALL
     }
@@ -41,7 +51,8 @@ public class Topic {
 
     public Topic(TopicName name, String description, RetentionTime retentionTime,
                  boolean migratedFromJsonType, Ack ack, boolean trackingEnabled, ContentType contentType,
-                 boolean jsonToAvroDryRunEnabled, boolean schemaVersionAwareSerializationEnabled) {
+                 boolean jsonToAvroDryRunEnabled, boolean schemaVersionAwareSerializationEnabled,
+                 int maxMessageSize) {
         this.name = name;
         this.description = description;
         this.retentionTime = retentionTime;
@@ -51,6 +62,7 @@ public class Topic {
         this.contentType = contentType;
         this.jsonToAvroDryRunEnabled = jsonToAvroDryRunEnabled;
         this.schemaVersionAwareSerializationEnabled = schemaVersionAwareSerializationEnabled;
+        this.maxMessageSize = maxMessageSize;
     }
 
     @JsonCreator
@@ -63,9 +75,11 @@ public class Topic {
             @JsonProperty("trackingEnabled") boolean trackingEnabled,
             @JsonProperty("migratedFromJsonType") boolean migratedFromJsonType,
             @JsonProperty("schemaVersionAwareSerializationEnabled") boolean schemaVersionAwareSerializationEnabled,
-            @JsonProperty("contentType") ContentType contentType) {
+            @JsonProperty("contentType") ContentType contentType,
+            @JsonProperty("maxMessageSize") Integer maxMessageSize) {
         this(TopicName.fromQualifiedName(qualifiedName), description, retentionTime, migratedFromJsonType, ack,
-                trackingEnabled, contentType, jsonToAvroDryRunEnabled, schemaVersionAwareSerializationEnabled);
+                trackingEnabled, contentType, jsonToAvroDryRunEnabled, schemaVersionAwareSerializationEnabled,
+                maxMessageSize == null ? DEFAULT_MAX_MESSAGE_SIZE : maxMessageSize);
     }
 
     public RetentionTime getRetentionTime() {
@@ -150,5 +164,9 @@ public class Topic {
 
     public boolean isSchemaVersionAwareSerializationEnabled() {
         return schemaVersionAwareSerializationEnabled;
+    }
+
+    public int getMaxMessageSize() {
+        return maxMessageSize;
     }
 }
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
@@ -1,6 +1,7 @@
 package pl.allegro.tech.hermes.common.config;
 
 import com.google.common.io.Files;
+import pl.allegro.tech.hermes.api.Topic;
 import pl.allegro.tech.hermes.common.util.InetAddressHostnameResolver;
 
 import static java.lang.Math.abs;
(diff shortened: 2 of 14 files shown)


hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionFetchingConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    public static SubscriptionFetchingConfiguration create(Map<String, Object> properties) {
+        return new SubscriptionFetchingConfiguration(
+                (Integer) properties.getOrDefault("maxPartitionFetchBytes", DEFAULT_MAX_FETCH_BYTES)
+        );
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
sputnikci: \[Checkstyle\] ERROR: 'if' construct must use '\{\}'s\.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionFetchingConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    @JsonCreator
+    public static SubscriptionFetchingConfiguration create(Map<String, Object> properties) {
+        return new SubscriptionFetchingConfiguration(
+                (Integer) properties.getOrDefault("maxPartitionFetchBytes", DEFAULT_MAX_FETCH_BYTES)
+        );
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
sputnikci: \[PMD\] INFO: Avoid using if statements without curly braces

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaMessageReceiverFactory.java
```diff
@@ -98,7 +101,11 @@ public MessageReceiver createMessageReceiver(Topic topic,
         props.put(SESSION_TIMEOUT_MS_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_SESSION_TIMEOUT_MS_CONFIG));
         props.put(HEARTBEAT_INTERVAL_MS_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_HEARTBEAT_INTERVAL_MS_CONFIG));
         props.put(METADATA_MAX_AGE_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_METADATA_MAX_AGE_CONFIG));
-        props.put(MAX_PARTITION_FETCH_BYTES_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_MAX_PARTITION_FETCH_BYTES_CONFIG));
+
+        SubscriptionFetchingConfiguration fetchingConfiguration = subscription.getFetchingConfiguration();
+        int globalMaxBytes = configs.getIntProperty(Configs.KAFKA_CONSUMER_MAX_PARTITION_FETCH_BYTES_CONFIG);
+        props.put(MAX_PARTITION_FETCH_BYTES_CONFIG, min(fetchingConfiguration.getMaxPartitionFetchBytes(), globalMaxBytes));
```
adamdubiel: why min? isn't the whole point of it to be able to increase the global size? maybe we should have default (1Mb) and max_max (like 10Mb)

wojtkiewicz: This option is intended to be a higher bound, if we also implement a lower bound then we can enforce minimum buffer size. What you're suggesting would work if we store in zk some sort of pointer to global value and not the value itself. Then we can change that dynamically like that.

integration/src/integration/java/pl/allegro/tech/hermes/integration/management/TopicManagementTest.java
```diff
@@ -2,15 +2,20 @@
 
 import org.assertj.core.api.Assertions;
 import org.testng.annotations.Test;
+import pl.allegro.tech.hermes.api.ContentType;
 import pl.allegro.tech.hermes.api.ErrorCode;
+import pl.allegro.tech.hermes.api.Group;
 import pl.allegro.tech.hermes.api.Topic;
+import pl.allegro.tech.hermes.api.TopicName;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- pl\.allegro\.tech\.hermes\.api\.TopicName\.

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingTest.java
```diff
@@ -15,6 +16,7 @@
 import pl.allegro.tech.hermes.integration.metadata.TraceContext;
 import pl.allegro.tech.hermes.integration.shame.Unreliable;
 import pl.allegro.tech.hermes.test.helper.builder.SubscriptionBuilder;
+import pl.allegro.tech.hermes.test.helper.builder.TopicBuilder;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- pl\.allegro\.tech\.hermes\.test\.helper\.builder\.TopicBuilder\.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Topic.java
```diff
@@ -25,6 +27,12 @@
     @NotNull
     private ContentType contentType;
 
+    @Min(1024)
+    @Max(2 * 1024 * 1024)
```
druminski: You implemented limits in two places. The first one is here and the second one is in Configs.java. When you change limits in one place then you have to remember to change them in other one as well. Can we declare them just in one place so that I don't need to remember about changes in other place?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -37,7 +37,7 @@
     KAFKA_CONSUMER_SESSION_TIMEOUT_MS_CONFIG("kafka.consumer.session.timeout.ms", 200_000),
     KAFKA_CONSUMER_HEARTBEAT_INTERVAL_MS_CONFIG("kafka.consumer.heartbeat.interval.ms", 3000),
     KAFKA_CONSUMER_METADATA_MAX_AGE_CONFIG("kafka.consumer.metadata.max.age.ms", 5 * 60 * 1000),
-    KAFKA_CONSUMER_MAX_PARTITION_FETCH_BYTES_CONFIG("kafka.consumer.max.partition.fetch.bytes", 10 * 1024 * 1024),
+    KAFKA_CONSUMER_MAX_PARTITION_FETCH_BYTES_CONFIG("kafka.consumer.max.partition.fetch.range", 1024 + "-" + 5 * 1024 * 1024),
```
adamdubiel: can we not create new config type? i think 2 properties min & max will be better considering our current setup, this bring the element of "magic" into configs

adamdubiel:
You did not include changes in hermes-console.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Topic.java
wojtkiewicz: I can't see how this is configured in Configs. There is an option for `KAFKA_CONSUMER_MAX_PARTITION_FETCH_BYTES_CONFIG` but these values do not map 1-1 to each other for all cases.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
wojtkiewicz: It wasn't obvious to me how to call these two properties:
* `kafka.consumer.max.min.partition.fetch.bytes`
* `kafka.consumer.max.max.partition.fetch.bytes`
Looks just weird. If you have something else in mind that would be very helpful.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/ContentLengthChecker.java
```diff
@@ -29,4 +34,11 @@ private static boolean isChunked(HeaderMap headerMap, long requestContentLength)
                     expected, contentLength));
         }
     }
+
+    public static final class ContentTooLargeException extends IOException {
+        ContentTooLargeException(long contentLength, int max) {
+            super(format("Content-Length is too large [length:%s, max:%s].",
```
wojtkiewicz: done

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
adamdubiel: `kafka.consumer.partition.fetch.bytes.min` and `kafka.consumer.partition.fetch.bytes.max`

adamdubiel: thats why it has `partition` there? :P why does thing with same name configure st completly different?

wojtkiewicz: Tricky part is that size of fetch per partition is configured by min (default=1 byte) and max (default=1 mb). What we're dealing with here is a range allowed for max value.

wojtkiewicz: I added two separate properties that perhaps are a bit easier to understand

hermes-console/static/partials/modal/editTopic.html
```diff
@@ -58,6 +58,15 @@ <h3 class="modal-title" ng-show="operation === 'EDIT'"><small>Edit topic:</small
                     </select>
                 </div>
             </div>
+            <div class="form-group {{topicForm.topicMaxMessageSize.$valid ? '' : 'has-error'}}">
+                <label for="topicMaxMessageSize" class="col-md-3 control-label">Max message size</label>
+                <div class="col-md-9">
+                    <div class="input-group">
+                        <input type="number" min="1024" required class="form-control" id="topicMaxMessageSize" name="topicMaxMessageSize" ng-model="topic.maxMessageSize" placeholder="how large messages are for this topic?"/>
```
adamdubiel: we need to have default value here

PR #663, opened 2016-12-12
fetch values form -2minutes from Graphite to avoid incomplete data

adamdubiel:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/graphite/WebTargetGraphiteClient.java
```diff
@@ -25,8 +25,8 @@
     public WebTargetGraphiteClient(WebTarget webTarget) {
         this.webTarget = webTarget
                 .path("render")
-                .queryParam("from", "-1minutes")
-                .queryParam("until", "now")
+                .queryParam("from", "-5minutes")
```
chemicL: Do we need 3 minutes of data? The logic fetches just the last data point. I just want to confirm whether graphite guarantees to return just the data points which are available for this query.

chemicL: That's what I mean - can we tell if graphite returned 0 because it's zero, or because it's filling for missing data? If we can't then maybe take last non-zero value or output zero if missing.

adamdubiel: i added search for last nonnull value

Issue #643, opened 2016-11-23
Make `consumers-workload.monitor.missing` Gauge instead of Counter


sslavic:
As ever increasing `Counter`, `consumers-workload.monitor.missing` metric is tricky to work with for alarming, especially for corner cases where `missing` changes but alarms likely aren't appropriate.

E.g. when cluster restarts (for example for rolling upgrade) with enough of consumers assigned per instance, but with total below capacity, positive number can be reported now (not all `ConsumerProcess`es started yet at the time metric was sampled and Counter got incremented), and then it remains constant until real to be alarmed "missing" `ConsumerProcess` event occurs. Hard part here with this standard scenario is that if one would alarm on change of missing, it wouldn't be possible to clear based on data that nothing is missing, since missing will never go down back to 0, since it's a `Counter`.

If `Gauge` would be used instead, reading values from `int` or `AtomicInteger`, that `ConsumersRuntimeMonitor` would updates as it runs, it would return to 0 after all `ConsumerProcess` have started and there are no missing ones.

Issue #641, opened 2016-11-23
Introduce consumers-workload.monitor.running metric


sslavic:
`ConsumersRuntimeMonitor` is reporting `consumers-workload.monitor.missing` and `consumers-workload.monitor.oversubscribed` metrics already. Please consider reporting `consumers-workload.monitor.running` metric as well.

Knowing how many there are running `ConsumerProcess`es and capacity per instance, one can alarm/notify sooner and act (increase capacity or whatever appropriate), before problem actually happens (cluster is over capacity which is signaled already by missing-resources metric).

adamdubiel:
We actually use `threads` metrics for this, as it shows how many subscription threads are currently running, but you are right - it would be better to have it explicitly named in `consumers-workload` metrics.

PR #664, opened 2016-12-12
introduces gauges in consumers-workload metrics

### LINKED ISSUE #642: Make missing-resources Gauge instead of Histogram
Alternatively to using reservoir flushed on snapshot (see https://github.com/allegro/hermes/issues/606) please consider making `missing-resources` use `Gauge` instead. It can be taking value from an `int` or `AtomicInteger`, which `BalancingJob` would update. This will enable one to clearly report whether `hermes-consumers` cluster is missing resources or not now, without history. Yes, all but last measurements sampled between two reporter snapshots would be lost, but they are of no value for monitoring/alarming anyway, and if one really needs to investigate if it's oscillating in between reporter calls - logs contain raw information with every `BalancingJob` run.

adamdubiel:
Fixes #642 #641 #642

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
@@ -145,7 +145,10 @@ public class HermesMetrics {
     }
 
     public void registerGauge(String name, Gauge<?> gauge) {
-        metricRegistry.register(pathCompiler.compile(name), gauge);
+        String path = pathCompiler.compile(name);
+        if (!metricRegistry.getGauges().containsKey(name)) {
+            metricRegistry.register(path, gauge);
+        }
     }
 
     private String metricRegistryName(String metricDisplayName, TopicName topicName, String subscription) {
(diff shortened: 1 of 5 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/selective/SelectiveWorkBalancer.java
```diff
@@ -57,7 +58,7 @@ private int countMissingResources(SubscriptionAssignmentView state) {
     }
 
     private void equalizeWorkload(SubscriptionAssignmentView state, SubscriptionAssignmentView.Transformer transformer) {
-        if (state.getSubscriptionsCount() > 1) {
+        if (state.getSubscriptionsCount() > 1 && !state.getConsumerNodes().isEmpty()) {
```
adamdubiel: this should protect us agains different weird effects, because there are methods inside that do not tolerate empty set of consumer nodes

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
```diff
@@ -252,13 +255,6 @@ public Timer consumersWorkloadRebalanceDurationTimer(String kafkaCluster) {
         return metricRegistry.timer(pathCompiler.compile(Timers.CONSUMER_WORKLOAD_REBALANCE_DURATION, pathContext));
     }
 
-    public void reportConsumersWorkloadStats(String kafkaCluster, int missingResources, int deletedAssignmentsCount, int createdAssignmentsCount) {
-        PathContext pathContext = pathContext().withKafkaCluster(kafkaCluster).build();
-        metricRegistry.histogram(pathCompiler.compile(Histograms.CONSUMERS_WORKLOAD_SELECTIVE_MISSING_RESOURCES, pathContext)).update(missingResources);
```
chemicL: Should we delete the constants? If the gauges use the same names then we should make sure they don't get reused.

adamdubiel: Fixed.

adamdubiel:
Also fixes #643

PR #678, opened 2016-12-21
Fixing max sleep time for zookeeper connection retries  

druminski:
This PR adds conversion from seconds to milliseconds which was missing for max sleep time between zookeeper connection retries. Thanks this retries will be done between 30 seconds instead of 30 milliseconds.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
@@ -12,7 +12,8 @@ public enum Configs {
 
     ZOOKEEPER_CONNECT_STRING("zookeeper.connect.string", "localhost:2181"),
     ZOOKEEPER_BASE_SLEEP_TIME("zookeeper.base.sleep.time", 1000),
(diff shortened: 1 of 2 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -12,7 +12,8 @@
 
     ZOOKEEPER_CONNECT_STRING("zookeeper.connect.string", "localhost:2181"),
     ZOOKEEPER_BASE_SLEEP_TIME("zookeeper.base.sleep.time", 1000),
-    ZOOKEEPER_MAX_SLEEP_TIME("zookeeper.max.sleep.time.seconds", 30),
+    ZOOKEEPER_MAX_SLEEP_TIME_IN_SECONDS("zookeeper.max.sleep.time.seconds", 30),
+    ZOOKEEPER_MAX_RETRIES("zookeeper.max.retries", 29),
```
druminski: Decreasing to 29 as this is maximum value which can be used in ExponentialBackoffRetry.

PR #680, opened 2016-12-22
resolve #679 | Check if response is not null in retry condition

### LINKED ISSUE #679: Hermes client hangs on sender error
If ```HermesSender.send``` throw an exception (_not_ returns exceptionally completed future), then NPE is thrown from ```HermesClientBasicRetryCondition``` which somehow causes the request to never complete.

msulima:


hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClientBasicRetryCondition.java
```diff
@@ -7,7 +7,7 @@
 public class HermesClientBasicRetryCondition implements Predicate<HermesResponse> {
     @Override
     public boolean test(HermesResponse response) {
-        return isClientTimeoutOrServerError(response) || isFailedExceptionally(response);
+        return (response != null) && (isClientTimeoutOrServerError(response) || isFailedExceptionally(response));
```
adamdubiel: if i get it right: if response is null, we will not retry. i think we should retry even if response is null. btw what caused this in your case?

so it would be:

```
response == null || ....
```

msulima: Issue was caused by an exception from discovery service (no consul service on host).

There is a separate mechanism that repeats requests in case of exceptions, but anyway I've changed the predicate according to your suggestion.

Issue #694, opened 2017-01-20
Leaking file descriptors


druminski:
From time to time we are observing file descriptors leak in hermes-frontend module.

![screen shot 2017-01-20 at 11 58 54](https://cloud.githubusercontent.com/assets/4473056/22147243/fd5b39a8-df07-11e6-80fe-f0715064a3a7.png)

It is probably related to problems with saving messages in backup storage.

PR #696, opened 2017-01-20
#694 Additional exception handling in hermes-frontend

druminski:
I have noticed that exceptions from backup storage can be swallowed. Therefore this PR:

- handles exceptions at the beginning of tasks which are run directly on worker thread without `Handler` layer
- reports number of messages in backup storage. This value can help us diagnose whether we have a problem with deleting messages from backup storage
- handles exceptions in few other places

Further investigation revealed corner case which was not handled during race between ack and timeout tasks. Because of it number of messages in backup storage can grow with a time. This PR fixes this problem providing additional message state `DELAYED_SENT_TO_KAFKA`

Solves #694

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/Gauges.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/Gauges.java
@@ -18,5 +18,6 @@ public class Gauges {
 
             THREADS = "threads",
             INFLIGHT_REQUESTS = "inflight-requests",
-            OUTPUT_RATE = "output-rate." + GROUP + "." + TOPIC + "." + SUBSCRIPTION;
+            OUTPUT_RATE = "output-rate." + GROUP + "." + TOPIC + "." + SUBSCRIPTION,
+            BACKUP_STORAGE_SIZE = "backup-storage.size";
 }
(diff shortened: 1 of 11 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/Gauges.java
```diff
@@ -18,5 +18,6 @@
 
             THREADS = "threads",
             INFLIGHT_REQUESTS = "inflight-requests",
-            OUTPUT_RATE = "output-rate." + GROUP + "." + TOPIC + "." + SUBSCRIPTION;
+            OUTPUT_RATE = "output-rate." + GROUP + "." + TOPIC + "." + SUBSCRIPTION,
+            BACKUP_STORAGE_MESSAGES = "backup-storage-messages-number";
```
adamdubiel: `backup-storage.size`? this way we can put some more metrics in `backup-storage` if we need in future

druminski: sure, done

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/message/MessageState.java
```diff
@@ -76,7 +77,11 @@ public boolean setDelayedProcessing() {
         return timeoutHasPassed && state.compareAndSet(SENDING_TO_KAFKA, DELAYED_PROCESSING);
     }
 
+    public boolean setDelayedSentToKafka() {
+        return state.compareAndSet(DELAYED_SENDING, DELAYED_SENT_TO_KAFKA) || state.compareAndSet(DELAYED_PROCESSING, DELAYED_SENT_TO_KAFKA);
```
sputnikci: \[Checkstyle\] INFO: Line is longer than 140 characters \(found 141\)\.

PR #704, opened 2017-02-03
Make the docker-compose work with the latest versions

miguelpuyol:
- Use Kafka advertised.listeners
- Change Bootstrap servers to kafka:9092 (from the default localhost:9092)


miguelpuyol:
I closed it as I did a new one with only the intended changes for Allegro

PR #705, opened 2017-02-03
Make the docker-compose work with the latest versions

miguelpuyol:
- Use Kafka advertised.listeners
- Change Bootstrap servers to kafka:9092 (from the default localhost:9092)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,48 +1,52 @@
-zookeeper:
-  image: wurstmeister/zookeeper
-  ports:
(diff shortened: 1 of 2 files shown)


adamdubiel:
@miguelpuyol i would like to merge #709 instead of this. #709 is heavily based on your work, but i also tweaked the settings a bit to the point when everything seems to work flawlessly (i had some problems with your version). I hope this is okay with you :)

miguelpuyol:
Sounds good. I will close it

PR #709, opened 2017-02-06
Fix docker-compose and docker setup

adamdubiel:
@miguelpuyol i been trying to test your PR and found some more places to improve, not to mess up with your PR i decided to create my own that was based on your - i hope this is okay with you?

I also switched to official Confluent Kafka docker images. They are bit bigger than `wurstmeister` ones, but things seem to work stable on them.


miguelpuyol:
Sounds great! I will try to switch to the official confluent images too.

PR #713, opened 2017-02-07
Admin scripts

chemicL:
Adding admin scripts catalogue with first script for migrating consumers to 0.10.5

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,36 @@
+Admin scripts that might come in handy.
+
+# Python scripts
+
+The python scripts are simple to run using virtualenv and pip.
+
+Run the commands to prepare the environment:
+
+```bash
+# setup and activate virtualenv
+$ virtualenv .
+$ . bin/activate
+
+# install the dependencies
+$ pip install -r requirements.txt
+
+# if you want to deactivate virtualenv
+$ deactivate
+```
+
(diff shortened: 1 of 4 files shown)


scripts/migrate_consumers_0.10.5.py
```diff
@@ -0,0 +1,34 @@
+import sys
+import argparse
+import logging
+from kazoo.client import KazooClient
+
+logging.basicConfig()
+
+parser = argparse.ArgumentParser(description="Update assignments to auto assigned.")
```
adamdubiel: use click instead of parsers

scripts/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Python scripts
+
+The python scripts are simple to run using virtualenv and pip.
+
+Run the commands to prepare the environment:
+
+```bash
+# setup and activate virtualenv
+$ virtualenv venv
```
adamdubiel: why in venv? :P

scripts/requirements.txt
```diff
@@ -0,0 +1,5 @@
+appdirs==1.4.0
```
adamdubiel: we need all those? can't see them being used

chemicL: dependencies

Issue #717, opened 2017-02-08
NPE in Hermes frontend related to BlacklistZookeeperNotifyingCache


gamefundas:
Noticed a few of these erros appear in frontend logs. Not seeing message loss at the moment but any pointers what could cause this?

java.lang.NullPointerException: null
    at pl.allegro.tech.hermes.frontend.blacklist.BlacklistZookeeperNotifyingCache.childEvent(BlacklistZookeeperNotifyingCache.java:27)
    at org.apache.curator.framework.recipes.cache.PathChildrenCache$5.apply(PathChildrenCache.java:522)
    at org.apache.curator.framework.recipes.cache.PathChildrenCache$5.apply(PathChildrenCache.java:516)
    at org.apache.curator.framework.listen.ListenerContainer$1.run(ListenerContainer.java:93)
    at com.google.common.util.concurrent.MoreExecutors$DirectExecutorService.execute(MoreExecutors.java:299)
    at org.apache.curator.framework.listen.ListenerContainer.forEach(ListenerContainer.java:84)
    at org.apache.curator.framework.recipes.cache.PathChildrenCache.callListeners(PathChildrenCache.java:513)
    at org.apache.curator.framework.recipes.cache.EventOperation.invoke(EventOperation.java:35)
    at org.apache.curator.framework.recipes.cache.PathChildrenCache$9.run(PathChildrenCache.java:773)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)

adamdubiel:
I noticed the same messages in our logs yesterday. From the looks of it we received some faulty event from Curator watcher - an object that is null. This is not connected to how we process incoming messages, so no loss should be expected :)

chemicL:
This is just zookeeper reconnect and connect events, which don't carry data. A null check is required - I'll submit a PR later during the day. It has no negative side effects, apart from causing an ugly log, so no worries :)

chemicL:
[...]

PR #718, opened 2017-02-08
Null check for cache event in BlacklistZookeeperNotifyingCache

chemicL:
Fixing https://github.com/allegro/hermes/issues/717


adamdubiel:
`develop`, not master!

chemicL:
Changed base to develop. Github should allow configuring this as the default ;-)

adamdubiel:
Commit message is too vague, you changed specific cache not "null check for cache event"

chemicL:
Updated the commit message

PR #723, opened 2017-02-15
Correcting issues caused by subscriptions with rate set to too low value

chemicL:
In case of negotiated max rate algorithm, if subscription rate is lower than consumer count, the algorithm could calculate a negative max rate value.

This in turn would collapse the whole rate limiting supervisor, as the iteration was stopped by a misconfigured rate limit.


adamdubiel:
@pbetkier lets squeeze it into 0.11, it's pretty serious

Issue #728, opened 2017-02-17
User friendly message when sending garbage as Avro


adamdubiel:
Currently when user sends JSON on Avro topic without `Content-Type: application/json`, Frontend returns `400 Bad Content`, which is ok, but also internally throws:

```
pl.allegro.tech.hermes.frontend.validator.InvalidMessageException: Could not deserialize avro message with provided schema Errors: ArrayIndexOutOfBoundsException: 
	at pl.allegro.tech.hermes.frontend.validator.AvroTopicMessageValidator.check(AvroTopicMessageValidator.java:26)
	at pl.allegro.tech.hermes.frontend.validator.MessageValidators.lambda$check$0(MessageValidators.java:19)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at pl.allegro.tech.hermes.frontend.validator.MessageValidators.check(MessageValidators.java:19)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.createAvroMessage(MessageFactory.java:108)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.create(MessageFactory.java:86)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.create(MessageFactory.java:60)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageCreateHandler.handleRequest(MessageCreateHandler.java:35)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.PreviewHandler.handleRequest(PreviewHandler.java:22)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.messageRead(MessageReadHandler.java:138)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.lambda$partialMessageRead$2(MessageReadHandler.java:105)
	at io.undertow.io.AsyncReceiverImpl.receivePartialBytes(AsyncReceiverImpl.java:582)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.readMessage(MessageReadHandler.java:80)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.handleRequest(MessageReadHandler.java:54)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.TopicHandler.lambda$handleRequest$0(TopicHandler.java:47)
[...]

PR #738, opened 2017-03-09
728 payload content-type check & error handling

rprzystasz:
Added check which accepts only "binary/avro" or "application/json" message content-type.
In other cases,  http status 400 and proper error message will be returned.

#728

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/UnsupportedContentTypeException.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/UnsupportedContentTypeException.java
@@ -1,5 +1,6 @@
 package pl.allegro.tech.hermes.common.message.wrapper;
 
+import org.apache.avro.Schema;
 import pl.allegro.tech.hermes.api.Subscription;
 import pl.allegro.tech.hermes.api.Topic;
 import pl.allegro.tech.hermes.common.exception.InternalProcessingException;
@@ -19,6 +20,14 @@ public class UnsupportedContentTypeException extends InternalProcessingException
                 "Unsupported content type %s for subscription %s",
                 subscription.getContentType(),
                 subscription.getQualifiedName()
+        ));
+    }
+
+    public UnsupportedContentTypeException(String payloadContentType, Schema schema) {
+        super(String.format(
+                "Unsupported payload content type %s for %s",
+                payloadContentType,
+                schema.getFullName()
         ));
     }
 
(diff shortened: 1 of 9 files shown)


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/message/MessageContentTypeEnforcer.java
```diff
@@ -1,25 +1,39 @@
...
-                contentType.startsWith(APPLICATION_JSON_WITH_DELIM) : contentType.equals(APPLICATION_JSON));
+        return isOfType(contentType, APPLICATION_JSON, APPLICATION_JSON_WITH_DELIM);
+    }
+
+    private boolean isAvro(String contentType) {
+        return isOfType(contentType, AVRO_BINARY, AVRO_BINARY_WITH_DELIM);
+    }
+
+    private boolean isOfType(String contentType, String expectedContentType, String expectedWithDelim) {
+        return contentType != null && (contentType.length() > expectedContentType.length() ?
```
chemicL: shouldn't we just check `startsWith`?

chemicL: also, content type is not case sensitive (https://www.w3.org/Protocols/rfc1341/4_Content-Type.html) so we can match with `String.toLowerCase`

rprzystasz: checking with startsWith doesn't enforce ';' to be the first character after content-type value.
It would accept values like 'application/jsonANYTEXT'
Do we want to allow such values?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/UnsupportedContentTypeException.java
```diff
@@ -22,4 +23,12 @@ public UnsupportedContentTypeException(Subscription subscription) {
         ));
     }
 
+    public UnsupportedContentTypeException(String payloadContentType, Schema schema) {
```
adamdubiel: why Avro Schema here? i don't get it, we probably want topic name, not schema name

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/message/MessageContentTypeEnforcer.java
adamdubiel: no, the current algorithm is fine

chemicL: Good point.
We can still simplify the check
`contentType != null && (contentType.equals(expectedContentType)) || contentType.startsWith(expectedWithDelim))`
`String.equals` has a check for length.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/UnsupportedContentTypeException.java
rprzystasz: That's right. In most cases schema name matches topic name, and tests didn't catch that typo.
Fixed

Issue #448, opened 2016-04-16
Underscore in topic name leads to invalid nodes creation in zk.


druminski:
When topic is created via Hermes API and it contains underscore in the name, for example `topic_v1` then node with the same name is created in zk. However when messages are published on that topic then counters are reported in zk under `topic.v1/metrics/published` node. The problem is that when metrics are reported to zk, then underscore is converted to dot and new empty topic node is created. This leads to invalid structures creation in zk and prevents me from listing all subscriptions via /query/subscriptions endpoint because `Unable to read data from path` is thrown.

adamdubiel:
Fixed by #741

PR #741, opened 2017-03-14
Topics with underscore have invalid metrics paths in zookeeper

### LINKED ISSUE #740: Invalid metrics names in zookeeper for topics with underscore in name
When topic has underscore char in name then its counters are kept in zookeeper under wrong path (underscores are escaped to dots). Hence, for this topic:

- counter values are not read properly so user sees 0 for published, delivered, discarded counters
- counted number of consumers assigned to subscription is always 1. Because of this calculated max-rate is too high

druminski:
Resolves #740

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporter.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporter.java
@@ -93,7 +93,7 @@ public class ZookeeperCounterReporter extends ScheduledReporter {
         TopicName topicName = fromQualifiedName(qualifiedTopicName);
         return new TopicName(
                 escapeMetricsReplacementChar(topicName.getGroupName()),
-                escapeMetricsReplacementChar(topicName.getName())
+                topicName.getName()
         );
     }
 
--- a/hermes-common/src/test/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporterTest.java
+++ b/hermes-common/src/test/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporterTest.java
@@ -37,23 +37,23 @@ public class ZookeeperCounterReporterTest {
     public static final SortedMap<String, Gauge> EMPTY_GAUGES = null;
     public static final String GROUP_NAME_UNDERSCORE = "pl_allegro_tech_skylab";
     public static final String GROUP_NAME = "pl.allegro.tech.skylab";
-    public static final String TOPIC_NAME = "topic1";
+    public static final String TOPIC_NAME_UNDERSCORE = "topic_1";
     public static final String SUBSCRIPTION_NAME_UNDERSCORE = "subscription_name";
     public static final String SUBSCRIPTION_NAME = "subscription.name";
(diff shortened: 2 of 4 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporter.java
```diff
@@ -93,7 +93,7 @@ private static TopicName escapedTopicName(String qualifiedTopicName) {
         TopicName topicName = fromQualifiedName(qualifiedTopicName);
         return new TopicName(
                 escapeMetricsReplacementChar(topicName.getGroupName()),
-                escapeMetricsReplacementChar(topicName.getName())
+                topicName.getName()
```
adamdubiel: will it be backwards compatible? what with all the metrics we already collected?

chemicL:
If I remember correctly, subscription name could also be a problem.
Writing:
    escapeMetricsReplacementChar(matcher.getSubscriptionName()),
Reading:
    subscription.getName()

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporter.java
druminski: Yes, this change is backward compatible. It only touches broken metrics which actually were not read by Hermes (symptoms described here #740). Topics without underscore in names have valid metrics names and nothing will change for them.

adamdubiel: okay, one more: why do we fix topic name and not group name? what if group name has an underscore?

druminski:
@chemicL no, it is not a problem for subscription because hermes-console does not allow to use underscore char in subscription name. I also checked our production state and confirm that we don't have subscriptions with underscores.

As for
```
escapeMetricsReplacementChar(matcher.getSubscriptionName())
```
we have to do this because ZookeeperCounterReporter reads metrics from MetricRegistry and subs there are escaped from `.` to `_`. So to keep original sub name we have to escape it again before send it to zookeeper.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/counter/zookeeper/ZookeeperCounterReporter.java
druminski: Currently we don't have underscores in groups names on prod and in this PR I am adding restriction to proper field in hermes-console, so this situation will not occur in the future. 

When metric is created in MetricRegistry then Hermes escapes group, topic and subscription names from `.` to `_`. However topic name can't have dots so this transformation will do nothing for topic name. This is the reason why we can use underscore in topic name but not in group and subscription name. In ZookeeperCounterReporter Hermes again transforms `_` to `.` to keep original names but it shouldn't do this for topic name.

chemicL:
Great, I also checked if we allow underscores in subscription names: we do not. Let's go :-)

adamdubiel:
Will it fix #448  as well?

druminski:
@adamdubiel yes, after this fix structures for metrics in zk will be created correctly. However, to keep neatness in zookeeper we have to remove those broken one.

adamdubiel:
Ok, but i can close #448 as related problem.

druminski:
sure

Issue #710, opened 2017-02-06
Limit size of messages in preview


adamdubiel:
Some topics might publish messages too big to fit into the preview and into ZK document in general, which causes `IOException` at ZK driver level. We should set limit for preview size and either trim messages or discard them (and leave a *Message too big to preview* note).

mszarlinski:
I'm working on it.

PR #752, opened 2017-04-06
Limit size of messages in preview #710

mszarlinski:
Message preview will be truncated in `MessagePreviewLog#add` if its size exceeds limit defined in `Configs.FRONTEND_MESSAGE_PREVIEW_MAX_SIZE_KB` (10 by default). Additionally to preview content, a `truncated` flag will be stored in Zookeeper for later display in "Topic message preview" panel:

image

Previews already stored in ZK (flat structure, only preview content) are not supported, meaning that they won't be displayed. Once new messages arrive, they will be stored in new format and older ones should be discarded.


adamdubiel:
@mszarlinski please change branch to develop

PR #754, opened 2017-04-11
Add filtering by headers #753

### LINKED ISSUE #753: Add filtering by headers
Hermes forwards some of http headers (configurable) to subscribers. We should allow on defining filters based on header values, just like we do in message body.

mszarlinski:
This PR introduces a new type of message filter and registers it in `FilterChain`. 

closes #753


adamdubiel:
Please add an integration test.

PR #764, opened 2017-05-15
Custom button on ui

druminski:
Configuring custom view near topic buttons area on UI.

I will use this space to add migration button for JSON topics. The button will forward a user to an external migration tool.

This PR fixes also NPE threw by schema save endpoint on JSON topic.


adamdubiel:
can you add some examples to sample `config.json`?

druminski:
@adamdubiel just added an example to the `config.json.example`

Issue #773, opened 2017-06-12
max-rate metric already exists


druminski:
Few days ago one of our consumers machine encountered network issues. They lasted about 20 minutes and because of them consumers instance lost contact with kafka, zookeeper, graphite, logs server. After this unreachability most of consumers processes were able to recover. Unfortunately three of them were not. They were restarted indefinitely because of this exceptions:

```
java.lang.IllegalArgumentException: A metric named consumers-rate.max-rate.node.{groupName}.{topicName}.{subscriptionName}.max-rate already exists
        at com.codahale.metrics.MetricRegistry.register(MetricRegistry.java:91) ~[metrics-core-3.1.1.jar:3.1.1]
        at pl.allegro.tech.hermes.common.metric.HermesMetrics.registerMaxRateGauge(HermesMetrics.java:299) ~[hermes-common-0.11.3.jar:na]
        at pl.allegro.tech.hermes.consumers.consumer.rate.maxrate.NegotiatedMaxRateProvider.start(NegotiatedMaxRateProvider.java:92) ~[hermes-consumers-0.11.3.jar:na]
        at pl.allegro.tech.hermes.consumers.consumer.rate.calculator.OutputRateCalculator.start(OutputRateCalculator.java:62) ~[hermes-consumers-0.11.3.jar:na]
        at pl.allegro.tech.hermes.consumers.consumer.rate.SerialConsumerRateLimiter.initialize(SerialConsumerRateLimiter.java:48) ~[hermes-consumers-0.11.3.jar:na]
        at pl.allegro.tech.hermes.consumers.consumer.SerialConsumer.initialize(SerialConsumer.java:130) ~[hermes-consumers-0.11.3.jar:na]
        at pl.allegro.tech.hermes.consumers.supervisor.process.ConsumerProcess.start(ConsumerProcess.java:127) ~[hermes-consumers-0.11.3.jar:na]
        at pl.allegro.tech.hermes.consumers.supervisor.process.ConsumerProcess.run(ConsumerProcess.java:57) ~[hermes-consumers-0.11.3.jar:na]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_45]
        at java.util.concurrent.FutureTask.run(FutureTask.java:266) [na:1.8.0_45]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_45]
[...]

PR #774, opened 2017-06-13
Ensuring that consumer process exists for processed signal

druminski:
Signals for `ConsumerProcessSupervisor` are produced from few threads on a single queue. These threads are not synchronized with each other which means that signals can be mixed on the queue.
Next, signals are consumed in batches by a single thread. Single batch can contain mixed sequence of signals like so:

```
1. CLEAN - from some STOP signal
2. COMMIT - from commit thread
``` 

At the moment, `ConsumerProcessSupervisor` verifies whether consumer process exists only at the beginning of a batch process. However, some signals from the batch can remove consumer process and next signals from this batch can end up with NPE as consumer process was not found.

This PR fixes described problem by verifying whether consumer process exists for each processed signal.

This is a first step to solve #773

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
@@ -16,6 +16,10 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Future;
 
+import static pl.allegro.tech.hermes.consumers.supervisor.process.Signal.SignalType.KILL_UNHEALTHY;
+import static pl.allegro.tech.hermes.consumers.supervisor.process.Signal.SignalType.RESTART;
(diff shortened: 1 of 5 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
```diff
@@ -85,61 +89,70 @@ private void restartUnhealthy() {
...
     }
 
-    private ConsumerProcess process(Signal signal) {
-        return runningProcesses.getProcess(signal.getTarget());
+    private void onConsumerProcess(Signal signal, java.util.function.Consumer<ConsumerProcess> consumerProcessConsumer) {
+        if (runningProcesses.hasProcess(signal.getTarget())) {
+            consumerProcessConsumer.accept(runningProcesses.getProcess(signal.getTarget()));
+        } else {
+            metrics.counter("supervisor.signal.dropped." + signal.getType().name()).inc();
+            logger.warn("Dropping signal {} as dedicated consumer process does not exist.", signal);
```
adamdubiel: `..as target consumer process...`

Issue #779, opened 2017-06-19
Backwards compatible schemaVersionAwareSerializationEnabled flag


adamdubiel:
Currently it is impossible to switch `schemaVersionAwareSerializationEnabled` flag on topic on-the-fly. This can only be turned on for new topics.

This flag enables the "linkedin serialization" of Avro schema version in Hermes, i.e. prepending magic byte and schema version to Avro payload.

We need to enable this flag on-the-fly (part of messages on the topic will have magic-byte prepended and part will not).

PR #781, opened 2017-06-22
Schema version aware deserialization is backward compatible

druminski:
Adding fallback from schema version aware deserialization to any schema deserialization. Solves #779

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/DeserializationMetrics.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/DeserializationMetrics.java
@@ -1,0 +1,41 @@
+package pl.allegro.tech.hermes.common.message.wrapper;
+
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.MetricRegistry;
+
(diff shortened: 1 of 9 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/DeserializationMetrics.java
```diff
@@ -0,0 +1,41 @@
...
+    public Counter missedSchemaVersionInPayload() {
+        return metricRegistry.counter(name(deserializationPath(), "missed", "schemaVersionInPayload"));
+    }
+
+    private String deserializationErrorsPath() {
+        return deserializationPath() + ".errors";
+    }
+
+    private String deserializationPath() {
+        return "avro.deserialization";
```
adamdubiel: maybe sth like `content.avro.deserialization`? so in the future we might add `content.json.*` metrics, it would be weird to have `.json` as root metric

druminski: sure, changed

PR #785, opened 2017-06-27
Improved topic and schema management

### LINKED ISSUE #692: Better UX with schema support in hermes-console
One of our clients gave us a feedback about schema support in hermes-console, here are the notes:

- when schema is invalid then displayed error message is not easy to read. Notification is too small and displayed too short

- when topic create button is clicked and schema is invalid then error message is displayed but topic is still created

- when remove topic button is clicked but schema is not available then error message is returned

- when topic is created after removal, schema registry may have retained the schema (e.g. production policy), we should inform the client properly

dankraw:
* `/topics` now accepts topic enhanced with schema field for in-transaction management
* simplified exceptions model for schema repository clients allowing client/server errors separation (so we get 400 bad request instead of 500 when schema is not accepted by the external repository)

fixes #692


hermes-api/src/main/java/pl/allegro/tech/hermes/api/TopicWithSchema.java
```diff
@@ -0,0 +1,49 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.util.Objects;
+
+public class TopicWithSchema {
```
adamdubiel: this could be an object with `extends Topic`, Jackson is wise enough

hermes-api/src/main/java/pl/allegro/tech/hermes/api/endpoints/TopicWithSchemaEndpoint.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Response;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Path("topics-with-schema")
```
adamdubiel: i would rather have `TopicEndpoint` working with `TopicWithSchema` object in management, adding new endpoint feels weird

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicWithSchemaService.java
```diff
@@ -0,0 +1,99 @@
...
+    public TopicWithSchema getTopicWithSchema(TopicName topicName) {
+        Topic topic = topicService.getTopicDetails(topicName);
+        Optional<RawSchema> schema = Optional.empty();
+        if (isAvroTopic(topic)) {
+            schema = schemaService.getSchema(topicName.qualifiedName());
+        }
+        return new TopicWithSchema(topic, schema.map(RawSchema::value).orElse(null));
+    }
+
+    public void createTopicWithSchema(TopicWithSchema topicWithSchema, String createdBy, CreatorRights isAllowedToManage) {
```
adamdubiel: i see no transaction here - if topic creation fails, schema will be created and subsequent tries to create topic will fail..

since it is not always possible to delete schema, i would rather:

* create topic
* create schema
* if schema creation fails - delete topic

adamdubiel: we can minimize the amount of rollbacks with calling Schema Repo validation endpoint before attempting to create topic/schema, same as we call topic validator

hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/TopicsWithSchemaEndpoint.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.SecurityContext;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.status;
+import static pl.allegro.tech.hermes.api.TopicName.fromQualifiedName;
+
+@Component
+@Path("/topics-with-schema")
```
adamdubiel: change to `topic`

hermes-test-helper/src/main/java/pl/allegro/tech/hermes/test/helper/builder/TopicBuilder.java
```diff
@@ -61,6 +64,14 @@ public Topic build() {
         );
     }
 
+    public TopicWithSchema buildAsTopicWithSchema() {
```
adamdubiel: can we use proper methods from `TopicWithSchema` where necessary instead of extending this builder? it feels somehow out of place to have build take arguments and return different type. Optionally, add `String schema` field to builder class and add `withSchema` method, then builder semantics is kept and you can get rid of `buildAsTopicWithSchema(String schema)` and only keep `buildTopicWithSchema()`:

```
topic().withSchema(...).buildTopicWithSchema()
```

dankraw: done + updated to current develop

PR #788, opened 2017-07-07
 Consumer SSL config changes to match the 0.12.0

ajayk:
* Update Consumer SSL config changes to match the 0.12.0


adamdubiel:
Thanks for the update and sorry for the inconvenience :)

PR #789, opened 2017-07-07
resolves #777 | username to allow .~- in accordance with URI spec

### LINKED ISSUE #777: The regex for username in Push subscription is not in alignment with URI spec.
The current regex for username (`[a-zA-Z0-9_\.]*`) constraints it to have only letters, numerals, "_" and ".". 

However, the [URI spec](https://tools.ietf.org/html/rfc3986) is less restrictive for the user information. Please refer to [user information section](https://tools.ietf.org/html/rfc3986#section-3.2.1) and [unreserved characters section](https://tools.ietf.org/html/rfc3986#section-2.3).

Making it less restrictive and in alignment with URI spec helps to integrate the Push Subscription with tools that auto-generate usernames as per the URI spec.

The code refers to the section: `pl.allegro.tech.hermes.api.EndpointAddress#URL_PATTERN`

ajayk:


adamdubiel:
Thanks for fixing this!

PR #793, opened 2017-07-11
Minor documentation fix

ajayk:
* Add required fields in the create subscription call 
* Correct the Json Payload in Subscription docs


adamdubiel:
Nice, thanks a lot :)

Issue #758, opened 2017-04-13
Use seconds for ElasticSearch message tracking


chemicL:
For fetching last undelivered messages from ElasticSearch, we sort them by timestamp.
Unfortunately for ElasticSearch, milliseconds is not the best choice for sorting - check  https://mixmax.com/blog/30x-faster-elasticsearch-queries

We might be better off by adding an additional field just for sorting purposes, which contains the timestamp in seconds.

chemicL:
Resolved by #795

PR #795, opened 2017-07-13
Adds additional timestamp_seconds field for sorting in Elastic Search tracker

adamdubiel:
Solves #758


adamdubiel:
@chemicL will it fix your issue?

chemicL:
@adamdubiel sure does, thanks!

Issue #780, opened 2017-06-20
ZookeeperCounterReporter can block topic removal


chemicL:
`ZookeeperCounterReporter` stores metrics in `/groups/{group}/topics/{topic}/metrics` and is accessed by different processes than management (consumers, frontend). Because of that, if a topic is removed, for which there are still published messages, the topic removal will fail as the zookeeper path is persisted.

This occurs as console errors while displaying topic details (empty json) and prevents removal.

PR #800, opened 2017-07-28
#780 Fixed problem with topic removal due to metrics

jakubdyszkiewicz:
There is validation on topic removal that there are no subscriptions on
that topic. But in the same time the metric which has subpath same as
subscription path is stored (e.g. subscription path is
/groups/{group}/topics/{topic}/subscriptions/{sub} and metric path is
/groups/{group}/topics/{topic}/subscriptions/{sub}/metrics/delivered
Validation checks only if path exists.

The situation when there were no subscription but metric was stored
may be caused by missing subscription check while storing inflight
counter or because of the race condition (removing topic and storing
metric at the same time).

The solution is to check if data for subscription path is actually
present and also to add missing check.


adamdubiel:
Fixes #780

PR #818, opened 2017-09-11
Query subscriptions metrics

faderskd:
Added endpoint for querying subscriptions metrics.


hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionNameWithMetrics.java
```diff
@@ -0,0 +1,111 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+public class SubscriptionNameWithMetrics {
+    private final String topicQualifiedName;
+    private final String name;
```
druminski: below fields also can be final

Issue #802, opened 2017-08-01
Special chars restriction in group, topic and subscription name


druminski:
In our production environment we have few topics which contain `-` char in their names. Unfortunately, Avro specification allows only to use `[A-Za-z0-9_]` and because of this we can not migrate these topics easily. Actually we have to create new ones.

To avoid this problem for future topics Hermes should not allow to use `-`.

Furthermore, Hermes should limit special chars usage in group, topic & subscription name to minimum, because as you can see special chars can be problematic. Also regexp validation should detect typos in which too much dots were provided, for example: `group..topic`

PR #823, opened 2017-09-27
Special chars restriction in group, topic and subscription name

piorkowskiprzemyslaw:
Implements #802

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/constraints/Names.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/constraints/Names.java
@@ -6,11 +6,13 @@ public final class Names {
 
     public static final String ALLOWED_NAME_REGEX = ALLOWED_CHARACTERS + "+";
 
-    public static final String ALLOWED_GROUP_NAME_REGEX =
+    public static final String ALLOWED_NAME_WITH_DELIMITER_REGEX =
             ALLOWED_NAME_REGEX + "(" + DELIMITER + ALLOWED_NAME_REGEX + ")*";
+
+    public static final String ALLOWED_GROUP_NAME_REGEX = ALLOWED_NAME_WITH_DELIMITER_REGEX;
 
     public static final String ALLOWED_TOPIC_NAME_REGEX = ALLOWED_NAME_REGEX;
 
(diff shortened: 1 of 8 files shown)


adamdubiel:
There are two test failing: `MetricsTest.shouldNotCreateNewTopicWhenAskingForNonExistingMetrics`, would you mind fixing them?

Also i changed the base from `master` to `develop` (as per contributors guide).

piorkowskiprzemyslaw:
Sure, I’ll fix it today.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/constraints/Names.java
```diff
@@ -1,5 +1,16 @@
...
+    private static final String DELIMITER = "\\.";
+
+    public static final String ALLOWED_NAME_REGEX = ALLOWED_CHARACTERS + "+";
+
+    public static final String ALLOWED_GROUP_NAME_REGEX =
+            ALLOWED_NAME_REGEX + "(" + DELIMITER + ALLOWED_NAME_REGEX + ")*";
+
+    public static final String ALLOWED_TOPIC_NAME_REGEX = ALLOWED_NAME_REGEX;
+
+    public static final String ALLOWED_SUBSCRIPTION_NAME_REGEX = ALLOWED_NAME_REGEX;
```
druminski: Subscription name also can contain implicit namespace separated by dots. Although, we do not force it to use it by Hermes users but in Allegro we are using it widely as it is more verbose name. For example we have hundreds of subscriptions named in convention `pl.allegro.{domain}.{context}`, for instance `pl.allegro.advertising.inventory`. Therefore, please do not restrict dots usage in subscription name here and and in Hermes-console.

piorkowskiprzemyslaw: Hi, thanks for review. I've addressed your comment in latest commit (regexp for subscription name validation).

adamdubiel:
I'm sorry but i have to revert this commit - it breaks too much in existing subscriptions and topic. Actually you are not able to edit them anymore. We need this change, but it has to be done in a soft manner for now (like popup on Console saying that using `-` will break Avro compatibility).

PR #829, opened 2017-10-20
resolve #821 | decoding avro message to json for preview

### LINKED ISSUE #821: Topic message preview in human readable form
Topic message preview in Hermes-console displays original message send on a topic. In JSON topics this feature works great, however in Avro topics displaying binary message on UI is not a good idea as the message is not readable. Example:

screen shot 2017-09-21 at 15 05 37

It would be nice to convert this Avro message to JSON before displaying it in Hermes-console.

mictyd:
Decodes avro message for preview (writes to zookeeper decoded value). If it fails, raw message is previewed. For json, behaviour is not changed.

### DIFF
--- a/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/PreviewHandler.java
+++ b/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/PreviewHandler.java
@@ -21,7 +21,7 @@ public class PreviewHandler implements HttpHandler {
         try {
             next.handleRequest(exchange);
         } finally {
-            messagePreviewLog.add(attachment.getTopic().getName(), attachment.getMessageContent());
+            messagePreviewLog.add(attachment.getTopic().getName(), attachment.getMessage());
         }
     }
 }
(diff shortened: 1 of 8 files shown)


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/PreviewHandler.java
```diff
@@ -21,7 +24,8 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
         try {
             next.handleRequest(exchange);
         } finally {
-            messagePreviewLog.add(attachment.getTopic().getName(), attachment.getMessageContent());
+            byte[] previewBytes = converter.convert(attachment.getMessage());
+            messagePreviewLog.add(attachment.getTopic().getName(), previewBytes);
```
wojtkiewicz: this code runs for every message?

adamdubiel: i think it might be better to add this somewhere in log, so it runs for like 1 msg/minute instead for every msg

mictyd: I'll move it to the MessagePreviewFactory class, it seems it is a right place for it. Now it trims bytes to desired length and puts in MessagePreview object, it could convert whole Message object to Message preview instead of working on raw bytes.

adamdubiel:
@mictyd would you mind rebasing with develop so that builds start passing? we fixed couple issues to make builds stable and upgraded the ALPN versions

PR #831, opened 2017-10-20
Splits unit, integration and JMH tests into separate, parallel builds on Travis

adamdubiel:
Removed caching because it basically doesn't work.


adamdubiel:
Proof that it works:
screen shot 2017-10-20 at 20 09 45

PR #850, opened 2018-02-01
Support for multi-cluster Zookeeper storage

pprocner:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDcRepositoryCommandExecutor.java
```diff
@@ -0,0 +1,76 @@
...
+                                                                       String dcName) {
+        if (toWrap instanceof HermesException) {
+            return (HermesException)toWrap;
+        }
+        return new InternalProcessingException("Execution of command '" + command + "' failed on DC '" +
+                dcName + "'.", toWrap);
+    }
+
+    private <T> void rollback(List<DcBoundRepositoryHolder<T>> repositories, RepositoryCommand<T> command) {
+        for (DcBoundRepositoryHolder<T> repository : repositories) {
```
adamdubiel: i wonder if we should keep on trying to rollback, even if one DC fails. for 2 DCs it is not a problem, but for 3,4.. less manual work to cleanup after failed process. No need to take actions on this now.

druminski:
Closing as work on this feature were continued here #1033

PR #857, opened 2018-05-18
added alpn version for newer java builds

TheCK:
The Java builds 171 and 172 also use the same ALPN version as builds 161 and 162


TheCK:
Test passes locally, and this change shouldn't break anything

adamdubiel:
There were some false positives, but all in all build passed. Thanks!

PR #859, opened 2018-05-23
Fixing ConcurrentModificationException on KafkaConsumer

faderskd:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
```diff
@@ -129,7 +130,8 @@ private void processSignal(Signal signal) {
                 break;
             case KILL_UNHEALTHY:
                 onConsumerProcess(signal, consumerProcess -> {
-                    taskQueue.offer(signal.createChild(START, clock.millis(), consumerProcess.getConsumer()));
+                    Consumer consumerCopy = consumerProcess.getConsumer().createCopy();
```
faderskd: Here the idea is to create new immutable object of Consumer so we avoid situation when there are two alive ConsumerProcesses using the same Consumer object

adamdubiel:
@faderskd can you describe reasoning behind this change? from what i see you moved some responsibilities to `ConsumerProcessSupervisor`, but i don't know why. And after a year you won't know either :)

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcess.java
```diff
@@ -199,23 +204,23 @@ public boolean equals(Object o) {
...
+        return Objects.hash(getSubscriptionName());
     }
 
-    public SubscriptionName getSubscriptionName() {
-        return subscriptionName;
-    }
-
-    public Consumer getConsumer() {
-        return consumer;
+    public Subscription getSubscription() {
```
jakubdyszkiewicz: Is `process.getSubscription().getQualifiedName()` the only one usage of this getter? If so, we could just expose `getSubscriptionName()` created below

faderskd: Unfortunatelly not :) there is one more usage which requires the Subscription object itself.

PR #864, opened 2018-05-29
Consumers maxrate eager start

chemicL:
Lazy loading of MaxRateRegistry's Zookeeper cache can take a very long time. Eagerly initializing the tree should significantly speed up start time of Hermes Consumers.
More improvements can be made, but this one is the low hanging fruit we ought to consider.


chemicL:
I received some questions regarding this solution, so I'd like to explain:

Eagerly loading the tree is faster because it doesn't create watches in Zookeeper - it statelessly reads what's currently available, which is faster.

PR #871, opened 2018-06-08
Improved processes signals management

faderskd:
The purpose of this pull request is to improve processes management to be more predictable and easy to understand. Changes mainly concern ```ConsumerProcessSupervisor``` which is responsible for applying signals to group of ```ConsumerProcess```. 

Changes:
1. Concept of two process groups. First of them contains consumer processes which are currently running and second processes already dying. This way we can monitor both groups and treat them differently.
2. Reduced number of signals: `KILL`, `RESTART`, `RESTART_UNHEALTHY`, `STOP_RESTART`, `KILL_UNHEALTHY`, `CLEANUP` are no longer necessary.


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/NonblockingConsumersSupervisor.java
```diff
@@ -69,8 +68,10 @@ public NonblockingConsumersSupervisor(ConfigFactory configFactory,
                 configFactory.getIntProperty(Configs.CONSUMER_COMMIT_OFFSET_PERIOD),
                 metrics
         );
-        monitor.register(SUBSCRIPTIONS, backgroundProcess::listRunningSubscriptions);
-        monitor.register(SUBSCRIPTIONS_COUNT, backgroundProcess::countRunningSubscriptions);
+        monitor.register(SUBSCRIPTIONS, backgroundProcess::runningSubscriptionsStatus);
+        monitor.register(SUBSCRIPTIONS_COUNT, backgroundProcess::countRunningProcesses);
+
```
faderskd: Do we want to measure dying processes ? These maybe for instance processes for subscriptions that not died after some constnt time despite stop/kill signal

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcess.java
```diff
@@ -63,7 +63,7 @@ public void run() {
         try {
             Thread.currentThread().setName("consumer-" + getSubscriptionName());
 
-            while (running && !Thread.interrupted()) {
+            while (running && !Thread.currentThread().isInterrupted()) {
```
faderskd: if we use Thread.interrupted() this fact will be useless because flag indicating interruption is cleared every time the method is called

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
```diff
@@ -76,27 +77,55 @@ public void run() {
...
                     Signal restartUnhealthy =
-                            Signal.of(RESTART_UNHEALTHY, consumerProcess.getSubscription().getQualifiedName());
+                            Signal.of(RESTART_UNHEALTHY, subscriptionName(consumerProcess));
                     logger.info("Lost contact with consumer {}, last seen {}ms ago {}. {}",
                             consumerProcess, consumerProcess.lastSeen(), restartUnhealthy.getLogWithIdAndType());
                     taskQueue.offer(restartUnhealthy);
                 });
     }
 
+    private void superviseDyingProcesses() {
```
faderskd: This method checks if there are processes which where scheduled with KILL/STOP but are still dying and did not died ultimately (e.g. because consumers objects inside processes are waiting for connection to crashed Kafka and never return). The method continuously sends FROCE_KILL_DYING signals for these stale processes.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
```diff
@@ -118,75 +147,93 @@ private void processSignal(Signal signal) {
...
+        if (consumerProcesses.hasRunningProcess(signal)) {
+            consumerProcessConsumer.accept(consumerProcesses.getRunningProcess(signal));
         } else {
             metrics.counter("supervisor.signal.dropped." + signal.getType().name()).inc();
-            logger.warn("Dropping signal {} as target consumer process does not exist.", signal);
+            logger.warn("Dropping signal {} as running target consumer process does not exist.", signal);
+        }
+    }
+
+    private void forDyingConsumerProcess(Signal signal, java.util.function.Consumer<ConsumerProcess> consumerProcessConsumer) {
```
faderskd: we want to clearly differentiate signals for running and dying processes

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisor.java
```diff
@@ -196,20 +243,24 @@ private void start(Signal start) {
...
 
-                runningProcesses.add(process, future);
+                consumerProcesses.add(process, future);
                 logger.info("Started consumer process for subscription {}. {}", start.getTarget(), start.getLogWithIdAndType());
             } catch (Exception ex) {
                 logger.error("Failed to create consumer for subscription {}", subscription.getQualifiedName(), ex);
             }
+        } else if (consumerProcesses.hasDyingProcess(start)) {
+            logger.info("Consumer process for {} is already dying, startup deferred.", subscription.getQualifiedName());
+	        accept(start);
```
faderskd: Process startup has to be done but can't be done when process for the same subscription is already dying. The idea is to put this signal on queue until it can be properly processed.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ManagedConsumerProcesses.java
```diff
@@ -0,0 +1,160 @@
...
+	ConsumerProcess getDyingProcess(Signal signal) {
+		return dyingProcesses.get(subscriptionName(signal)).getProcess();
+	}
+
+	void add(ConsumerProcess process, Future executionHandle) {
+		this.runningProcesses.put(
+				process.getSubscription().getQualifiedName(), new RunningProcess(process, executionHandle));
+	}
+
+	void moveToDyingProcesses(Signal signal, long shouldBeKilledAfter) {
```
faderskd: We don't need any synchronization due to fact, that all modifications are done using signals on single ConsumerProcessSupervisor thread

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisorSignalsTest.groovy
```diff
@@ -0,0 +1,160 @@
...
+        consumerProcessSupervisor.run()
+
+        when:
+        consumerProcessSupervisor.accept(
+                Signal.of(signalType, subscription.getQualifiedName(), subscription))
+        consumerProcessSupervisor.run()
+
+        then:
+        conditions.within(4 * delay) {
+            assert System.currentTimeMillis() - timestampContainer.timestamp > 2 * delay * 1000
```
faderskd: This is quite tricky, we just measure diff time between time in condition thread and thread in consumer process. If consumer process is stopped/killed it does not update its timestamp and diff exceeds some threshold.

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisorSignalsTest.groovy
```diff
@@ -0,0 +1,160 @@
...
+        given:
+        def delay = 0.5
+        def timestampContainer = new TimestampContainer()
+
+        def consumer = Stub(Consumer)
+        def conditions = new PollingConditions(timeout: 4 * delay, delay: delay)
+
+        consumer.consume(_) >> { Runnable signalsInterrupt ->
+            while (true) {
+                Thread.sleep(200)
```
faderskd: Unfortunatelly we need this sleep() for consumer to be stoppable. Without this java does not kill the process (but FutureTask says that cancel() method was stopped correctly). This can be good way to simulate the real production problem and find some better solution to kill stale processes.

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcessSupervisorSignalsTest.groovy
jakubdyszkiewicz: Ok, but why `2 *` and what's delay (and why it is `0.5`). I assume `* 1000` is for converting to seconds.

faderskd: Ok, it maybe a bit confusing. Firstly "delay" is passed to PollingConditions and it basically means the rate (in seconds) at which condition will be checked. So here we've got every 0.5 seconds. 
```java
conditions.within(4 * delay)
```
This line of code means that we can spare at most 2 seconds for waiting for signal to be processed (I just think it is reasonable max time - maybe shorter would be ok too, signals are on the queue and there is supervisor run interval set on 100ms so it tries to process signals at most 20 times)
```java
assert System.currentTimeMillis() - timestampContainer.timestamp > 2 * delay * 1000
```
here 2 * delay * 1000 means that the condition loop ran at least 2 times  and the timestampContainer.timestamp was not updated

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -106,6 +107,11 @@ public KafkaSingleThreadedMessageReceiver(KafkaConsumer<byte[], byte[]> consumer
                 }
             }
             return Optional.ofNullable(readQueue.poll());
+        } catch (InterruptException ex) {
+            logger.info("Kafka consumer thread interrupted", ex);
+            logger.info("Is current thread interrupted ? ---> {} ", Thread.currentThread().isInterrupted());
```
faderskd: Wow, despite the Thread.currentThread().interrupt() is called in ```InterruptException```'s constructor this log prints false.

faderskd: But when you do this in local test it always logs true... very weird but it is the solution to problem of hung processes

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -154,7 +154,11 @@ private UnwrappedMessageContent getUnwrappedMessageContent(ConsumerRecord<byte[]
 
     @Override
     public void stop() {
-        consumer.close();
+        try {
```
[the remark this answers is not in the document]
wojtkiewicz: good question, I need to look at the code again

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/NonblockingConsumersSupervisor.java
wojtkiewicz: I think that you added such a gauge to `ConsumerProcessSupervisor`.

PR #879, opened 2018-06-26
Interrupting ConsumerProcess thread when catched InterruptException f…

faderskd:
This PR improves dealing with Kafka ```InterruptException``` thrown by ```ConsumerNetworkClient```. Due to the fact that ```InterruptException```'s constructor has ```Thread.currentThread().interrupt()``` call and in some unknown (for us) way does not change Thread to the interrupted state, we have to do it explicitly by making this call once again.
Changes:
1. Interrupting ```ConsumerProcess```'s  thread in places where ```InterruptException``` may be raised from Kafka client (```poll``` and ```commitSync``` methods of ```KafkaConsumer```class) .
2. Added ```Thread.currentThread.isInterrupted()``` to run method of ```ConsumerProcess``` (instead of ```Thread.interrupted()```) to preserve interrupt state once it is set.

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
@@ -7,6 +7,7 @@ import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.InterruptException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import pl.allegro.tech.hermes.api.ContentType;
@@ -106,6 +107,12 @@ public class KafkaSingleThreadedMessageReceiver implements MessageReceiver {
                 }
             }
             return Optional.ofNullable(readQueue.poll());
+        } catch (InterruptException ex ) {
+            // Despite that Thread.currentThread().interrupt() is called in InterruptException's constructor
+            // Thread.currentThread().isInterrupted() somehow returns false so we reset it.
+            logger.info("Kafka consumer thread interrupted", ex);
+            Thread.currentThread().interrupt();
+            return Optional.empty();
         } catch (KafkaException ex) {
             logger.error("Error while reading message for subscription {}", subscription.getQualifiedName(), ex);
             return Optional.empty();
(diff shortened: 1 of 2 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -106,7 +107,13 @@ public KafkaSingleThreadedMessageReceiver(KafkaConsumer<byte[], byte[]> consumer
                 }
             }
             return Optional.ofNullable(readQueue.poll());
-        } catch (KafkaException ex) {
+        }
+        catch (InterruptException ex ) {
+            logger.debug("Interrupting kafka consumer...", ex);
+            Thread.currentThread().interrupt();
```
chemicL: I think this is one of those rare moments where a comment like
`// KafkaConsumer clears the interrupted status, so we enable it again`

On the other hand, why not rethrow the `ex.getCause()` and handle it in the caller?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -106,7 +107,13 @@ public KafkaSingleThreadedMessageReceiver(KafkaConsumer<byte[], byte[]> consumer
...
+            return Optional.empty();
+        }
+        catch (KafkaException ex) {
```
chemicL: `} catch (...) {` (same line with braces, as in other parts of the code)

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
faderskd: I just found that constructor of this exception already interrupt the exception, I will study this topic more :) 
'''public class InterruptException extends KafkaException {

    private static final long serialVersionUID = 1L;
    
    public InterruptException(InterruptedException cause) {
        super(cause);
        Thread.currentThread().interrupt();
    }
    
    public InterruptException(String message, InterruptedException cause) {
        super(message, cause);
        Thread.currentThread().interrupt();
    }

    public InterruptException(String message) {
        super(message, new InterruptedException());
        Thread.currentThread().interrupt();
    }

}'''

chemicL: Great find. I'm sincerely surprised by this implementation. Nevertheless it feels like at current stage this invalidates the PR, unless we handle our own exception instead of relying on interrupted state.

PR #886, opened 2018-07-04
List topics by owners

jakubdyszkiewicz:
We want to be able to list topics by owners. We can do this right now by using `QueryEndpoint`, but quering requires scanning through whole Zookeeper, which is too slow (something about 1 to 15 seconds on production).

Another reason that seperate endpoint is better is that we can change response structure. We may also add list subscriptions for topics.

I built cache from `OwnerId` to `TopicName` which is filled after application startup. Then every CRUD operation on topic updates the cache.

I created seperate class that is used by `TopicService`. I considered implementing existing `Audit` interface and inject `List<Audit>` instead of `Audit` so that `TopicService` would not know about cache, but I feel like `Audit` and `Cache` is a different thing. Moreover, Audit methods are called by multiple Services, not only `TopicService` so I would have to check class with `instanceof`.

### DIFF
--- a/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicOwnerCache.java
+++ b/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicOwnerCache.java
@@ -1,0 +1,60 @@
+package pl.allegro.tech.hermes.management.domain.topic;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+import pl.allegro.tech.hermes.api.OwnerId;
+import pl.allegro.tech.hermes.api.Topic;
+import pl.allegro.tech.hermes.api.TopicName;
+import pl.allegro.tech.hermes.domain.topic.TopicRepository;
+import pl.allegro.tech.hermes.management.domain.group.GroupService;
+
+import javax.annotation.PostConstruct;
+import java.util.Collection;
+
+@Component
+public class TopicOwnerCache {
+
+    private static final Logger logger = LoggerFactory.getLogger(TopicOwnerCache.class);
+
+    private final TopicRepository topicRepository;
+    private final GroupService groupService;
+
+    private final Multimap<OwnerId, TopicName> cache = Multimaps.synchronizedMultimap(ArrayListMultimap.create());
+
+    public TopicOwnerCache(TopicRepository topicRepository, GroupService groupService) {
+        this.topicRepository = topicRepository;
+        this.groupService = groupService;
+    }
+
+    public Collection<TopicName> get(OwnerId ownerId) {
+        return cache.get(ownerId);
+    }
+
+    @PostConstruct
+    public void fillCache() {
+        logger.info("Starting filling Owner Id to Topic Name cache");
+        long start = System.currentTimeMillis();
+        groupService.listGroupNames().stream()
+                .flatMap(groupName -> topicRepository.listTopics(groupName).stream())
+                .forEach(topic -> cache.put(topic.getOwner(), topic.getName()));
+        long end = System.currentTimeMillis();
+        logger.info("Cache filled. Took {}ms", end - start);
+    }
+
+    public void onRemovedTopic(Topic topic) {
+        cache.remove(topic.getOwner(), topic.getName());
+    }
+
+    public void onCreatedTopic(Topic topic) {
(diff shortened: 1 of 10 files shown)


hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicOwnerCache.java
```diff
@@ -0,0 +1,60 @@
...
+import pl.allegro.tech.hermes.api.Topic;
+import pl.allegro.tech.hermes.api.TopicName;
+import pl.allegro.tech.hermes.domain.topic.TopicRepository;
+import pl.allegro.tech.hermes.management.domain.group.GroupService;
+
+import javax.annotation.PostConstruct;
+import java.util.Collection;
+
+@Component
+public class TopicOwnerCache {
```
adamdubiel: but we have more than one `hermes-management` instance, how will you sync the cache between nodes?

cristaloleg: do we really need a synchronised cache for that?

adamdubiel: no, we can do it in background once per X. however in current implementation cache is reloaded only on `hermes-management` start. so if i change topic owner, i need to restart management instances? or am i missing something :)

cristaloleg: `TopicOwnerCache` is used in `TopicService`(especially line https://github.com/allegro/hermes/pull/886/files/db8bf84063b6f6a9b5c6174bcea6eb09aa031cb2#diff-c14949a99fc0d34090e3953832d9f208R133). So cache is always updated, but it's not limited btw.

adamdubiel: I know, but it is only updated on single instance :) Assume:

* there are two instances of `hermes-management` (`A` and `B`)
* instance `A` and `B` fill cache on startup
* owner of topic `T` is changed, request was sent to `A`
* `A` saves changes in Zookeeper and updates its in-memory cache
* `B` doesn't know about the change
* you get stale data from `B` until it gets restarted :(

jakubdyszkiewicz: I can't belive I fell for that one... I will try to use `ZookeeperInternalNotificationBus` from `hermes-commons` that is currently used in `hermes-frontend` and `hermes-consumers`

adamdubiel: I think simplest solution would be better - notification bus needs to create another set of watches. How about simple refresh every few minutes? It's not critical.

jakubdyszkiewicz: Ok, let's try with refreshing for now. We can move to more consistent cache (by using `ZookeeperInternalNotificationBus`) later if that will be a problem

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicOwnerCache.java
```diff
@@ -22,20 +27,31 @@
...
+        scheduledExecutorService.shutdown();
     }
 
     public Collection<TopicName> get(OwnerId ownerId) {
         return cache.get(ownerId);
     }
 
-    @PostConstruct
-    public void fillCache() {
+    private void fillCache() {
```
faderskd: może dajmy tutaj try-catch prewencyjnie, skoro to zaciąga coś z zookeepera to może wylecieć jakiś wyjątek i executor się zatrzyma

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicOwnerCache.java
```diff
@@ -0,0 +1,78 @@
...
+    @PreDestroy
+    public void stop() {
+        scheduledExecutorService.shutdown();
+    }
+
+    public Collection<TopicName> get(OwnerId ownerId) {
+        return cache.get(ownerId);
+    }
+
+    private void refillCache() {
```
faderskd: może dajmy tutaj try-catch prewencyjnie, skoro to zaciąga coś z zookeepera to może wylecieć jakiś wyjątek i executor się zatrzymać

jakubdyszkiewicz: Good catch, added.

PR #891, opened 2018-07-31
Setting Hermes ID as empty when unwrapping avro message without given…

faderskd:
This PR stops generating missing Hermes ID for messages from Kafka. When message id is missing or whole metadata is empty Hermes does not generate message id at all. It is just leaved as empty.
Changes:
* ```unwrapContent()``` method in ```AvroMessageContentWrapper``` now sets missing message id as empty


hermes-common/src/test/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapperTest.java
```diff
@@ -71,14 +72,42 @@ public void shouldUnwrapAvroMessageAndGenerateMetadataWhenNotExists() throws Thr
         UnwrappedMessageContent unwrappedMessage = avroMessageContentWrapper.unwrapContent(wrappedMessage, avroUser.getCompiledSchema());
 
         //then
-        assertThat(unwrappedMessage.getMessageMetadata().getId()).isNotEmpty();
+        assertThat(unwrappedMessage.getMessageMetadata().getId()).isEmpty();
```
chemicL: While this test is in package common, I'm not sure whether this test makes sense. It would make sense if it was in consumers package - then we'd be sure it's the case of non-generated ID's. But if this code can be used in both consumers and frontend, I'd suggest clearly defining where the wrapping/unwrapping can happen and moving to consumers code if necessary.

faderskd: Ok ```AvroMessageContentWrapper.unwrapContent()``` is package private so we need to use higher level classes to test it but I understand your intentions.

faderskd: I think writing this kind of test in consumers package (with appropriate name will be best option). What do you think ?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/util/MessageId.java
```diff
@@ -1,26 +1,10 @@
...
-                .toString();
-
-    }
-
-    public static String forTimestamp(long timestamp) {
-        return HASH_FUNCTION.newHasher()
-                .putLong(timestamp)
-                .hash()
-                .toString();
+    public static String empty() {
```
chemicL: Seems in this case, `MessageId` class becomes useless, IMO we can just remove it.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
```diff
@@ -45,11 +45,12 @@ UnwrappedMessageContent unwrapContent(byte[] data, CompiledSchema<Schema> schema
     private MessageMetadata getMetadata(Map<Utf8, Utf8> metadata) {
         if (metadata == null) {
             long timestamp = clock.millis();
-            return new MessageMetadata(timestamp, MessageId.forTimestamp(timestamp), Collections.EMPTY_MAP);
+            return new MessageMetadata(timestamp, MessageId.empty(), Collections.EMPTY_MAP);
```
chemicL: Maybe skip the `MessageId` param in that constructor and have `MessageMetadata` handle the case of missing ID - we get single responsibility for free.

hermes-common/src/test/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapperTest.java
chemicL: I had a look at the usage of `AvroMessageContentWrapper` and noticed it's used in consumers and management code. At this point perhaps it would be enough to add a javadoc describing the responsibility of the class - that it does not generate any ID, only deals with wrapping and unwrapping what is provided?

PR #894, opened 2018-08-09
Sending delay

jakubdyszkiewicz:
This PR implements feature of delayed events. We want to give users possibility to postpone sending an event for given time (max 5 seconds) so if there are multiple topics that sends messages at the same time, then can increase chance of receiving an event from one topic before an event from another topic.

I implemented this feature in serial subscriptions. It will require more code and testing to support batch subscriptions. I believe that clients that asked about this feature use serial subscriptions. We can always support batches in a separate PR.

### DIFF
--- a/hermes-console/static/partials/modal/editSubscription.html
+++ b/hermes-console/static/partials/modal/editSubscription.html
@@ -134,7 +134,7 @@
                 <div class="col-md-9">
                     <div class="input-group">
                         <input type="number" min="0" max="7200" step="1" required class="form-control" id="subscriptionMessageTtl" name="subscriptionMessageTtl" placeholder="Time when message can be resent to endpoint after failed sending attempts" ng-model="subscription.subscriptionPolicy.messageTtl"/>
-                        <span class="input-group-addon">seconds</span>
+                        <span class="input-group-addon">milliseconds</span>
                         <span class="input-group-addon helpme-addon" uib-tooltip="Amount of time a message can be held in sending queue and retried. If message will not be delivered during this time, it will be discarded.">?</span>
                     </div>
                 </div>
--- a/hermes-console/static/partials/modal/editSubscription.html
+++ b/hermes-console/static/partials/modal/editSubscription.html
@@ -119,7 +119,16 @@
                     </div>
                 </div>
             </div>
-
+            <div class="form-group {{subscriptionForm.subscriptionSendingDelay.$valid ? '' : 'has-error'}}" ng-show="subscription.deliveryType === 'SERIAL'">
+                <label for="subscriptionSendingDelay" class="col-md-3 control-label">Sending delay</label>
+                <div class="col-md-9">
+                    <div class="input-group">
+                        <input type="number" min="0" max="5000" step="1" required class="form-control" id="subscriptionSendingDelay" name="subscriptionSendingDelay" placeholder="Delay after which an event will be send" ng-model="subscription.subscriptionPolicy.sendingDelay"/>
+                        <span class="input-group-addon">seconds</span>
+                        <span class="input-group-addon helpme-addon" uib-tooltip="Amount of time in ms after which an event will be send. Useful if events from two topics are sent at the same time and you want to increase chance that events from one topic will be deliver after events from other topic.">?</span>
+                    </div>
+                </div>
+            </div>
             <div class="form-group {{subscriptionForm.subscriptionMessageTtl.$valid ? '' : 'has-error'}}">
                 <label for="subscriptionMessageTtl" class="col-md-3 control-label">Inflight message TTL</label>
                 <div class="col-md-9">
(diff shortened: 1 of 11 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -89,18 +89,26 @@ public void shutdown() {
...
+            return 0;
+        }
+    }
+
     /**
      * Method is calling MessageSender and is registering listeners to handle response.
      * Main responsibility of this method is that no message will be fully processed or rejected without release on semaphore.
      */
-    public void sendMessage(final Message message) {
+    private void sendMessage(final Message message) {
```
jakubdyszkiewicz: I made this private, because it was used only in above `sendAsync` methods and `ConsumerMessageSenderTest`. This test can use `sendAsync` as well, so I changed it there.

integration/src/integration/java/pl/allegro/tech/hermes/integration/management/SubscriptionManagementTest.java
```diff
@@ -127,6 +130,43 @@ public void shouldUpdateSubscriptionEndpoint() throws Throwable {
         remoteService.waitUntilReceived();
     }
 
+    @Test
+    public void shouldUpdateSubscriptionPolicy() {
```
jakubdyszkiewicz: I had problem with updating topic with `sendingDelay` so I created test for it. I turned out that I forgot to change hashCode and equals. I decided that I could just leave this test, it might be useful in future.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
```diff
@@ -89,18 +89,26 @@ public void shutdown() {
...
-        sendAsync(message, 0);
+        sendAsync(message, delayForSubscription());
     }
 
     private void sendAsync(Message message, int delayMillis) {
         retrySingleThreadExecutor.schedule(() -> sendMessage(message), delayMillis, TimeUnit.MILLISECONDS);
     }
 
+    private int delayForSubscription() {
+        if (!subscription.isBatchSubscription()) {
```
adamdubiel: AFAIR `ConsumerMessageSender` is used only for serial subscriptions, so this if is redundant

adamdubiel:
This implementation adds absolute delay for every message on a subscription. So even if messages are old (for example: retransmission was applied) they are going to get delayed. Maybe we can inspect timestamp added by `hermes-frontend` and only delay message if it was published at `T < now() + delay()`?

hermes-console/static/partials/modal/editSubscription.html
```diff
@@ -119,7 +119,16 @@ <h3 class="modal-title" ng-show="operation === 'EDIT'"><small>Edit:</small> {{su
                     </div>
                 </div>
             </div>
-
+            <div class="form-group {{subscriptionForm.subscriptionSendingDelay.$valid ? '' : 'has-error'}}" ng-show="subscription.deliveryType === 'SERIAL'">
+                <label for="subscriptionSendingDelay" class="col-md-3 control-label">Sending delay</label>
+                <div class="col-md-9">
+                    <div class="input-group">
+                        <input type="number" min="0" max="5000" step="1" required class="form-control" id="subscriptionSendingDelay" name="subscriptionSendingDelay" placeholder="Delay after which an event will be send" ng-model="subscription.subscriptionPolicy.sendingDelay"/>
+                        <span class="input-group-addon">seconds</span>
```
cristaloleg: s/seconds/milliseconds ?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSender.java
cristaloleg: Yep, it's used only in `SerialConsumer`.

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/ConsumerMessageSenderTest.java
```diff
@@ -332,10 +338,34 @@ public void shouldDeliverToNewSenderAfterModifiedTimeout() {
...
-        verify(otherMessageSender).send(message);
+        verify(otherMessageSender, timeout(1000)).send(message);
+    }
+
+    @Test
+    public void shouldDelaySendingMessageForOneSecond() {
+        // given
+        Subscription subscription = subscriptionBuilderWithTestValues()
+                .withSubscriptionPolicy(subscriptionPolicy().applyDefaults()
+                        .withSendingDelay(500)
```
cristaloleg: Tests says `shouldDelaySendingMessageForOneSecond` but here we've `500`. Am I missing something?

jakubdyszkiewicz: Good catch. Started from one second, changed it later to speed up test. It is fine, but forgot to change method name.

PR #909, opened 2018-09-19
Schema update improvements

### LINKED ISSUE #908: Notifications about schema update
When user updates schema through hermes-console then this schema will be visible in hermes-frontends after cache refresh which is 10 minutes by default. Because of this one of our client received errors as he published events on topic with new schema version header too quickly. It would be nice to see updated schema in hermes instances as soon as possible.

Furthermore, right now when we delete latest version of schema in schema-registry and add new one then Hermes will not see any changes as it has cached versions which don't change for this kind of operation. In this case we have to restart Hermes.

druminski:
Fixes #908


hermes-common/src/main/java/pl/allegro/tech/hermes/common/schema/AvroCompiledSchemaRepositoryFactory.java
```diff
@@ -40,6 +50,20 @@ public AvroCompiledSchemaRepositoryFactory(RawSchemaClient rawSchemaClient, Conf
             public void onTopicRemoved(Topic topic) {
                 repository.removeFromCache(topic);
             }
+
+            @Override
+            public void onTopicCreated(Topic topic) {
+                logger.info("Loading latest schema for created topic {}", topic.getQualifiedName());
+                schemaVersionsRepository.onlineLatestSchemaVersion(topic).ifPresent(
```
dankraw: Can we check whether the topic is avro-type? we'll get rid of unnecessary schema registry requests for json-topics. 
Same on update below.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicService.java
```diff
@@ -55,6 +59,10 @@
     private final Clock clock;
     private final Auditor auditor;
     private final TopicOwnerCache topicOwnerCache;
+    ScheduledExecutorService scheduledTopicExecutor = Executors.newSingleThreadScheduledExecutor(
```
dankraw: private final

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicService.java
```diff
@@ -176,7 +183,10 @@ public void updateTopicWithSchema(TopicName topicName, PatchData patch, String m
         Topic topic = getTopicDetails(topicName);
         boolean validateAvroSchema = AVRO.equals(topic.getContentType());
         extractSchema(patch)
-                .ifPresent(schema -> schemaService.registerSchema(topic, schema, validateAvroSchema));
+                .ifPresent(schema -> {
+                    schemaService.registerSchema(topic, schema, validateAvroSchema);
+                    scheduleTouchTopic(topicName);
```
dankraw: Why not touch always (even without schema being present), so this feature will be more consistent - "the topic is updated, the notification is sent"?

druminski: In the below line we have `updateTopic()` which already sends notification when there are topic changes. However the problem was with only schema update. Previously Hermes didn't send any notification when only schema was updated so other instances were not aware of schema update. This is why I've added here topic touch.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/schema/AvroCompiledSchemaRepositoryFactory.java
druminski: Good idea, done

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicService.java
druminski: done

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicService.java
dankraw: 👍

Issue #913, opened 2018-09-28
Flooding logs about non matching schema while unwrapping messages


dankraw:
For each avro message that consumers are not able to deserialize using specific version of schema we log an error message. Maybe we should somehow manage to limit those messages.

```
Failed to match schema for message for topic ***, schema version ***, fallback to previous.
```

```
pl.allegro.tech.hermes.common.message.wrapper.UnwrappingException: Could not read avro message
	at pl.allegro.tech.hermes.common.message.wrapper.AvroMessageContentWrapper.unwrapContent(AvroMessageContentWrapper.java:43)
	at pl.allegro.tech.hermes.common.message.wrapper.MessageContentWrapper.tryDeserializingUsingAnySchemaVersion(MessageContentWrapper.java:98)
	at pl.allegro.tech.hermes.common.message.wrapper.MessageContentWrapper.tryDeserializingUsingAnySchemaVersion(MessageContentWrapper.java:82)
	at pl.allegro.tech.hermes.common.message.wrapper.MessageContentWrapper.unwrapAvro(MessageContentWrapper.java:53)
	at pl.allegro.tech.hermes.consumers.consumer.receiver.kafka.KafkaSingleThreadedMessageReceiver.getUnwrappedMessageContent(KafkaSingleThreadedMessageReceiver.java:148)
	at pl.allegro.tech.hermes.consumers.consumer.receiver.kafka.KafkaSingleThreadedMessageReceiver.convertToMessage(KafkaSingleThreadedMessageReceiver.java:130)
	at pl.allegro.tech.hermes.consumers.consumer.receiver.kafka.KafkaSingleThreadedMessageReceiver.next(KafkaSingleThreadedMessageReceiver.java:99)
	at pl.allegro.tech.hermes.consumers.consumer.receiver.ThrottlingMessageReceiver.next(ThrottlingMessageReceiver.java:34)
	at pl.allegro.tech.hermes.consumers.consumer.receiver.kafka.FilteringMessageReceiver.next(FilteringMessageReceiver.java:37)
	at pl.allegro.tech.hermes.consumers.consumer.SerialConsumer.consume(SerialConsumer.java:94)
	at pl.allegro.tech.hermes.consumers.supervisor.process.ConsumerProcess.run(ConsumerProcess.java:63)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #926, opened 2018-10-16
Aggregating logger for logback

dankraw:
See the `AggregatingTurboFilter` docs for more info.


dankraw:
Related to #913

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/logback/AggregatingTurboFilter.java
```diff
@@ -0,0 +1,192 @@
...
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            LoggingEventKey that = (LoggingEventKey) o;
+            return level == that.level &&
+                    Objects.equals(message, that.message) &&
+                    Arrays.equals(params, that.params) &&
+                    Objects.equals(marker, that.marker);
```
cristaloleg: Maybe this one should be before `Arrays.equals`. wdyt?

dankraw: It's IDE-generated, but yes, this would make sense.

dankraw: OK it simply depends on the order of fields, I'll fix it.

dankraw: fixed :)

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/logback/AggregatingTurboFilter.java
```diff
@@ -110,6 +119,15 @@ private boolean isAggregatedLog(Marker marker) {
         return marker != null && (marker.equals(MARKER) || marker.contains(MARKER));
     }
 
+    private Optional<Throwable> extractLastParamThrowable(Object[] params) {
+        return Optional.ofNullable(params)
+                .map(Arrays::stream)
+                .flatMap(a -> a.skip(params.length - 1)
```
cristaloleg: Is interested `Throwable` always on a last place?

dankraw: When calling `logger.error("some message {}", "something", exception);` - yes, it will be included as a standard param, this is why I'm fixing it (I'm taking it out, and storing it as the last exception thrown, and removing it from the passed params, so it won't be considered as a part of the "aggregating key").
When calling `logger.error("some message", exception);` - no, it won't be, it will be passed as the `throwable` argument.

So it depends on the exact method API, and slf4j has multiple variants of each.

dankraw: btw similar logic exists in logback, see `ch.qos.logback.classic.spi.LoggingEvent` class, some snippet when creating `LoggingEvent` instance:

```java
    this.argumentArray = argArray;

    if(throwable == null) {
      throwable = extractThrowableAnRearrangeArguments(argArray); // <---- here
    }

    if (throwable != null) {
      this.throwableProxy = new ThrowableProxy(throwable);
      LoggerContext lc = logger.getLoggerContext();
      if (lc.isPackagingDataEnabled()) {
        this.throwableProxy.calculatePackagingData();
      }
    }
```

...but it all takes place later (after TurboFilters are called), so I have to take care of it by myself and remove the exception.

dankraw: Oh, and the difference is I pass the exception as the `throwable` argument, not as a standard param. But in the end it doesn't matter.

cristaloleg: ok, got it :)

PR #936, opened 2018-10-19
resolve #860 | added message filter type validation

### LINKED ISSUE #860: Validate subscription filter type
We shouldn't allow to create a `jsonpath` subscription filter on an Avro topic via management REST API. Right now we allow adding it but it then has no effect.

klacia:
Message filter type validation added.


adamdubiel:
Looks good, but one thing needs to change :) The type of allowed filters are specified by **topic** content type, not subscription. Can you change it?

PR #937, opened 2018-10-19
resolve #844| 503 response without Retry-After does not ignore rate limiting

### LINKED ISSUE #844: "503 Service Unavailable" response without Retry-After header causes rate calculation to be ignored
If the subscriber responds with "503 Service Unavailable" without `Retry-After` header Hermes ignores output rate calculation and retries according to `subscriptionPolicy.messageBackoff` (100 ms by default).
A question was asked about this behaviour #842.
IMO we should consider this as a bug and change this behaviour to similar to standard 5xx.

spooz:
503 response without Retry-After does not ignore rate limiting (increments failures). I also did some refactoring to make the algorithm more readable.


spooz:
thx! :)

PR #939, opened 2018-10-19
Added option for creating suspended subscriptions

### LINKED ISSUE #194: Create a suspended subscription

mtrojanowski:
Resolves #194


integration/src/integration/java/pl/allegro/tech/hermes/integration/management/SubscriptionManagementTest.java
```diff
@@ -320,6 +320,26 @@ public void shouldReturnNoDataStatusWhenGraphiteRespondsWithAnError() {
         assertThat(subscriptionHealth).isEqualTo(SubscriptionHealth.NO_DATA);
     }
 
+    @Test
+    public void shouldCreateSubscriptionWithSuspendedStatus() {
```
mtrojanowski: I was wondering wether this test is really necessary. I wanted to be sure that `ACTIVE` is not set as default and overwrites the value that comes from the request.

hermes-console/static/partials/modal/editSubscription.html
```diff
@@ -248,6 +248,16 @@ <h3 class="modal-title" ng-show="operation === 'EDIT'"><small>Edit:</small> {{su
                 </div>
             </div>
 
+            <div class="form-group" ng-show="operation === 'ADD'">
+                <label for="mode" class="col-md-3 control-label">Initial state</label>
+                <div class="col-md-9">
+                    <select required class="form-control" id="initialState" name="initialState" ng-model="subscription.state">
+                        <option value="ACTIVE" selected="selected">ACTIVE</option>
```
pwolaq: I think that using `selected="selected"` with `ng-model` is not necessary.

cristaloleg: But looks explicit, so no harm 😉

pwolaq: Ok, but what if someone changes initial `state` in factory? It will be misleading, unless also changed. Other `select`s rely only on `ng-model`.

cristaloleg: We can add a comment in a factory.

cristaloleg:
@pwolaq wdyt?  https://github.com/allegro/hermes/pull/939/commits/afee5e5596b0ef366bba962dd7f02161c22886cd

hermes-console/static/partials/modal/editSubscription.html
mtrojanowski: Actually I added it before I found out there's a Factory file. I think it's not needed and can be safely removed.

druminski:
Closing this PR as we don’t need it. We will back to it when few users need it.

PR #944, opened 2018-10-21
Updating info about simplified release flow

druminski:
For some time we are calling pull requests directly into **master** branch. This flow is simpler than previous one and works fine for us. Therefore I would like to make it official. In this PR I am updating Hermes `RELEASE_FLOW.md`, comments are welcome.


RELEASE_FLOW.md
```diff
@@ -35,27 +34,26 @@ use them on production environment. They also might not be documented.
 
 ### Creating enhancement/bugfix
 
-Branch from **rc** or **release-{version}** branch (if this is a bugfix for a scheduled release), code, merge back to
-the branch that you used in the first place.
+Branch from **master** branch, code, merge back to the branch that you used in the first place.
```
chemicL: in the first place = **master**, right? then "Branch from **master** branch, code, merge back when feature is ready."

chemicL: in the first place = **master**, right? then "Branch from **master** branch, code, merge back when feature is ready."

PR #955, opened 2018-11-23
Change format for subscription healthcheck endpoint

cristaloleg:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/SubscriptionsOwnershipEndpoint.java
```diff
@@ -19,38 +22,44 @@
...
+    @Produces(APPLICATION_JSON)
+    @Path("/{ownerSourceName}/{ownerId}")
+    public List<Subscription> listForOwner(@PathParam("ownerSourceName") String ownerSourceName, @PathParam("ownerId") String id) {
+        OwnerId ownerId = resolveOwnerId(ownerSourceName, id);
+        return subscriptionService.getForOwnerId(ownerId);
+    }
+
+    @GET
+    @Produces(APPLICATION_JSON)
+    @Path("/unhealthy")
```
druminski: To call this endpoint we have to send http request with following path `/subscriptions/owner/unhealthy`. This suggests that this endpoint is directly related to subscriptions. What about topics? In the near future we would like to add logic which will inform us about unhealthy topics as well. Therefore, I suggest to extract this endpoint to a new class, for example `UnhealthyStatusEndpoint` with path `/unhealthy` and to call it we will have to send http request on a an address `http://{hermes-management-ip}:{port}/unhealthy`. This will allow us implement new features without changing API.

PR #965, opened 2019-01-10
Updating kafka version to 2.0.0 in Vagrant

piotrrzysko:


vagrant_provisioning/install_kafka.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-KAFKA_VERSION=0.10.2.1
-SCALA_VERSION=2.10
+KAFKA_VERSION=2.0.0
+SCALA_VERSION=2.11
```
cristaloleg: Maybe 2.12.8 ? https://www.scala-lang.org/download/

cristaloleg: Ok, got it why it/s 2.11 https://kafka.apache.org/downloads#2.0.0

PR #970, opened 2019-01-17
Fixes metrics for batch and filtered subscriptions

piotrrzysko:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/filtering/FilteredMessageHandler.java
```diff
@@ -52,8 +52,8 @@ public void handle(FilterResult result, Message message, Subscription subscripti
         }
     }
 
-    protected void updateMetrics(Message message, Subscription subscription) {
-        metrics.meter(Meters.FILTERED_METER).mark();
+    private void updateMetrics(Subscription subscription) {
+        metrics.meter(Meters.FILTERED_METER, subscription.getTopicName(), subscription.getName()).mark();
```
dankraw: maybe we should keep both meters (aggregate + subscription-level)

PR #971, opened 2019-01-17
Use OpenJDK8 instead of OracleJDK8

cristaloleg:


cristaloleg:
Should we use OpenJDK 10 or 11 instead of 8 ? 🤔

adamdubiel:
Lets try running it on 11 on dev/test, if it works -> we can try & switch in main project as well.

cristaloleg:
Roger that 👍

PR #974, opened 2019-01-21
Setting kafka consumer max.poll.interval.ms

dankraw:
We have Kafka consumer's `max.poll.interval.ms` set to default value (5 minutes). 
This setting may lead to consmuer leaving the consumer group when rarely polling for new messages (longer message ttl, inflight queue full, etc) and not committing messages properly.

I see two options, setting the value to the value little bit higher than message-ttl configured for subscription (like 2 hours - the current maximum value for message ttl), or just using integer max value, so simply we'll rely on our consumer supervision mechanisms.  In this PR, I use the latter as the default value.

https://cwiki.apache.org/confluence/display/KAFKA/KIP-62%3A+Allow+consumer+to+send+heartbeats+from+a+background+thread


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaMessageReceiverFactory.java
```diff
@@ -121,6 +121,7 @@ public MessageReceiver createMessageReceiver(Topic topic,
         props.put(REQUEST_TIMEOUT_MS_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_REQUEST_TIMEOUT_MS_CONFIG));
         props.put(CONNECTIONS_MAX_IDLE_MS_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_CONNECTIONS_MAX_IDLE_MS_CONFIG));
         props.put(MAX_POLL_RECORDS_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_MAX_POLL_RECORDS_CONFIG));
+        props.put(MAX_POLL_INTERVAL_MS_CONFIG, configs.getIntProperty(Configs.KAFKA_CONSUMER_MAX_POLL_INTERVAL_CONFIG));
```
cristaloleg: Do we use `MAX_POLL_INTERVAL_MS_CONFIG` somewhere?

PR #993, opened 2019-03-11
resolve #991 | spring-boot upgrade in hermes-management

### LINKED ISSUE #991: Spring-boot update in Hermes-management
Hermes-management uses `org.springframework.boot:spring-boot-starter-{web, actuator, jersey}` dependencies in version `1.4.1.RELEASE`. They were update long time ago. We should update them to the latest version.

alasun:


integration/build.gradle
```diff
@@ -84,6 +84,8 @@ dependencies {
     testCompile group: 'org.eclipse.jetty', name: 'jetty-server', version: versions.jetty
     testCompile group: 'org.eclipse.jetty', name: 'jetty-servlet', version: versions.jetty
     testCompile group: 'org.eclipse.jetty', name: 'jetty-servlets', version: versions.jetty
+    testCompile group: 'com.google.code.gson', name: 'gson', version: '2.8.5'
```
druminski: Instead of adding this maybe we should update `javers-core` dependency to version with newer `gson`. What do you think about this?

alasun: Yes, we can do that ;) I will add a commit for this.

PR #995, opened 2019-03-21
Consumer groups in console

cristaloleg:


hermes-console/static/js/console/MetricStoreUrlResolver.js
```diff
@@ -70,4 +70,10 @@ function GraphiteUrlResolver(config) {
                 ]))
         };
     };
+
+    this.subscriptionDiagnostics = function (group, topic, subscription) {
+        return {
+            consumerGroups: compile('/topics/{}.{}/subscriptions/{}/consumer-groups', [group, topic, subscription]),
```
cristaloleg: Should we add something like `config.hermes.url` as a prefix here?

dankraw: if it's an url to hermes-management, then yes as it may be running on different port, etc.

PR #996, opened 2019-03-22
Attaching HTTP headers with subscription identity

piotrrzysko:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/JsonMessageBatch.java
```diff
@@ -36,21 +39,16 @@
     private boolean closed = false;
     private int retryCounter = 0;
 
-    public JsonMessageBatch(String id, ByteBuffer buffer, int size, int batchTime, Clock clock, List<Header> additionalHeaders) {
+    JsonMessageBatch(String id, ByteBuffer buffer, Subscription subscription, Clock clock) {
```
cristaloleg: Why it's not `public` anymore?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -88,6 +88,8 @@
     @Valid
     private SubscriptionOAuthPolicy oAuthPolicy;
 
+    private boolean subscriptionIdentityHeadersEnabled;
```
cristaloleg: Should we duplicate domain(`subscription*`) for this field?
WDYT about `identityHeadersEnabled`?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpRequestFactory.java
```diff
@@ -52,6 +53,11 @@ Request buildRequest(Message message, URI uri) {
                 .timeout(timeout, TimeUnit.MILLISECONDS)
                 .content(new BytesContentProvider(message.getData()));
 
+        if (message.hasSubscriptionIdentityHeaders()) {
+            request.header(TOPIC_NAME.getName(), message.getTopic());
```
faderskd: Do we need to different headers ? Maybe it would be enough if you use SUBSCRIPTION_NAME as fully qualified name ?

faderskd: I don't know the requirements, but if there is no any specific we can think about it ;)

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/JsonMessageBatch.java
piotrrzysko: `public` in this case was never needed in fact

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/ApacheHttpClientMessageBatchSender.java
```diff
@@ -64,6 +66,11 @@ public MessageSendingResult send(MessageBatch batch, URI address, int requestTim
         httpPost.addHeader(HTTP.CONTENT_TYPE, contentType.getMimeType());
         httpPost.addHeader(RETRY_COUNT.getName(), Integer.toString(batch.getRetryCounter()));
 
+        if (batch.hasSubscriptionIdentityHeaders()) {
```
faderskd: Do we need to check this condition every time during sending message ? You can check condition and add required headers during subscription initialization only once :) This way you always would just iterate over headers. WDYT ?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
piotrrzysko: I think that in this case it is better to add `subscription` prefix because it makes it cleaner what kind of identity we are going to attach to message.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/JsonMessageBatch.java
cristaloleg: backward compatibility?

PR #1000, opened 2019-03-27
Create LICENSE

cristaloleg:


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
druminski: @cristaloleg read this. We only need introduced boilerplate notice instead of copying whole license. We need something like this: https://github.com/allegro/vaas/blob/master/LICENSE

PR #1005, opened 2019-04-03
Enable setting idleTimeout for Jetty client

### LINKED ISSUE #1004: Enable setting Jetty idleTimeout for connections
Actually we have no option to set connection idle timeout for jetty message sender. This timeout indicates how long connection can be unused before it is closed by Jetty client. It can be useful to preventing server from killing silently the connection. When it happens, you can get 408 error, because you are using connection already closed by the server. Setting idleTimeout for jetty client to time lower than server idle timeout will cause the connection to be re-established before it will expire on the server site.

faderskd:
This PR fixes #1004

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
@@ -181,12 +181,12 @@ public enum Configs {
     CONSUMER_HTTP_CLIENT_VALIDATE_CERTS("consumer.http.client.validate.certs", true),
     CONSUMER_HTTP_CLIENT_VALIDATE_PEER_CERTS("consumer.http.client.validate.peer.certs", true),
     CONSUMER_HTTP_CLIENT_ENABLE_CRLDP("consumer.http.client.enable.crldp", true),
-    CONSUMER_HTTP_CLIENT_IDLE_TIMEOUT("consumer.http.client.idleTimeout", 0),
+    CONSUMER_HTTP_CLIENT_IDLE_TIMEOUT("consumer.http.client.idleTimeout", 30_000),
 
     CONSUMER_HTTP2_ENABLED("consumer.http2.enabled", true),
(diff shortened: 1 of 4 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -181,11 +181,13 @@
     CONSUMER_HTTP_CLIENT_VALIDATE_CERTS("consumer.http.client.validate.certs", true),
     CONSUMER_HTTP_CLIENT_VALIDATE_PEER_CERTS("consumer.http.client.validate.peer.certs", true),
     CONSUMER_HTTP_CLIENT_ENABLE_CRLDP("consumer.http.client.enable.crldp", true),
-
+    CONSUMER_HTTP_CLIENT_IDLE_TIMEOUT("consumer.http.client.idleTimeout", 0),
```
faderskd: The 0 value will apply default idleTimeout for jetty client. It will be 30000ms

faderskd: Maybe it will be better to set it explicitly

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -181,11 +181,13 @@
     CONSUMER_HTTP_CLIENT_VALIDATE_CERTS("consumer.http.client.validate.certs", true),
     CONSUMER_HTTP_CLIENT_VALIDATE_PEER_CERTS("consumer.http.client.validate.peer.certs", true),
     CONSUMER_HTTP_CLIENT_ENABLE_CRLDP("consumer.http.client.enable.crldp", true),
-
+    CONSUMER_HTTP_CLIENT_IDLE_TIMEOUT("consumer.http.client.idleTimeout", 30_000),
```
cristaloleg: I thought the proposition was to set 19 sec as a timeout.

faderskd: Yes but it is good only for us ;) We will configure it configuration files.

PR #1006, opened 2019-04-04
Set idleTimeout for Jetty http client

faderskd:
This PR fixes https://github.com/allegro/hermes/issues/1004


hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
```diff
@@ -15,6 +15,7 @@
     private static final int DEFAULT_MESSAGE_TTL = 3600;
     private static final int DEFAULT_MESSAGE_BACKOFF = 100;
     private static final int DEFAULT_REQUEST_TIMEOUT = 1000;
+    private static final int DEFAULT_SOCKET_TIMEOUT = 0;
```
faderskd: When you set it to 0 it will be no socket timeout at all

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyMessageSenderTest.java
```diff
@@ -170,10 +169,10 @@ public void shouldSendAuthorizationHeaderIfAuthorizationProviderAttached() {
     }
 
     @Test
-    public void shouldUseSuppliedTimeout() throws ExecutionException, InterruptedException, TimeoutException {
+    public void shouldUseSuppliedRequestTimeout() throws ExecutionException, InterruptedException, TimeoutException {
         // given
         HttpRequestFactory httpRequestFactory = new HttpRequestFactory(client,
-                100,
+                100,1000,
```
cristaloleg:
```suggestion
                100, 1000,
```

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
faderskd: So currently created subscriptions are unaffected. With this default option, only subscription's request timeout will be respected

hermes-consumers/src/test/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/JettyMessageSenderTest.java
faderskd: Intellij arguments hints :D

hermes-api/src/main/java/pl/allegro/tech/hermes/api/OAuthProviderWithOptionalSocketTimeout.java
```diff
@@ -0,0 +1,43 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Optional;
+
+public class OAuthProviderWithOptionalSocketTimeout {
```
dankraw: do we really need this class (a class with such explicit name)? :) I mean, can't we just add the optional socket timeout to the existing one?

faderskd: This class is only needed to be able to fetch existing zookeeper saved oauth providers. Its implementation detail. BTW this class should be placed in infrastructure package.

faderskd: done

PR #1028, opened 2019-05-28
Improved handling unavailable metrics

piotrrzysko:


hermes-api/src/main/java/pl/allegro/tech/hermes/api/MetricDecimalValue.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private final boolean unavailable;
+    private final String value;
+
+    private MetricDecimalValue(boolean unavailable, String value) {
+        this.unavailable = unavailable;
+        this.value = value;
+    }
+
+    public static MetricDecimalValue unavailable() {
```
cristaloleg: Why we cannot make `UNAVAILABLE` as a public field and just use it?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MetricLongValue.java
```diff
@@ -0,0 +1,66 @@
...
+    private MetricLongValue(boolean unavailable, long value) {
+        this.unavailable = unavailable;
+        this.value = value;
+    }
+
+    public static MetricLongValue unavailable() {
+        return UNAVAILABLE;
+    }
+
+    public static MetricLongValue of(long lag) {
```
dankraw: lag -> value ? :)

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MetricDecimalValue.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.Objects;
+
+import static java.lang.Double.parseDouble;
+
+public class MetricDecimalValue {
+    private static final String UNAVAILABLE_STRING = "unavailable";
+    private static final MetricDecimalValue UNAVAILABLE = new MetricDecimalValue(true, "-1.0");
+
+    private final boolean unavailable;
```
cristaloleg: Storing non-negated field might be more readable, wdyt?

cristaloleg: or not

hermes-api/src/main/java/pl/allegro/tech/hermes/api/MetricDecimalValue.java
piotrrzysko: In my opinion `unavailable()` is better in this case because it is consistent with other factory method - `of(...)`.

PR #1033, opened 2019-06-10
Zk multi dc

cristaloleg:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/retransmit/KafkaRetransmissionService.java
```diff
@@ -51,7 +56,10 @@ public KafkaRetransmissionService(
                 PartitionOffset partitionOffset = new PartitionOffset(k.name(), offset, partitionId);
                 partitionOffsetList.add(partitionOffset);
                 if (!dryRun) {
-                    subscriptionOffsetChange.setSubscriptionOffset(topic.getName(), subscription, brokersClusterName, partitionOffset);
+                    multiDcExecutor.execute(new SetSubscriptionOffsetRepositoryCommand(topic.getName(), subscription,
```
druminski: Actually, we don’t have to write info about offsets from one DC to all DCs. If offset is from DC X then we only have to write it to Zookeeper in exactly this DC X, not to other ones.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/zookeeper/ZookeeperRepositoryManager.java
```diff
@@ -0,0 +1,128 @@
...
+        return getRepositoriesByType(repositoryType)
+                .entrySet()
+                .stream()
+                .map(entry -> new DcBoundRepositoryHolder<>(entry.getValue(), entry.getKey()))
+                .collect(Collectors.toList());
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> Map<String, T> getRepositoriesByType(Class<T> type) {
+        if (type == GroupRepository.class) {
```
druminski: We can also put these repositories in Map<Class<?>, Object> and get them by class key. During the get operation we will have to cast value to (Map<String, T>) but thanks to this we will be able to remove this big block of ifs.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
```diff
@@ -156,7 +208,18 @@ public SubscriptionHealth getSubscriptionHealth(TopicName topicName, String subs
...
+        List<SentMessageTrace> traces = new ArrayList<>();
+        for(DcBoundRepositoryHolder<UndeliveredMessageLog> holder : holders) {
+            try {
+                holder.getRepository().last(topicName, subscriptionName).ifPresent(traces::add);
+            } catch (Exception e) {
+                logger.warn("Could not load latest undelivered message from DC: {}", holder.getDcName());
+            }
+        }
+        return traces.stream().max(Comparator.comparing(SentMessageTrace::getTimestamp));
+//        return undeliveredMessageLog.last(topicName, subscriptionName);
```
druminski: Some leftover.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/storage/StorageConfiguration.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Bean
+    public DistributedEphemeralCounter distributedCounter() {
+        CuratorFramework curatorFramework = clientManager().getLocalClient().getCuratorFramework();
+        return new DistributedEphemeralCounter(curatorFramework);
+    }
+
+    @Bean
+    GroupRepository groupRepository() {
```
druminski: These below repositories are misleading now as they still allow for making changes in local DC. Because of this, a developer instead of using *Command approach and making a change in all DCs, can by mistake make only change in single DC. I’m wondering how to solve this. Maybe we should remove write operation from local repositories or at least add 'local' prefix to the repository names.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/storage/StorageConfiguration.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Bean
+    AdminTool adminTool() {
+        ZookeeperClient localClient = clientManager().getLocalClient();
+        return new ZookeeperAdminTool(zookeeperPaths(), localClient.getCuratorFramework(),
+                objectMapper, Configs.ADMIN_REAPER_INTERAL_MS.getDefaultValue());
+    }
+
+    @Bean
+    UndeliveredMessageLog undeliveredMessageLog() {
```
druminski: We don’t need this bean anymore as it stopped being used. Now we have ZookeeperUndeliveredMessageLog per cluster.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/storage/StorageConfiguration.java
cristaloleg: Discussed face to face, we're going to postpone this, currently it isn't a blocker, rather than 'nice to have' feature.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/kafka/KafkaConfiguration.java
```diff
@@ -115,6 +129,22 @@ MultiDCAwareService multiDCAwareService(KafkaNamesMappers kafkaNamesMappers, Sch
...
     }
 
+    private SubscriptionOffsetChangeIndicator getRepository(
+            List<DcBoundRepositoryHolder<SubscriptionOffsetChangeIndicator>> repostories,
+            KafkaProperties kafkaProperties) {
+        return repostories.stream()
+                .filter(x -> kafkaProperties.getDc().equals(x.getDcName()))
+                .findFirst().orElseThrow(() ->
+                        new IllegalArgumentException(
+                                String.format("Kafka brokers cluster name %s not matched with Zookeeper cluster name %s",
```
druminski: It would be nice to update this message since you have changed filter predicate. Maybe something like this will be fine: `"Kafka cluster dc name '%s'  not matched with Zookeeper dc names `%s`", kafkaProperties.getDc(), ...`

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/group/commands/CreateGroupRepositoryCommand.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public Class<GroupRepository> getRepositoryType() {
+        return GroupRepository.class;
+    }
+
+    @Override
+    public String toString() {
+        return "CreateGroup(" + group.getGroupName() + ")";
```
dankraw: I guess, those `toString()` methods for commands could be done automatically using class name (substring etc.), placed in parent `RepositoryCommand` maybe. :)

Something like:

```java
// base class
private final static int NAME_LENGTH = RepositoryCommand.class.getSimpleName().length();

protected String printableName(String id) {
    String name = this.getClass().getSimpleName();
    return name.substring(0, name.length() - NAME_LENGTH) + "(" + id + ")";
}

@Override
public String toString() {
    return printableName(""); // default without an identifier, not sure if needed at all
}
```

```java
// command class
@Override
public String toString() {
    return printableName(provider.getName());
}
```

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
```diff
@@ -94,17 +110,49 @@ public SubscriptionService(SubscriptionRepository subscriptionRepository,
...
     public Subscription getSubscriptionDetails(TopicName topicName, String subscriptionName) {
-        return subscriptionRepository.getSubscriptionDetails(topicName, subscriptionName).anonymize();
+        Subscription subscription = subscriptionRepository.getSubscriptionDetails(topicName, subscriptionName)
+                .anonymize();
+        subscription.setState(getEffectiveState(topicName, subscriptionName));
+        return subscription;
+    }
+
+    private Subscription.State getEffectiveState(TopicName topicName, String subscriptionName) {
+        Set<Subscription.State> states = loadSubscriptionStatesFromAllDc(topicName, subscriptionName);
```
dankraw: how about logging a warning if the set size is > 1 (which means something is out of sync for this sub)?

cristaloleg: `logger.warn("Some states are out of sync: {}", states);` ?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDcRepositoryCommandExecutor.java
```diff
@@ -0,0 +1,76 @@
...
+        List<DcBoundRepositoryHolder<T>> repositories = repositoryManager.getRepositories(command.getRepositoryType());
+        List<DcBoundRepositoryHolder<T>> succeededRepositories = new ArrayList<>();
+
+        for (DcBoundRepositoryHolder<T> repository : repositories) {
+            try {
+                command.execute(repository.getRepository());
+                succeededRepositories.add(repository);
+            } catch (Exception e) {
+                if (rollbackEnabled) {
+                    rollback(succeededRepositories, command);
```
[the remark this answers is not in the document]
cristaloleg: Right now rollback is done for all the repositories (loop inside a `rollback` func). Should we make per repo - dunno :(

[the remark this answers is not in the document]
dankraw: I mean, we pass `succeededRepositories` here, and then for each of them we call `rollback()`. The actual exception was thrown because of a failure related to the "current" `repository`, for which we don't call `rollback()`.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java
dankraw: Yep, something alike. But with full subscription name included and operation related.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/group/commands/CreateGroupRepositoryCommand.java
cristaloleg: Good point, but I would like to leave these 16 classes as they're :P

dankraw: cool :)

hermes-common/src/main/java/pl/allegro/tech/hermes/domain/NodePassword.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public int hashCode() {
+        return Arrays.hashCode(hashedPassword);
+    }
+
+    public static NodePassword fromString(String string) {
+        return new NodePassword(string);
+    }
+
+    public static String getRandomPassword() {
```
dankraw: is it used anywhere?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/kafka/KafkaConfiguration.java
```diff
@@ -108,10 +117,30 @@ MultiDCAwareService multiDCAwareService(KafkaNamesMappers kafkaNamesMappers, Sch
...
                 ofMillis(subscriptionProperties.getIntervalBetweenCheckinIfOffsetsMovedInMillis()),
-                ofSeconds(subscriptionProperties.getOffsetsMovedTimeoutInSeconds()));
+                ofSeconds(subscriptionProperties.getOffsetsMovedTimeoutInSeconds()),
+                multiDcExecutor);
+    }
+
+    private SubscriptionOffsetChangeIndicator getRepository(
+            List<DcBoundRepositoryHolder<SubscriptionOffsetChangeIndicator>> repostories,
+            KafkaProperties kafkaProperties) {
+        if (repostories.size() == 1) {
```
dankraw: since we have a`stream/firstFirst` below - do we need those lines?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/kafka/KafkaProperties.java
```diff
@@ -2,6 +2,8 @@
 
 public class KafkaProperties {
 
+    private String dc = "dc";
```
dankraw: Can we change the property name to `datacenter` ?
So we'd have `getDatacenter()` etc.

I'm afraid in the future people will ask: `Does dc stand for datacenter?` so it'll be clear. :)

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/storage/StorageClustersProperties.java
```diff
@@ -1,18 +1,28 @@
...
+public class StorageClustersProperties {
 
     private String pathPrefix = "/hermes";
-    private String connectionString = "localhost:2181";
-    private int sessionTimeout = 10000;
-    private int connectTimeout = 1000;
     private int retryTimes = 3;
     private int retrySleep = 1000;
     private int sharedCountersExpiration = 72;
+    private DcNameSource dcNameSource;
```
dankraw: Same here, I'd rename `DcNameSource` to `DatacenterNameSource` in class name as well as the property name.

PR #1055, opened 2019-07-02
Update CHANGELOG.md

cristaloleg:


CHANGELOG.md
```diff
@@ -2,6 +2,40 @@
 
 ### ...
 
+## 1.1.0 (02.07.2019)
+
+### Enhancements
+
+#### ([1033](https://github.com/allegro/hermes/pull/1033)) Zookeeper multi datacenter
+
+Starting from this release it's possible to run Hermes on multiple independent Zookeeper clusters.
```
druminski: I would add something like this: Changes are only in hermes-management module which from now on can operate on multiple Zookeeper clusters.

CHANGELOG.md
```diff
@@ -2,6 +2,40 @@
...
+
+To match Kafka clusters with Zookeeper clusters we've an additional field in a config called `datacenter`.
```
druminski: a config -> a hermes-management config

CHANGELOG.md
```diff
@@ -2,6 +2,40 @@
...
+
+This field allows you to tell Hermes which Zookeeper clusters belong to which Kafka.
+
+Example:
```
druminski: When there is only one zookeeper cluster then it will be always used, example

CHANGELOG.md
```diff
@@ -2,6 +2,40 @@
...
+
+storage:
+  pathPrefix: /run/hermes
+  clusters:
+    - 
+      datacenter: dc
+      clusterName: zk
+      connectionString: zookeeper:2181
+```
+
```
druminski: Otherwise, Zookeeper cluster is matched with Kafka cluster by datacenter field.

PR #1060, opened 2019-07-06
Added support for java 11

lukidzi:
Build passed with openjdk version "11.0.3" locally

### DIFF
--- a/hermes-api/build.gradle
+++ b/hermes-api/build.gradle
@@ -12,7 +12,6 @@ dependencies {
     compile group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.11'
     compile group: 'com.sun.xml.bind', name: 'jaxb-core', version: '2.2.11'
     compile group: 'com.sun.xml.bind', name: 'jaxb-impl', version: '2.2.11'
-    compile group: 'javax.activation', name: 'activation', version: '1.1.1'
 
     testCompile group: 'org.spockframework', name: 'spock-core', version: versions.spock
     testCompile project(':hermes-test-helper')
--- a/hermes-api/build.gradle
+++ b/hermes-api/build.gradle
@@ -9,6 +9,10 @@ dependencies {
     compile group: 'com.fasterxml.jackson.jaxrs', name: 'jackson-jaxrs-json-provider', version: versions.jackson
     compile group: 'com.google.guava', name: 'guava', version: versions.guava
     compile group: 'com.damnhandy', name: 'handy-uri-templates', version: '2.0.2'
+    compile group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.11'
+    compile group: 'com.sun.xml.bind', name: 'jaxb-core', version: '2.2.11'
+    compile group: 'com.sun.xml.bind', name: 'jaxb-impl', version: '2.2.11'
+    compile group: 'javax.activation', name: 'activation', version: '1.1.1'
 
     testCompile group: 'org.spockframework', name: 'spock-core', version: versions.spock
     testCompile project(':hermes-test-helper')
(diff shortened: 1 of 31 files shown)


druminski:
It would be nice to use [openjdk11 in travis](https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later)

hermes-api/build.gradle
```diff
@@ -9,6 +9,10 @@ dependencies {
     compile group: 'com.fasterxml.jackson.jaxrs', name: 'jackson-jaxrs-json-provider', version: versions.jackson
     compile group: 'com.google.guava', name: 'guava', version: versions.guava
     compile group: 'com.damnhandy', name: 'handy-uri-templates', version: '2.0.2'
+    compile group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.11'
```
druminski: Do we need all of these dependencies?

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/consumer/filtering/FilterChainSpec.groovy
```diff
@@ -129,7 +129,7 @@ class FilterChainSpec extends Specification {
         def order
 
         RecordingSubscriptionMessageFilterCompiler(String type, AtomicInteger counter) {
-            super(type, { m -> test(m) })
+            super(type, null)
```
druminski: Why did you change it to null?

lukidzi: With newer version of groovy-spock lib you cannot pass method of instance which hasn't been constructed yet.

hermes-api/build.gradle
lukidzi: Those are java ee dependencies removed from jdk 11 and they are need in this package

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/consumer/filtering/FilterChainSpec.groovy
dankraw: so the 2nd argument wasn't used at all?

lukidzi: Yes it wasn't, this overrided method was called because  RecordingSubscriptionMessageFilterCompiler extends MessageFilter

hermes-common/src/main/java/pl/allegro/tech/hermes/common/ssl/SslContextFactory.java
```diff
@@ -1,8 +1,11 @@
 package pl.allegro.tech.hermes.common.ssl;
 
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
 
 public interface SslContextFactory {
 
     SSLContext create();
+
+    TrustManager[] getTrustManagers();
```
dankraw: As we talked already, we could change this interface, so it would have a single method:
`SSLContextHolder create()` or something similar... :)
and `SSLContextHolder` would hold both the `TrustManagers[]` and `SSLContext` instances. This way we explicitly say that these object always stick together and are created at the same time by the factory.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/ssl/SSLContextHolder.java
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+
+public final class SSLContextHolder {
+
+    private final SSLContext sslContext;
+    private final TrustManager[] trustManagers;
+
+    public SSLContextHolder(SSLContext sslContext) {
```
dankraw: Maybe we should remove this constructor?

dankraw:
@lukidzi 👏 👏

PR #1061, opened 2019-07-08
Introduced HdrHistogram-based metrics reservoir

piotrrzysko:


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/MetricsReservoirType.java
```diff
@@ -0,0 +1,5 @@
+package pl.allegro.tech.hermes.common.metric;
+
+public enum MetricsReservoirType {
```
cristaloleg: WDYT to move this to `MetricsReservoirType` ?

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 
+### Enhancements
+#### ([1061](https://github.com/allegro/hermes/pull/1061)) Introduced HdrHistogram-based metrics reservoir
+
+Introduced [HdrHistogram](http://hdrhistogram.org/)–based implementation of reservoir in metrics.
+To switch between implementations, use `metrics.reservoir.type` option:
+
+* `exponentially_decaying` - exponentially decaying reservoir (previous version, currently default)
+* `hdr` - HdrHistogram–based reservoir 
+
```
druminski: I would add an information why it’s worth to consider using hdr histogram. Something like: 'HdrHistogram-based reservoir has much less memory footprint than exponentially decaying. Because of this GC in Hermes has less job to do which leads to better Hermes performance. In the near future HDR will be default Hermes metrics reservoir.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/MetricsReservoirType.java
piotrrzysko: I would like to be consistent with the current implementation of similar case: `pl.allegro.tech.hermes.common.schema.SchemaRepositoryType`.

PR #1068, opened 2019-07-17
Fixing integration tests

druminski:
Changes: 
* enabled reuse address flag in port randomization ([doc](https://docs.oracle.com/javase/8/docs/api/java/net/ServerSocket.html#setReuseAddress-boolean-)).
* `HierarchicalCacheLevel` removes detected empty nodes. Empty nodes can be created after topic/subscription deletion due to race for example in zookeeper metrics reporter. From now on no more topics/subscriptions leftovers are generated under tree `/groups/{groupName}/topics/{topicName}/subscriptions/{subscriptionName}`
* all topics and subscriptions after each integration test are removed. To do this schema removal had to be enabled.
* fixed integration test related to migration from json to avro by removing secondary zookeeper cluster from `hermes-management` configuration.
* used randomTopic() in integration tests. Thanks to `randomTopic()` test retry has much higher chance to pass.
* handled potential exception in `HermesIntegrationEnvironment.java`

Due to many changes in this PR I recommend to review them by commits.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
@@ -121,7 +121,7 @@ class HierarchicalCacheLevel extends PathChildrenCache implements PathChildrenCa
                     path, children.stream().map(Object::toString).collect(Collectors.joining(",")));
             printChildrenWithEmptyParentRecursively(path, children);
         } catch (KeeperException.NoNodeException e) {
-            // probably already removed
+            logger.info("Could not receive list of children for path {} as the path does not exist", path);
         } catch (Exception e) {
             logger.warn("Could not receive list of children for path {} due to error", path, e);
         }
@@ -136,7 +136,7 @@ class HierarchicalCacheLevel extends PathChildrenCache implements PathChildrenCa
 
                 printChildrenWithEmptyParentRecursively(nextPath, nextChildren);
             } catch (KeeperException.NoNodeException e) {
-                // probably already removed
+                logger.info("Could not receive list of children for path {} as the path does not exist", pathWithEmptyParent);
             } catch (Exception e) {
                 logger.warn("Could not receive list of children for path {} due to error", pathWithEmptyParent, e);
             }
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
@@ -160,7 +160,7 @@ class HierarchicalCacheLevel extends PathChildrenCache implements PathChildrenCa
         try {
             HierarchicalCacheLevel subcache = subcacheMap.remove(cacheName);
             if (subcache == null) {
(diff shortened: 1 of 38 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
```diff
@@ -36,11 +42,14 @@
                            String path,
                            int depth,
                            CacheListeners eventConsumer,
-                           Optional<BiFunction<Integer, String, HierarchicalCacheLevel>> nextLevelFactory) {
+                           Optional<BiFunction<Integer, String, HierarchicalCacheLevel>> nextLevelFactory,
```
mkopylec: Maybe it would be better to create some value object instead of using `Optional<BiFunction<Integer, String, HierarchicalCacheLevel>>`

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
```diff
@@ -94,12 +112,51 @@ private void addSubcache(String path, String cacheName) throws Exception {
...
+    private void printChildrenWithEmptyParentRecursively(String pathWithEmptyParent, List<String> children) {
+        children.forEach( c -> {
+            try {
+                String nextPath = pathWithEmptyParent + "/" + c;
+                logger.warn("Node with empty parent: {}", nextPath);
+                List<String> nextChildren = curatorClient.getChildren().forPath(nextPath);
+
+                printChildrenWithEmptyParentRecursively(nextPath, nextChildren);
+            } catch (KeeperException.NoNodeException e) {
+                // probably already removed
```
mkopylec: "probably" tells that you are not 100% sure :)
IMO the exception should be logged.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/SubscriptionAssignmentCache.java
```diff
@@ -60,7 +61,7 @@ public SubscriptionAssignmentCache(CuratorFramework curator,
         this.pathSerializer = new SubscriptionAssignmentPathSerializer(basePath, AUTO_ASSIGNED_MARKER);
         ThreadFactory threadFactory = new ThreadFactoryBuilder().setNameFormat("subscription-assignment-cache-%d").build();
         this.cache = new HierarchicalCache(
-                curator, Executors.newSingleThreadScheduledExecutor(threadFactory), basePath, 2, Collections.emptyList()
+                curator, Executors.newSingleThreadScheduledExecutor(threadFactory), basePath, 2, emptyList(), false
```
mkopylec: IMO `HierarchicalCacheBuidler` should be introduced.

druminski: Thanks for the hint, although as you mentioned this is big PR, so I would like not to introduce further changes not related to the scope of the issue which is fix of integration tests.

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/cache/HierarchicalCacheLevel.java
druminski: Ok, I’ve added log here.

Issue #1008, opened 2019-04-07
Original message available on HermesResponse when it wasn't published


rafal-glowinski:
When HermesMessage cannot be published and all retries are exhausted, client returns `CompletableFuture<HermesResponse>` and we can query the `HermesResponse` to see if it was a successful publication or not.

I think it would be useful to be able to retrieve the message that was not published using some additional method: `HermesResponse.failedMessage()`. 

Since `HermesClient` is asynchronous, it would be cleaner to have one, default, way of getting the failed message than have everyone to bind it somehow to async execution context.

jewertow:
Resolved in #1073.

PR #1073, opened 2019-07-23
Original message available on HermesResponse when it wasn't published.

jewertow:
#1008

### DIFF
--- a/hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesMessage.java
+++ b/hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesMessage.java
@@ -1,9 +1,11 @@
 package pl.allegro.tech.hermes.client;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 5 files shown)


hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesMessage.java
```diff
@@ -124,6 +126,23 @@ public void consumeHeaders(BiConsumer<String, String> consumer) {
         headers.forEach(consumer);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
adamdubiel: why not use Objects class for both equals and dashcode?

adamdubiel: and why do you need eq/hashcode in this class at all? I don't see other changes in code that would require this

jewertow: I removed this code: https://github.com/allegro/hermes/pull/1077

PR #1078, opened 2019-07-31
Subscription ids cache

dankraw:
Introducing a subscription id cache, based on [Znode czxid](https://zookeeper.apache.org/doc/r3.5.5/zookeeperProgrammers.html#sc_zkDataModel_znodes).
Subscription ids will be used for subscription identification in the underlying storage  instead of plain strings as  it is currently.


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCache.java
```diff
@@ -0,0 +1,58 @@
...
+import pl.allegro.tech.hermes.domain.notifications.InternalNotificationsBus;
+import pl.allegro.tech.hermes.domain.notifications.SubscriptionCallback;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class NotificationAwareSubscriptionIdsCache implements SubscriptionIds, SubscriptionCallback {
+
+    private final Map<SubscriptionId, SubscriptionName> ids = new HashMap<>();
```
jewertow: Don't we have to care for thread safety?

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCacheTest.groovy
```diff
@@ -0,0 +1,92 @@
...
+        def sub2 = subscriptionNames.next()
+        def sub1Id = SubscriptionId.of(sub1)
+        def sub2Id = SubscriptionId.of(sub2)
+        subscriptionCache.listActiveSubscriptionNames() >> []
+
+        when:
+        subscriptionIds.start()
+
+        then:
+        !subscriptionIds.getSubscriptionName(sub1Id).ifPresent()
```
jewertow: `ifPresent()`? I suspect you wanted to use `.isPresent()`.

dankraw: ouch, good catch.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCache.java
dankraw: Hmm it supposed to be used by a single thread, but as long as it's not used anywhere yet, I think it should allow concurrent access. Good point, I'll fix it, thanks. :)

hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCacheTest.groovy
dankraw: fixed.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCache.java
dankraw: Fixed.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/ZookeeperSubscriptionIdProvider.java
```diff
@@ -0,0 +1,39 @@
...
+        this.curatorFramework = curatorFramework;
+        this.zookeeperPaths = zookeeperPaths;
+    }
+
+    @Override
+    public SubscriptionId getSubscriptionId(SubscriptionName name) {
+        return Optional.ofNullable(getZnodeStat(name))
+                .map(Stat::getCzxid)
+                .map(czxid -> SubscriptionId.from(name, czxid))
+                .orElseThrow(() -> new IllegalStateException(
```
jewertow: I'm not sure if it's appropriate use of `IllegalStateException`. 
Documentation says: "Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation."
Maybe could we throw dedicated exception?

dankraw: IMHO it fits actually - when there is no znode related to specific subscription, it means the method was called in inappropriate time, probably after the subscription was removed (or was not created yet, but it's impossible).
I will rethink your suggestion, thanks.

PR #1084, opened 2019-08-10
hermes-client supports WebClient

jewertow:
Since `AsyncRestTemplate` is deprecated, hermes client should provide implementation of `HermesSender` based on `WebClient`.
Even though the `WebClient` is a reactive client and returns reactive types (Mono or Flux), it is easy to convert a response to `CompletableFuture`.

### DIFF
--- a/hermes-client/src/main/java/pl/allegro/tech/hermes/client/restTemplate/RestTemplateHermesSender.java
+++ b/hermes-client/src/main/java/pl/allegro/tech/hermes/client/restTemplate/RestTemplateHermesSender.java
@@ -66,7 +66,7 @@ public class RestTemplateHermesSender implements HermesSender {
                 .build();
     }
 
-    public static TreeMap<String, String> convertToCaseInsensitiveMap(Map<String, String> hashMap) {
+    private TreeMap<String, String> convertToCaseInsensitiveMap(Map<String, String> hashMap) {
         return hashMap.entrySet().stream()
                 .collect(toMap(
                         Map.Entry::getKey,
--- a/hermes-client/src/main/java/pl/allegro/tech/hermes/client/restTemplate/RestTemplateHermesSender.java
+++ b/hermes-client/src/main/java/pl/allegro/tech/hermes/client/restTemplate/RestTemplateHermesSender.java
@@ -66,7 +66,7 @@ public class RestTemplateHermesSender implements HermesSender {
                 .build();
     }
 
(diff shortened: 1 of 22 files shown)


hermes-client/src/main/java/pl/allegro/tech/hermes/client/restTemplate/RestTemplateHermesSender.java
```diff
@@ -66,7 +66,7 @@ private HermesResponse fromHttpStatusCodeException(HttpStatusCodeException excep
                 .build();
     }
 
-    private TreeMap<String, String> convertToCaseInsensitiveMap(Map<String, String> hashMap) {
+    public static TreeMap<String, String> convertToCaseInsensitiveMap(Map<String, String> hashMap) {
```
adamdubiel: you introduced dependency between WebClient and RestTemplate - why? do they really depend on each other? making this method public means it becomes part of a contract and can't be changed. as Go proverb says: "A little copying is better than a little dependency."

jewertow: You're right. I fixed it.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/webclient/WebClientHermesSender.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public CompletableFuture<HermesResponse> send(URI uri, HermesMessage message) {
+        final Mono<ClientResponse> clientResponse = webClient.post()
+                .uri(uri)
+                .syncBody(message.getBody())
+                .headers(httpHeaders -> httpHeaders.setAll(message.getHeaders()))
+                .exchange();
+
+        return clientResponse
```
mkopylec: This could probably be shorten to:
```java
clientRespponse.flatMap(response -> response
                        .bodyToMono(String.class)
                        .map(body -> hermesResponse()
                                .withBody(body)
                                .withHttpStatus(response.rawStatusCode())
                                ...))
                .toFuture();
```

dankraw: plus we need something like:

```
.switchIfEmpty(NO_BODY)
```

where
```
 private final static Mono<String> NO_BODY = Mono.just("");
```

Map will not be called if there's no body. 
You should check how frontends behave, but I guess they send no body on 2xx response.

PR #1089, opened 2019-08-22
Add topic metadata refresh job

krzysztofpcy:


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingJob.java
```diff
@@ -0,0 +1,53 @@
...
+                                   ConfigFactory config) {
+        this(topicMetadataLoadingRunner, config.getIntProperty(Configs.FRONTEND_TOPIC_METADATA_LOADING_JOB_INTERVAL));
+    }
+
+    TopicMetadataLoadingJob(TopicMetadataLoadingRunner topicMetadataLoadingRunner, int intervalSeconds) {
+        this.topicMetadataLoadingRunner = topicMetadataLoadingRunner;
+        this.intervalSeconds = intervalSeconds;
+
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat("TopicMetadataLoadingExecutor-%d").build();
```
cristaloleg: 👍

cristaloleg: BTW, should we name it as `TopicMetadataLoadingJob` ?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -148,6 +148,9 @@
     FRONTEND_STARTUP_TOPIC_METADATA_LOADING_RETRY_COUNT("frontend.startup.topic.metadata.loading.retry.count", 5),
     FRONTEND_STARTUP_TOPIC_METADATA_LOADING_THREAD_POOL_SIZE("frontend.startup.topic.metadata.loading.thread.pool.size", 16),
 
+    FRONTEND_TOPIC_METADATA_LOADING_JOB_ENABLED("frontend.startup.topic.metadata.job.enabled", false),
+    FRONTEND_TOPIC_METADATA_LOADING_JOB_INTERVAL("frontend.startup.topic.metadata.job.interval", 60),
```
dankraw: I think the `startup` word just doesn't fit well here, as it runs all the time. :)

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -148,6 +148,9 @@
     FRONTEND_STARTUP_TOPIC_METADATA_LOADING_RETRY_COUNT("frontend.startup.topic.metadata.loading.retry.count", 5),
     FRONTEND_STARTUP_TOPIC_METADATA_LOADING_THREAD_POOL_SIZE("frontend.startup.topic.metadata.loading.thread.pool.size", 16),
 
+    FRONTEND_TOPIC_METADATA_REFRESH_JOB_ENABLED("frontend.topic.metadata.refresh.job.enabled", true),
+    FRONTEND_TOPIC_METADATA_REFRESH_JOB_INTERVAL("frontend.topic.metadata.refresh.job.interval", 60),
```
dankraw: can we change it to `...JOB_INTERVAL_SECONDS` and `...interval.seconds`? :)

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingJob.java
```diff
@@ -0,0 +1,53 @@
...
+        this.executorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+    }
+
+    @Override
+    public void run() {
+        topicMetadataLoadingRunner.refreshMetadata();
+    }
+
+    public void start() {
+        job = executorService.scheduleAtFixedRate(this, intervalSeconds, intervalSeconds, TimeUnit.SECONDS);
```
jewertow: I think if `initialDelay` (second argument) would be equal to 0 instead of `intervalSeconds` then we wouldn't need `TopicMetadataLoadingStartupHook`.

dankraw: `TopicMetadataLoadingStartupHook` plays a specific role when booting frontends, it blocks readiness health checks from becoming green, i.e. it is required to load all topics metadata before we start accepting incoming traffic. Here it would be just async, so it's a different case.

krzysztofpcy: You can disable refreshing job by setting `FRONTEND_TOPIC_METADATA_REFRESH_JOB_ENABLED` to `false`, but metadata should always be loaded on startup

dankraw: actually there's `frontend.startup.topic.metadata.loading.enabled` flag, but anyway... :) IMO it's different purpose.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingJob.java
```diff
@@ -0,0 +1,53 @@
...
+        this.intervalSeconds = intervalSeconds;
+
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat("TopicMetadataLoadingJob-%d").build();
+        this.executorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
+    }
+
+    @Override
+    public void run() {
+        topicMetadataLoadingRunner.refreshMetadata();
```
adamdubiel: just one thing here - please add `try {} catch(Exception)`, with proper logging, as if you will throw any exception from scheduled code, no more jobs will be ever scheduled. just a precaution. otherwise jobs will stop running and we will never know it occurred

adamdubiel: to clarify: I can see that you have error handling in `TopicMetadataLoadingRunner` class, but it is better to add this layer explicitly in `run` method, so future changes in `TopicMetadataLoadingRunner` won't introduce any new exceptions that are thrown

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingRunner.java
```diff
@@ -0,0 +1,98 @@
...
+        long start = System.currentTimeMillis();
+        logger.info("Loading topics metadata");
+        List<CachedTopic> topics = topicsCache.getTopics();
+        List<MetadataLoadingResult> allResults = loadMetadataForTopics(topics);
+        logResultInfo(allResults, System.currentTimeMillis() - start);
+    }
+
+    private List<MetadataLoadingResult> loadMetadataForTopics(List<CachedTopic> topics) {
+        try (TopicMetadataLoader loader = new TopicMetadataLoader(brokerMessageProducer, retryCount, retryInterval, threadPoolSize)) {
+            return allComplete(topics.stream().map(loader::loadTopicMetadata).collect(toList())).join();
```
druminski: I have an idea for optimization. Instead of refreshing metadata for all topics maybe we should only refresh metadata for unused topics. Unused topic can be a topic for which `topic.meter.getOneMinuteRate() == 0`. Thanks to this much less requests will be sent to Kafka. What do you say about this?

druminski: Ok, we’ve discussed this and I learned that kafka-producer doesn’t send metadata request if it has required information. Therefore there is no need for further optimization.

PR #1090, opened 2019-08-29
Workload registry/assignment cache refactor

dankraw:
Refactoring few things in order to be able to introduce different type of storage for consumer workload, key points here:
- `ConsumerWorkloadRegistry` is responsible for making changes in the storage
- `SubscriptionAssignmentNotifyingCache` gives read access to the storage (caches data, notifies about changes) 
- `WorkTracker` is used by nodes that read from/write to storage (like leader node when using "selective" algorithm) 

Anyway, this PR does not introduce any new feature, the consumer workload logic is unchanged.


dankraw:
I'm not sure if this PR introduces proper design of the components, probably it will change more than I though. I'm closing it.

PR #1092, opened 2019-09-05
Fix Ports.nextAvailable

jewertow:


jewertow:
I didn't add description because I just wanted trigger CI.
So when I run tests locally they take a lot of time. This change shortened the execution time of local build from 15 min to 4 min.

cristaloleg:
That's interesting, ok. (I heard something similar about firewall and 0.0.0.0, this might be a case)

Issue #1093, opened 2019-09-05
hermes-mock returns response with invalid "Content-Type" header


piekar294:
Currently `hermes-mock` adds `"Content-Type"` header to stubbed response, despite the fact that response body is not set.
https://github.com/allegro/hermes/blob/master/hermes-mock/src/main/java/pl/allegro/tech/hermes/mock/HermesMockHelper.java#L73

I assume that matching `.withHeader("Content-Type", contentType)` should verify request, not a response.

PR #1094, opened 2019-09-05
#1093 hermes-mock matches content-type header

piekar294:
I found that response from `wireMockServer.stubFor` was never verified in test, so I also added status code verification to base tests in `HermesMockAvroTest` and `HermesMockTest`.
This was mostly done to ensure that `WireMock` still matches request after addition of `.withHeader("Content-Type", equalTo(contentType))`.


cristaloleg:
Thank you @piekar294

PR #1096, opened 2019-09-10
resolve #1007 | failed messages metrics

### LINKED ISSUE #1007: Client metrics for failed messages that won't be retried
Currently, there is a client metric counting all failed attempts to publish message ("failure" suffix). However, there is no metric that could be used to monitor for messages that have failed to be published and will not be retried anymore.

Robert-Duda:
resolves #1007 - Client metrics for failed messages that won't be retried 
 - hermes-client.*.retries.exhausted - the number of unsent messages
 - hermes-client.*.retries.success - the number of retried messages with success
 - hermes-client.*.retries.attempts - how many attempts was made before sending message
 - hermes-client.*.retries.count - the number of retried messages


hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -1,30 +1,35 @@
...
+import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
```
cristaloleg: We're trying not to use wildcards in imports 😉

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesResponse.java
```diff
@@ -66,7 +63,7 @@ default String getDebugLog() {
                 .append(getBody());
         getFailureCause().ifPresent(ex ->
                 builder.append(", exception: ")
-                       .append(ex.getMessage())
+                        .append(ex.getMessage())
```
cristaloleg: Unexpected ` ` ? ;D

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -128,4 +139,30 @@ public void close(long pollInterval, long timeout) throws InterruptedException,
...
+            return;
+        }
+
+        HermesMessage message = event.getResult().getHermesMessage();
+        if (shouldPublishMetrics()) {
+            String prefix = MetricsUtils.getMetricsPrefix(message.getTopic());
+            metrics.counter(prefix + ".failure.unsent").inc();
+        }
+
+        LOGGER.error(MessageFormat.format("Failed to send message to topic {0} after {1} attempts",
```
jewertow: Do we really need `MessageFormat`? `slf4j` supports parametrised logging. (https://www.slf4j.org/faq.html#logging_performance)

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesResponse.java
Robert-Duda: intellij formatter

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesResponse.java
```diff
@@ -35,10 +36,6 @@ default boolean isFailure() {
         return Optional.empty();
     }
 
-    default Optional<HermesMessage> getFailedMessage() {
```
jewertow: I'm not sure if it's a good idea to remove this method. It was added a few weeks ago and some users can already use this api. I think we can mark this method as deprecated if necessary, but we should keep it.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -128,4 +142,30 @@ public void close(long pollInterval, long timeout) throws InterruptedException,
         }
     }
 
+    private void handleMaxRetriesExceeded(ExecutionCompletedEvent<HermesResponse> event) {
+        if (event.getResult().isSuccess()) {
```
druminski: Can getResult() be null?

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesResponse.java
Robert-Duda: ok i will leave it as deprecated method, now we always have response message not only in case of error

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
Robert-Duda: it can't be, here we always have the result, each sending operation ends with an appropriate result with an error or success

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesResponse.java
jewertow: We have no information what to use instead. What do you think about comment:
```
/**
 * @deprecated as of Hermes 1.2.4, in favor of {@link #getHermesMessage()}
 */
```
It would be helpful for user.

build.gradle
```diff
@@ -117,6 +120,42 @@ configure(subprojects - project(':integration')) {
             }
         }
     }
+
+    apply plugin: 'maven-publish'
```
druminski: We’ve discussed this and in the next step we will use only `maven-publish` plugin instead of `com.bmuschko.nexus`, so duplicated `pom` configuration will be eliminated.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MicrometerMetricsProvider.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void timerRecord(String name, long duration, TimeUnit unit) {
+        metrics.timer(name).record(duration, unit);
+    }
+
+    @Override
+    public void histogramUpdate(String name, int value) {
+        DistributionSummary.builder(name)
```
druminski: Do we have to build DistributionSummary every time histogramUpdate() is called? Maybe it can be initialize once and then reuse it to record values.

build.gradle
Robert-Duda: unfortunately we have to stay with gradle-nexus-plugin and nexus-staging because we use the mechanism of staging repository deploy not direct deploy

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -18,13 +25,15 @@
 import static pl.allegro.tech.hermes.client.HermesMessage.hermesMessage;
 
 public class HermesClient {
+    private static final Logger LOGGER = LoggerFactory.getLogger(HermesClient.class);
```
dankraw: LOGGER -> logger

hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/DropwizardMetricsProvider.java
```diff
@@ -0,0 +1,40 @@
...
+        metrics.timer(name).update(duration, unit);
+    }
+
+    @Override
+    public void histogramUpdate(String name, int value) {
+        metrics.histogram(name).update(value);
+    }
+
+    private String buildCounterName(String prefix, String name, Map<String, String> tags) {
+        return prefix + "." + name + "." + tags.values().stream().collect(Collectors.joining("."));
```
dankraw: How about `String.join(".", tags.values())` instead of `tags.values().stream().collect(Collectors.joining(".")`

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -128,4 +146,39 @@ public void close(long pollInterval, long timeout) throws InterruptedException,
...
+    private void handleMaxRetriesExceeded(ExecutionCompletedEvent<HermesResponse> event) {
+        if (event.getResult().isSuccess()) {
+            return;
+        }
+
+        HermesMessage message = event.getResult().getHermesMessage();
+        if (shouldPublishMetrics()) {
+            String prefix = MetricsUtils.getMetricsPrefix(message.getTopic());
+            metrics.counterIncrement(prefix + ".retries.exhausted");
+        }
```
dankraw: IMO it would be nice to have this metric-specific logic hidden in a `HermesClientMetrics` (or similar) class, which would have the `MetricsProvider` inside. We'd call `metrics.incrementExhausted(message.getTopic())`, `metrics.incrementRetries(message.getTopic())`, etc. There would be no calls for `shouldPublishMetrics()` here.
Probably some classes could be package protected, like `MetricsUtils`.

Robert-Duda: take a look at the latest version, I change it a little bit

dankraw: it looks like HermesClient could now accept an interface - `MessageDeliveryListener`, by default it could accept a `NoopListener`, and the metered one could accept a `MessageDeliveryMeteredListener`. I would prefer it over client inheritance. What do you think? :)

dankraw: I'm not sure how it'll work with the metrics sender though, anyway, I leave it up to you, please decide whether or not it is worth refactoring. :)

Robert-Duda: I agree with You ;-)  I made some changes and I think that's what it was about

dankraw: Wonderful, better than what I suggested. :)

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClient.java
```diff
@@ -93,18 +108,24 @@ private String createUri(URI uri) {
...
 
     private CompletableFuture<HermesResponse> sendOnce(HermesMessage message) {
-        return sender
-                .send(URI.create(uri + message.getTopic()), message)
-                .exceptionally(e -> HermesResponseBuilder.hermesFailureResponse(e, message));
+        long startTime = System.nanoTime();
+
+        return sender.send(URI.create(uri + message.getTopic()), message)
+                .exceptionally(e -> HermesResponseBuilder.hermesFailureResponse(e, message))
+                .whenComplete((resp, cause) -> messageDeliveryListeners.forEach(l -> l.onSend(resp, System.nanoTime() - startTime)));
```
dankraw: I wonder if it's a problem that every listener will get a little bit different time.

dankraw: Mybe we could make a single variable and pass it to each one.

dankraw: btw, shouldn't we pass the difference of current time and start time (latency)?

PR #1101, opened 2019-09-16
resolve #1009 | disable dynamic reloading of configuration files

### LINKED ISSUE #1009: Disable dynamic reloading of configuration files
Hermes uses archaius library (https://github.com/Netflix/archaius) for configuration management. Current setup allow to dynamically reload properties after changing them in config file. We would like to turn off this feature.

Robert-Duda:


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/ConfigFactory.java
```diff
@@ -41,4 +50,26 @@ private String getProperty(Configs opt) {
...
+            ConfigurationManager.install(new DynamicConfiguration(source, scheduler));
+        }
+        return DynamicPropertyFactory.getInstance();
+    }
+
+    private static AbstractPollingScheduler createDisabledPollingScheduler() {
+        return new AbstractPollingScheduler() {
+            @Override
+            protected void schedule(Runnable pollingRunnable) {
+                LOGGER.info("Periodical polling of a configuration source is turned off!");
```
druminski: Is this going to be logged once or every x period? if the second one then maybe we should change log level to debug.

Robert-Duda: this is only called once when the configuration is created. I gave an entry on info level so that nobody would be surprised that it does not work and there was a trace of it

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/ConfigFactory.java
```diff
@@ -1,13 +1,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.joining;
 
 public class ConfigFactory {
 
-    private final DynamicPropertyFactory propertyFactory = DynamicPropertyFactory.getInstance();
+    private static final Logger LOGGER = LoggerFactory.getLogger(ConfigFactory.class);
```
dankraw: LOGGER -> logger, for please :)
(I know it's static, but it's not a constant, an enum, it's just a regular object that we use)

Robert-Duda: such a habit

dankraw: ;D

PR #1106, opened 2019-09-23
Constraints management in hermes-console

jewertow:
Hermes console allows to create, edit and delete constraints for topics and subscriptions.

Due to the lack of support for roles in hermes-console home view has no redirection button to constraints list. In order to go to constraints list it's necessary to manually redirect to #/constraints.

Screenshot 2019-09-23 at 11 20 27


krzysztofpcy:
Shouldn't constraints be visible in topic/subscription views rather than creating the new one?

jewertow:
@krzysztofpcy we assumed to display topics and subscriptions separately.

hermes-console/static/js/console/constraints/ConstraintsRepository.js
```diff
@@ -0,0 +1,41 @@
...
+        var workloadConstraints = $resource(discovery.resolve('/workload-constraints'), {}, { query: { method: 'GET' } });
+
+        return {
+            getWorkloadConstraints: function () {
+                return workloadConstraints.query({}).$promise
+                    .then(function (response) {
+                        var workloadConstraints = {};
+                        workloadConstraints.topicConstraints = [];
+                        for (topicName in response.topicConstraints) {
+                            if (response.topicConstraints.hasOwnProperty(topicName)) {
```
dankraw: How about iterating with `for (topicName in Object.keys(response.topicConstraints)) {...}`

hermes-console/static/partials/home.html
```diff
@@ -38,4 +37,12 @@
             </a>
         </div>
     </div>
+
+    <div class="row" style="margin-top: 2em;">
+        <div class="col-md-6 col-md-offset-3">
```
dankraw: IMO `col-md-3` would be enough

hermes-console/static/partials/constraintsList.html
```diff
@@ -0,0 +1,59 @@
...
+                <small>{{$index + 1}}</small>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
+                <strong>{{item.subscriptionName}}</strong>
+                <div class="pull-right">
+                    <span class="label label-info">consumers number: {{ item.consumersNumber }}</span>
+                    <span class="glyphicon glyphicon-chevron-right"></span>
+                </div>
+            </a>
+        </div>
+
+        <div ng-show="fetching" class="text-center">
```
dankraw: does it ever show?

jewertow: I added this div by accident when I copied code. I will remove it.

hermes-console/static/js/console/constraints/ConstraintsRepository.js
jewertow: `for (topicName in Object.keys(response.topicConstraints)) {...}` iterates over indexes instead of keys. `for ... of Object.keys(...)` works but not in ES5.
It could be done as `Object.keys(...).forEach(function ...`, but I think it's less readable.

hermes-console/static/partials/home.html
jewertow: Yes, it's enough, but it looks ugly.

hermes-console/static/partials/constraintsList.html
jewertow: I removed it.

hermes-console/static/partials/home.html
dankraw: I meant changing the importance level of the button to lower (like "stats", "docs"...).
On the other hand I think this just doesn't belong here... Like, constraints are part of console, so we should click "console", then there should be a (new screen maybe), that has two buttons:
- groups / topics / subscriptions
- consumer constraints

Or we can redesign the main screen of console. I have to think more about it. 
Let's keep it as but the button should be displayed for admin users only.

hermes-console/static/js/console/constraints/ConstraintsRepository.js
dankraw: is the `hasOwnProperty` check still required?

jewertow: I removed it.

hermes-console/static/partials/home.html
jewertow: I removed this button.

hermes-console/static/partials/constraints.html
```diff
@@ -0,0 +1,51 @@
...
+                    <button class="btn btn-primary" ng-click="addTopicConstraints()">
+                        <span class="glyphicon glyphicon-plus"></span>
+                    </button>
+                </div>
+                <h4>
+                    Topic constraints
+                </h4>
+            </div>
+            <a class="list-group-item" ng-repeat="item in topicConstraints" ng-click="edit(item)">
+                <small>{{$index + 1}}</small>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
```
dankraw: ❤️

PR #1109, opened 2019-09-26
resolve #1047 | Cancel all waiting messages on stopping sender

### LINKED ISSUE #1047: Retrying to deliver messages even though subscription is stopped
Despite subscriptions being stopped/removed consumer senders continue on retrying to deliver messages until their TTL expire. We already ignore success/failure handlers of such messages (#1026), so we could easily add a check if consumer is in `running` state before we schedule the next retry.

krzysztofpcy:
The executor was waiting for all queued messages to be processed, but should just cancel them on shutdown


cristaloleg:
Do we have tests for this?

krzysztofpcy:
> Do we have tests for this?

Tested locally, it's hard to reproduce the case in integration tests.

PR #1110, opened 2019-09-26
Flat binary workload storage

dankraw:
Introducing consumer workload storage implemented as flat tree with binary (SBE) contents. 
Enabled with: `consumer.workload.registry.type=flat-binary`.

Just like with the flat-binary max-rate implementation, only one znode is watched by each consumer.
The leader reads cluster-wide assignments before each calculation and updates data znodes only for consumers that have updated assignments (a subscription assignment was added or removed).

What's more we have now a global consumer nodes registry (`ConsumerNodesRegistry`) - extracted from workload algorithm and now independent, used by max-rate as well. In short, we have a single nodes registry and leader latch. It's always turned on.

Few things about abstractions that were added:
- `ConsumerAssignmentCache` is an interface that provides assignment information for current consumer process
- `ClusterAssignmentCache` is an interface that provides cluster-wide information about assignments, i.e.: it is used by the leader node that performs the calculation

Both have flat-binary implementations. 

When it comes to the legacy `hierarchical` registry type, there's only one class (`HierarchicalConsumerAssignmentCache`) that implements both interfaces, it works as before, every consumer node knows everything about the whole cluster because it watches for changes in the whole workload tree. 

There's `ConsumerAssignmentRegistry` which is an entry point for making changes in zookeeper. It has two implementations as well. The `WorkTracker` class is now removed (it had similar responsibility).


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/ConsumerWorkloadDecoder.java
```diff
@@ -0,0 +1,43 @@
...
+
+    Set<SubscriptionName> decode(byte[] data) {
+        MessageHeaderDecoder header = new MessageHeaderDecoder();
+        AssignmentsDecoder body = new AssignmentsDecoder();
+
+        UnsafeBuffer buffer = new UnsafeBuffer(data);
+        header.wrap(buffer, 0);
+
+        if (header.templateId() != AssignmentsDecoder.TEMPLATE_ID) {
+            throw new IllegalStateException(String.format("AssignmentsDecoder TEMPLATE_ID=%d does not match encoded TEMPLATE_ID=%d",
```
druminski: Do we handle/log this exception?

dankraw: Good question. It would be logged, but throwing this exception is not very safe, as it may break the workload calculation job if unable to decode data for just a single consumer, it's better to just log it and return empty response. The leader will override the corrupted data with new content.
I've changed the behaviour + added check for schema id, meaning schema id and message id should both match. 

Btw I've changed workload xml schema id to 2, as 1 was already used by max rate, and they should be unique. :) It's not very important as we read the messages from different sources/zk paths  but anyway it's now clear.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/ZookeeperOperations.java
```diff
@@ -0,0 +1,69 @@
...
+        this.curator = curator;
+    }
+
+    void writeOrCreatePersistent(String path, byte[] serializedData) throws Exception {
+        try {
+            curator.setData().forPath(path, serializedData);
+        } catch (KeeperException.NoNodeException e) {
+            try {
+                curator.create().creatingParentContainersIfNeeded()
+                        .withMode(CreateMode.PERSISTENT)
```
krzysztofpcy: You can use `orSetData()` method here

dankraw: I'm afraid it's not available in the curator version used. :(

PR #1114, opened 2019-10-03
UX improvements

krzysztofpcy:
* Disable buttons if logging in is required
* Disable buttons if user has no permissions
* Show alert if "read-only" mode was enabled
* Hide buttons that are visible only for administrators


hermes-console/static/partials/group.html
```diff
@@ -7,17 +7,17 @@
...
 
     <h1><small>Group:</small> {{group.groupName}}</h1>
 
-    <div class="row">
+    <div class="row container">
         <div class="list-group">
             <div class="list-group-item">
                 <div class="pull-right">
-                    <button class="btn btn-primary" ng-click="addTopic()"><span class="glyphicon glyphicon-plus"></span></button>
+                    <button class="btn btn-primary" ng-disabled="disabled" ng-click="addTopic()"><span class="glyphicon glyphicon-plus"></span></button>
```
dankraw: I think this button should be enabled when not logged in, we have this feature of `root password` that allows creating topics by anonymous users.

PR #1122, opened 2019-10-08
Change order of role verification

cristaloleg:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/auth/Roles.java
```diff
@@ -1,8 +1,9 @@
 package pl.allegro.tech.hermes.management.api.auth;
 
 public final class Roles {
-    public static final String SUBSCRIPTION_OWNER = "subscriptionOwner";
-    public static final String TOPIC_OWNER = "topicOwner";
-    public static final String ADMIN = "admin";
     public static final String ANY = "any";
+    public static final String ADMIN = "admin";
+    public static final String RESOURCE_WITHOUT_OWNER = "resourceWithoutOwner";
```
dankraw: why adding a new role, isn't admin enough?

cristaloleg: Well, we can allow admins to edit topics and subscriptions, but sometimes this might overload such users(admins) with manual and easy to automate work, so adding new role will fix this. @druminski

dankraw: ok, so we want to automate cleanup up the possible mess. thx :]

dankraw: how about `RESOURCE_WITHOUT_OWNERSHIP_OWNER` ;) as it describes the user role that has rights to specific resource, not the resource itself?

Issue #1058, opened 2019-07-04
Return proper response when Avro lacks __metadata


adamdubiel:
When publishing on Avro topic, which has no `__metadata` field declared in schema, you get `500` with:

```
{
    "code": "INTERNAL_ERROR",
    "message": "Exception caught while creating message"
}
```

and exception is logged:

```
Exception caught while creating message; publishing on topic: XXX; message id: 20ae09bc-5578-4bbf-b061-d9254d123d7f; remote host: XXXXX

pl.allegro.tech.hermes.common.message.wrapper.WrappingException: Could not wrap avro message
	at pl.allegro.tech.hermes.common.message.wrapper.AvroMessageContentWrapper.wrapContent(AvroMessageContentWrapper.java:68)
	at pl.allegro.tech.hermes.common.message.wrapper.MessageContentWrapper.wrapAvro(MessageContentWrapper.java:123)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.createAvroMessage(MessageFactory.java:110)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.create(MessageFactory.java:86)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.create(MessageFactory.java:60)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageCreateHandler.handleRequest(MessageCreateHandler.java:35)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.PreviewHandler.handleRequest(PreviewHandler.java:22)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.finalizeMessageRead(MessageReadHandler.java:169)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.messageRead(MessageReadHandler.java:147)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.lambda$partialMessageRead$2(MessageReadHandler.java:115)
	at io.undertow.io.AsyncReceiverImpl.receivePartialBytes(AsyncReceiverImpl.java:582)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.readMessage(MessageReadHandler.java:90)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageReadHandler.handleRequest(MessageReadHandler.java:61)
[...]

PR #1124, opened 2019-10-10
Return proper response when Avro lacks __metadata

jewertow:
#1058


integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
```diff
@@ -233,6 +233,22 @@ public void shouldReturnServerInternalErrorResponseOnMissingSchemaAtSpecifiedVer
...
+        Topic topic = randomTopic("pl.allegro.test", "Topic")
+                .withContentType(AVRO)
+                .build();
+        operations.buildTopicWithSchema(topicWithSchema(topic, schema.toString()));
+
+        // when
+        Response response = publisher.publish(topic.getQualifiedName(), user.asJson());
+
+        // then
+        assertThat(response).hasStatus(BAD_REQUEST);
```
dankraw: The test should check for existence of a metadata-related user-friendly message.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
```diff
@@ -64,8 +65,11 @@ private MessageMetadata getMetadata(Map<Utf8, Utf8> metadata) {
             GenericRecord genericRecord = bytesToRecord(message, schema);
             genericRecord.put(METADATA_MARKER, metadataMap(id, timestamp, externalMetadata));
             return recordToBytes(genericRecord, schema);
-        } catch (Exception exception) {
-            throw new WrappingException("Could not wrap avro message", exception);
+        } catch (Exception e) {
+            if (e instanceof AvroRuntimeException && e.getMessage().equals("Not a valid schema field: __metadata")) {
+                throw new AvroInvalidMetadataException("Avro schema has invalid field __metadata", e);
+            }
+            throw new WrappingException("Could not wrap avro message", e);
```
dankraw: can't we just stick with `WrappingException` and extend it's message with the cause message?
Like: `throw new WrappingException(String.format("Could not wrap avro message: %s", e.getMessage()), e);`
I think the original message that goes with AvroRuntimeException is just fine, plus we'll avoid equals checking, etc. of the Avro's internal exception.

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
jewertow: Done

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
jewertow: I know it would look better, but I checked usages of `AvroRuntimeException` in the `org.apache.avro:avro:1.8.1` and this library throws this exception 109 times. 
In most cases the exception is thrown due to client error, but this exception is also used to wrap caught `Exception`s and `IOException`s, that's why I decided to handle this one specific case.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentWrapper.java
```diff
@@ -64,8 +65,11 @@ private MessageMetadata getMetadata(Map<Utf8, Utf8> metadata) {
             GenericRecord genericRecord = bytesToRecord(message, schema);
             genericRecord.put(METADATA_MARKER, metadataMap(id, timestamp, externalMetadata));
             return recordToBytes(genericRecord, schema);
-        } catch (Exception exception) {
-            throw new WrappingException("Could not wrap avro message", exception);
+        } catch (Exception e) {
+            if (e instanceof AvroRuntimeException && e.getMessage().equals("Not a valid schema field: __metadata")) {
+                throw new AvroInvalidMetadataException("Avro schema has invalid field __metadata", e);
```
adamdubiel: I would change from `Avro schema has invalid field __metadata` to `Schema does not contain mandatory __metadata field for Hermes internal metadata. Please fix topic schema.` - user will have precise information about the problem and how to fix it.  Other than that: +1

Issue #1081, opened 2019-08-06
Manually create consumer group/commit offsets on subscription creation


dankraw:
We could manually create consumer groups and commit their offsets to specified values (like current topic offsets) and at the same time have kafka's `auto.offset.reset` on consumers set to `earliest` (it would work with `latest` as well).
This way it would be safer to deal with offsets resetting that may happen when messages are removed due to data retention policies, but there are some stale offsets committed. In such cases some messages may be lost if consumer resets offsets to latest.

This can be implemented in consumers (when the subscription is still in pending state) or in management (IMO it's even better option as we control the subscription creation there and have access to all kafka clusters).

jewertow:
I will take care of it.

Issue #1072, opened 2019-07-23
Using TestContainers for integration testing


dankraw:
Hermes uses lots of embedded third party software running for integration testing, including:
kafka, zookeeper, elasticsearch, schema-registry, embedded jms.
We should consider https://www.testcontainers.org/ (or something similar) as an alternative.

What's more we could go even further and build containers of all our modules (frontend, consumers, management) and run them in integration tests in separate JVMs.

druminski:
@mareckmareck will work on this one

druminski:
This task is available again

jewertow:
I was trying to migrate integration tests from embedded third party software to testcontainers, but I give up, because since release of version 1.5.0 `FixedHostPortGenericContainer` will be deprecated. 
Another option is to use `GenericContainer`s, but these containers don't allow to map exposed port to fixed host port and therefore it is not possible to use static configurations in tests. I was trying to do it with `GenericContainer`s and override configuration at runtime, but it was more complex and less clear.
IMHO testcontainers is not a stable solution and probably works well for microservices, but not for large and mature projects like Hermes. I think it would be better to add docker-compose file and execute at startup `Runtime.getRuntime().exec("docker-compose up")` even if it is additional dependency.

druminski:
@jewertow thanks for the try, your feedback about this issue is very valuable for us and due to it we will not migrate Hermes integration tests to TestContainers. I’m closing the issue.

bsideup:
Uff, sorry to hear that you decided not to use Testcontainers!

> IMHO testcontainers is not a stable solution and probably works well for microservices, but not for large and mature projects like Hermes

Maybe, yeah... It is only used by a few small and young projects like Spring Boot/Framework/Data/Cloud, Quarkus, Micronaut, Akka's Alpakka, Apache Pulsar, JOOQ, Vert.x, Infinispan, Spinnaker, Apache Camel... 😅 

> but I give up, because since release of version 1.5.0 FixedHostPortGenericContainer will be deprecated.
> Another option is to use GenericContainers, but these containers don't allow to map exposed port to fixed host port and therefore it is not possible to use static configurations in tests

There is a reason why we decided to deprecate fixed ports. Consider reading the following article:
https://bsideup.github.io/posts/testcontainers_fixed_ports/

But I respect your decision! 👍

jewertow:
[...]

PR #1127, opened 2019-10-21
Manually create consumer group/commit offsets on subscription creation.

jewertow:
#1081 
`KafkaConsumerGroupsManager` creates `KafkaConsumer`, retrieves topic partitions and current offset for each partition sequentially, commits received offset and closes `KafkaConsumer`. 
To enable this feature it is necessary to add following configuration to application.yml:
```
management:
  consumer-groups:
    create-manually: true
```


hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -0,0 +1,111 @@
...
+    private final String clusterName;
+    private final ExecutorService executor;
+
+    public KafkaConsumerGroupManager(BrokerStorage brokerStorage,
+                                     KafkaNamesMapper kafkaNamesMapper,
+                                     Integer partitions,
+                                     String clusterName) {
+        this.brokerStorage = brokerStorage;
+        this.kafkaNamesMapper = kafkaNamesMapper;
+        this.partitions = partitions;
```
dankraw: Actually, this is the current number of partitions we want newly created topics to have.
Old topics that were created some time ago, or are some manually created, may have different number of partitions. It would be best to get the number from the exact kafka topic.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -0,0 +1,111 @@
...
+        this.clusterName = clusterName;
+        this.executor = Executors.newFixedThreadPool(
+                partitions, new ThreadFactoryBuilder().setNameFormat("kafka-consumer-group-manager-%d").build()
```
dankraw: I think a single thread could be enough, we could schedule the job in background and iterate over the partitions sequentially.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -0,0 +1,111 @@
...
+            this.topicPartition = new TopicPartition(subscription.getQualifiedTopicName(), partition);
+        }
+
+        @Override
+        public void run() {
+            logger.info("Creating consumer group for subscription {} on partition {}", subscription.getQualifiedName(), partition);
+
+            int leaderId = brokerStorage.readLeaderForPartition(new TopicAndPartition(topicPartition));
+            ConsumerGroupId groupId = kafkaNamesMapper.toConsumerGroupId(subscription.getQualifiedName());
+            KafkaConsumer<byte[], byte[]> kafkaConsumer = createKafkaConsumer(groupId, leaderId);
```
dankraw: do we have to create a consumer for each partition and connect to a partition leader?
I believe we could operate just as with consumer's module, have a single consumer that connects to the kafka cluster (not by selecting the leader at all), assign all partitions to it, and commit offsets we want to.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -0,0 +1,111 @@
...
+
+            kafkaConsumer.assign(ImmutableSet.of(topicPartition));
+            long offset = kafkaConsumer.position(topicPartition);
```
dankraw: Aren't we reading the current offset of the consumer here? 
We want to read the current offset of kafka topic itself and move the offset to newly created consumer group.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -0,0 +1,111 @@
...
+            return new KafkaConsumer<>(properties(groupId, bootstrapServers));
+        }
+
+        private Properties properties(ConsumerGroupId groupId, String bootstrapServers) {
+            Properties props = new Properties();
+            props.put(BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+            props.put(GROUP_ID_CONFIG, groupId.asString());
+            props.put("key.deserializer", "org.apache.kafka.common.serialization.ByteArrayDeserializer");
+            props.put("value.deserializer", "org.apache.kafka.common.serialization.ByteArrayDeserializer");
+            return props;
```
dankraw: you should include `ENABLE_AUTO_COMMIT_CONFIG`=false

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Properties;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.stream.IntStream;
+
+import static org.apache.kafka.clients.consumer.ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.GROUP_ID_CONFIG;
+
+public class KafkaConsumerGroupManager implements ConsumerGroupManager {
```
dankraw: This class needs some proper tests. :)

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
jewertow: Thanks, I fixed it.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
jewertow: Done.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
jewertow: Yes, it wasn't necessary. I fixed it.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
jewertow: Done

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
jewertow: You're right. We are reading current offset of the consumer, but the current offset of the consumer in this situation is current offset of kafka topic, because by default `auto.offset.reset` is set to `latest`.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/BrokersClusterService.java
```diff
@@ -97,6 +100,10 @@ public boolean allSubscriptionsHaveConsumersAssigned(Topic topic, List<Subscript
         }
     }
 
+    public void createConsumerGroup(Subscription subscription) {
+        consumerGroupManager.createConsumerGroup(subscription);
```
dankraw: I think this feature should be configurable (enable/disable), because there may be users that used the "earliest" offset reset policy intentionally. For them, this feature should be disabled, so consumers will behave as before.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
jewertow: I added one test based on testcontainers.

jewertow: Done

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/BrokersClusterService.java
jewertow: Done

jewertow:
PR still needs work, because this feature doesn't work with mirror workload algorithm.

jewertow:
For now I don't see any simple and non-invasive method of implementing this feature to make it work with mirror configuration.
Since this feature is optional we could disable it by default and describe in documentation that it will not work properly when hermes-consumers has enabled mirror workload algorithm.
Even if the user would run hermes-management with manually creating subscriptions and hermes-consumers with mirror algorithm the whole Hermes will work properly and only hermes-mamangement will log errors that subscriptions have already been created.
What do you think @dankraw @cristaloleg @druminski ?

hermes-management/src/test/groovy/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManagerSpec.groovy
```diff
@@ -0,0 +1,131 @@
+package pl.allegro.tech.hermes.management.infrastructure.kafka.service
+
+import org.apache.kafka.clients.admin.AdminClient
+import org.apache.kafka.clients.admin.NewTopic
+import org.apache.kafka.clients.producer.KafkaProducer
+import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.common.TopicPartition
+import org.junit.Rule
+import org.springframework.boot.test.rule.OutputCapture
+import org.testcontainers.containers.KafkaContainer
```
druminski: Wow, nice that you decided to use here Kafka container. In the future we would like to rewrite all hermes integration tests to use containers as well (#1072 ).

druminski:
> For now I don't see any simple and non-invasive method of implementing this feature to make it work with mirror configuration.
> Since this feature is optional we could disable it by default and describe in documentation that it will not work properly when hermes-consumers has enabled mirror workload algorithm.
> Even if the user would run hermes-management with manually creating subscriptions and hermes-consumers with mirror algorithm the whole Hermes will work properly and only hermes-mamangement will log errors that subscriptions have already been created.
> What do you think @dankraw @cristaloleg @druminski ?

@jewertow don’t worry about mirror algorithm. It’s very old and we don’t use it. The default one is `selective` and in the near future we will probably remove `mirror` one.

druminski:
I’ve applied minor refactor. From now on manual consumer groups are created by default. This feature can be disabled by setting `false` for `subscription.createConsumerGroupManuallyEnabled` flag. Also changed default value for `kafka.consumer.auto.offset.reset` to `earliest`.

PR #1128, opened 2019-10-22
Fix underscore in subscription name

cristaloleg:
PR fixes subscription metrics url handing via additional underscore replacement.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
@@ -25,6 +25,7 @@ import static pl.allegro.tech.hermes.metrics.PathContext.pathContext;
 public class HermesMetrics {
 
     public static final String REPLACEMENT_CHAR = "_";
+    public static final String TMP_REPLACEMENT_CHAR = "-x-";
 
     private final MetricRegistry metricRegistry;
     private final PathsCompiler pathCompiler;
@@ -38,7 +39,15 @@ public class HermesMetrics {
     }
 
     public static String escapeDots(String value) {
-        return value.replaceAll("\\.", REPLACEMENT_CHAR);
+        return value
+                .replaceAll("_", TMP_REPLACEMENT_CHAR)
+                .replaceAll("\\.", REPLACEMENT_CHAR);
+    }
+
+    public static String unescapeDots(String value) {
+        return value
+                .replaceAll(HermesMetrics.REPLACEMENT_CHAR, "\\.")
+                .replaceAll(HermesMetrics.TMP_REPLACEMENT_CHAR, "_");
     }
 
     public Timer timer(String metric) {
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
@@ -25,7 +25,7 @@ import static pl.allegro.tech.hermes.metrics.PathContext.pathContext;
 public class HermesMetrics {
 
     public static final String REPLACEMENT_CHAR = "_";
-    public static final String TMP_REPLACEMENT_CHAR = "-x-";
+    public static final String UNDERSCORE_REPLACEMENT_CHAR = "€";
 
     private final MetricRegistry metricRegistry;
     private final PathsCompiler pathCompiler;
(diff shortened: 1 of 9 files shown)


hermes-common/src/test/groovy/pl/allegro/tech/hermes/common/metric/counter/zookeeper/CounterMatcherTest.groovy
```diff
@@ -34,6 +34,22 @@ class CounterMatcherTest extends Specification {
         subscriptionName == "subscription"
     }
 
+    def "should match fancy subscription delivered"() {
```
adamdubiel: what does "fancy" mean? would be great to have precise name for this test

cristaloleg: ok

hermes-common/src/test/groovy/pl/allegro/tech/hermes/common/metric/counter/zookeeper/CounterMatcherTest.groovy
```diff
@@ -34,6 +34,22 @@ class CounterMatcherTest extends Specification {
         subscriptionName == "subscription"
     }
 
+    def "should match subscription delivered with underscores in name"() {
+        given:
+            def counterName = "delivered.pl_allegro_offercore.topic.auction_offer-ended-only"
```
dankraw: can we stick to the given/when/then code style without such indentations, just to be consistent (even in this single file)?

cristaloleg: Sorry, my bad

dankraw: I mean, it's fine, some people will say it's more readable (I personally prefer the no-indentation way as it's presented in spock documentation's examples http://spockframework.org/spock/docs/1.0/spock_primer.html), it simply would be nice to have only a single style across our code. If we change our minds, let's change it globally. :)

cristaloleg: Ah, ok, got it. Yeah, in the team we can discuss preferable style and change it gradually, but currently it will be easier to unindent my test case.

dankraw: Great, thanks!

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/HermesMetrics.java
```diff
@@ -25,6 +25,7 @@
 public class HermesMetrics {
 
     public static final String REPLACEMENT_CHAR = "_";
+    public static final String TMP_REPLACEMENT_CHAR = "-x-";
```
cristaloleg: If someone have a better temporary replacement char - feel free to suggest

dankraw: `-x-` is a chars combination that passes subscription name regexp: `[a-zA-Z0-9_.-]+`, shouldn't we use something that can't be provided by users, i.e. use special chars like `|`, `=`... ?

adamdubiel: does this mean that we are going to have names like `my-subscripti-x-on` in graphite? I would rather not, as this introduces some magic that our customers need to be aware of if they want to see metrics from they subscriptions

cristaloleg: Sorry, have totally missed you comment, Daniel. I've tried to use symbol `€` but somewhere in internals (Graphite url encoding?) it breaks the behaviour. So decided to ask for an another pair of eyes. 
Ok, will rework.

adamdubiel: *bump* are we going to have metrics published under names like `my€subscription` in Graphite? (or any other char), what does `tmp` mean in this context?

cristaloleg: `tmp` for `temporary`, probably `UNDERSCORE_REPLACEMENT_CHAR` should be better. Yes, metrics will be under another name. Objections?

adamdubiel: Yes, strong objections :) This means that all clients need to be aware of this transformation. Also we break existing behaviour - what if someone has alert triggered by this metric? We will break it for him.

Why do we have to make this change in Hermes?

cristaloleg:
Closing as not a safe change.

PR #1139, opened 2019-10-24
Resolves #1126 | Redirects are disabled by default

### LINKED ISSUE #1126: Hermes redirects request as GET for 301 response code
When Hermes sends a POST message to a subscriber and the subscriber returns 301 then Hermes sends second request but with GET method.

In the first step we have to confirm this hypothesis. If it’s true then we should have a flag on subscription which allows on redirects. By default it should be set to false. https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.2

druminski:
Added global flag for enabling redirects. By default they are disabled. Solves #1126

### DIFF
--- a/hermes-test-helper/src/main/java/pl/allegro/tech/hermes/test/helper/endpoint/RemoteServiceEndpoint.java
+++ b/hermes-test-helper/src/main/java/pl/allegro/tech/hermes/test/helper/endpoint/RemoteServiceEndpoint.java
@@ -80,7 +80,7 @@ public class RemoteServiceEndpoint {
         messages.forEach(m -> listener
             .register(
                 post(urlEqualTo(path))
-                .willReturn(aResponse().withStatus(returnedStatusCode).withFixedDelay(delay).withHeader("Location", "http://localhost:1234"))));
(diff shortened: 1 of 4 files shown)


jewertow:
I tested it, works very well.

hermes-test-helper/src/main/java/pl/allegro/tech/hermes/test/helper/endpoint/RemoteServiceEndpoint.java
```diff
@@ -79,7 +80,19 @@ public void expectMessages(List<String> messages) {
         messages.forEach(m -> listener
             .register(
                 post(urlEqualTo(path))
-                .willReturn(aResponse().withStatus(returnedStatusCode).withFixedDelay(delay))));
+                .willReturn(aResponse().withStatus(returnedStatusCode).withFixedDelay(delay).withHeader("Location", "http://localhost:1234"))));
```
dankraw: is the Location header required here?

druminski: Actually no, it’s not required, removed it.

PR #1141, opened 2019-10-24
resolves #1116 | Add sorting by name and search for constraints UI

### LINKED ISSUE #1116: Consumer constraints UI improvements
Things that can be improved in the UI of constraints management:
* alphabetical order of entries displayed
* link to the exact topic/subscription on each constraint
* filter/search box

pwolaq:
- [x] alphabetical order of entries displayed
- [ ] link to the exact topic/subscription on each constraint
- [x] filter/search box

Adding link to the exact topic/subscription is impossible as the whole row acts as a trigger for modal. For now the only solution is to wrap topic/subscription name with a link, but I think it would be not intuitive.

If you still want this behavior I can update PR.

### DIFF
--- a/hermes-console/static/partials/constraints.html
+++ b/hermes-console/static/partials/constraints.html
@@ -5,7 +5,11 @@
     <div class="row">
         <div class="list-group">
             <div class="list-group-item">
-                <div class="pull-right">
+                <div class="form-inline pull-right">
+                    <div class="input-group">
+                        <span class="input-group-addon"><span class="glyphicon glyphicon-search"></span></span>
+                        <input class="form-control" ng-model="search.topicName" placeholder="search...">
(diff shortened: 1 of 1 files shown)


dankraw:
In my opinion, link to the exact topic/subscription is a very useful feature and is worth adding anyway. :)

Maybe we could change the current behaviour, so the whole row is not clickable and add buttons on the right, like: 
`edit` or [an icon](https://fontawesome.com/icons/edit) (would open the modal), 
`>` or [an icon](https://fontawesome.com/icons/external-link-alt) (would redirect to target topic/subscription). 

What do you think @pwolaq?

druminski:
@pwolaq thanks for the PR. I will merge it in the current form and verify whether it is sufficient.

PR #1146, opened 2019-10-31
Subscription ids cache removed expire after specified time

### LINKED ISSUE #1145: Removing subscription may not stop consumer process
When using `flat-binary` workload registry type, removing the subscription may not stop the consumer process itself. This is due to the cleaning of subscription ids cache from removed subscriptions. When decoding the binary payload of workload configuration for a specific node, the leader node making computations will filter out ids that could not be translated to subscription names. Because of this it may not observe any changes after balancing the workload and may not write any changes for a specific consumer.

dankraw:
resolves #1145 

Ids of removed subscriptions should be still available for some time after the subscriptions are removed, due to the way selective workload algorithm works (it decodes them and should not filter them out in order to clean the mess properly). I'm introducing cache for removed subscriptions that is cleared after some specified time of inactivity (should be bigger than configured intervals for workload/max-rate computations).

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCache.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCache.java
@@ -1,5 +1,11 @@
 package pl.allegro.tech.hermes.consumers.subscription.id;
 
+import com.google.common.base.Ticker;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/subscription/id/NotificationAwareSubscriptionIdsCache.java
```diff
@@ -40,12 +73,14 @@ private void putSubscriptionId(SubscriptionName name) {
 
     @Override
     public Optional<SubscriptionId> getSubscriptionId(SubscriptionName subscriptionName) {
-        return Optional.ofNullable(nameToIdMap.get(subscriptionName));
+        return Optional.ofNullable(Optional.ofNullable(nameToIdMap.get(subscriptionName))
```
cristaloleg: Dunno, but maybe a static import for `ofNullable` isn't a bad idea at all, wdyt?

PR #1147, opened 2019-11-07
add support for IE 11

kuaikuai:
When using console in IE, because js in IE does not support endswith() , hermes-console does not load properly.

### DIFF
--- a/hermes-console/static/js/bootstrap.js
+++ b/hermes-console/static/js/bootstrap.js
@@ -1,8 +1,3 @@
-if (typeof String.prototype.endsWith != 'function') {  
-    String.prototype.endsWith = function(suffix) {  
-        return this.indexOf(suffix, this.length - suffix.length) !== -1;  
-    };  
-}
 deferredBootstrapper.bootstrap({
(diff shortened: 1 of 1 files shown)


hermes-console/static/js/bootstrap.js
```diff
@@ -1,3 +1,8 @@
+if (typeof String.prototype.endsWith != 'function') {  
+    String.prototype.endsWith = function(suffix) {  
+        return this.indexOf(suffix, this.length - suffix.length) !== -1;  
+    };  
+}
```
dankraw: Thanks for the PR.
I can see that this file has the only occurrence of `endsWith`, could you extract it to a separate method instead of extending the String prototype functionality (`endsWith(url, '/')` instead of `url.endsWith('/')`)? 
In the future we'll probably use Babel in order to get rid of such problems.

Or maybe it would be even better to extract the whole line as `var url = removeTrailingSlash(url)` or something, as this would make the code more explicit. :)

dankraw:
Thanks @kuaikuai !

PR #1153, opened 2019-11-28
Fix Hermes mock predicate

### LINKED ISSUE #1152: HermesMockExpect#avroMessagesOnTopic with a predicate does not work (when not all messages matches the predicate)
https://github.com/allegro/hermes/blob/0a512c8b0fe86cfca8e50e347f4b5ce2c8e37686/hermes-mock/src/main/java/pl/allegro/tech/hermes/mock/HermesMockExpect.java#L52
does not work when not all messages matches the predicate. 

An example:
```groovy
def "should get some filtered messages as avro"() {
        given:
        def topicName = "get-some-filtered-avro-topic"
        hermes.define().avroTopic(topicName)

        def messages = (1..5).collect { new TestMessage("key-" + it, "value-" + it) }
        def filter = { TestMessage m -> m.key.startsWith("key-1") || m.key.startsWith("key-3") }

        when:
        messages.each { publish(topicName, it) }

        then:
        hermes.expect().avroMessagesOnTopic(topicName, 2, schema, TestMessage, filter)
    }
```
I expect it to pass, but I get a failure instead:
`Expected exactly 2 requests matching the following pattern but received 5`

cristaloleg:
The problem lies in `HermesMockExpect.assertMessages`.

`expectMessages` and `expectSpecificMessages` are orthogonal, 'cause they assert different behaviour, total number of events vs specific events respectively.

Such behaviour was caused by a too generic predicate in tests.

Fixes #1152

### DIFF
--- a/hermes-mock/src/main/java/pl/allegro/tech/hermes/mock/HermesMockExpect.java
+++ b/hermes-mock/src/main/java/pl/allegro/tech/hermes/mock/HermesMockExpect.java
@@ -54,7 +54,6 @@ public class HermesMockExpect {
     }
 
     private <T> void assertMessages(String topicName, int count, Supplier<List<T>> messages) {
-        expectMessages(topicName, count);
         expectSpecificMessages(count, messages.get());
     }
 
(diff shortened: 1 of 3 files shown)


platan:
@cristaloleg LTGM. Thanks!

PR #1171, opened 2020-02-17
Develop branch improvements

druminski:
This PR introduces changes to few different functionalities on a branch develop. Therefore I suggest to review it by commits. 

Improvements:

* used Kafka topic name mapper in KafkaConsumerGroupManager. Also changed place where this manager is called. Now it is called before subscription creation. Thanks to this hermes-consumer creates subscription after kafka consumer group is created in hermes-management.

* Extended Jersey configuration in hermes-management. We need this change to expose spring boot actuator endpoints.

* added popovers displayed near disabled buttons when user isn’t signed in

* hermes-console footer can be configured via hermes-management config

### DIFF
--- a/hermes-console/static/partials/group.html
+++ b/hermes-console/static/partials/group.html
@@ -17,6 +17,7 @@
         <div class="list-group">
             <div class="list-group-item">
                 <div class="pull-right">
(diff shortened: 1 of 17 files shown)


druminski:
@jewertow I also test it in our environment, so you don't have to, unless you really want to

PR #1173, opened 2020-02-18
Metadata headers in messages

mareckmareck:
This PR allows sending some of the metadata (message id, message timestamp and schema version) directly as Kafka Headers. Thanks to this change it will be possible to get rid of magic byte mechanism. It also prevents a few rare cases in which consumers try to deserialize Avro-serialized message with different schema than it was serialized in, resulting in various crashes.


hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageContentUnwrapperResult.java
```diff
@@ -0,0 +1,33 @@
+package pl.allegro.tech.hermes.common.message.wrapper;
+
+class AvroMessageContentUnwrapperResult {
```
mareckmareck: I'm considering sacrificing this class (together with some readability) for the sake of performance and returning simple null/not null result.

mareckmareck: Yeah, that was exactly what I was thinking about in that comment

cristaloleg: I did a review, but forgot to click submit PR, sorry)

hermes-common/src/main/java/pl/allegro/tech/hermes/common/message/wrapper/AvroMessageSchemaVersionAwareContentWrapper.java
```diff
@@ -0,0 +1,70 @@
...
+            logger.warn("Could not deserialize schema version aware payload for topic [{}] - falling back", topic.getQualifiedName(), ex);
+
+            deserializationErrorsForSchemaVersionAwarePayload.inc();
+
+            return AvroMessageContentUnwrapperResult.failure();
+        }
+    }
+
+    @Override
+    public boolean isApplicable(byte[] data, Topic topic, Integer schemaVersion) {
```
cristaloleg: unused `schemaVersion` ?

mareckmareck: Intentional, this way of unwrapping does not use schemaVersion (because it takes it from schema anyway). I could document it somehow here. Do you have some different solution in mind for that?

PR #1177, opened 2020-03-05
HttpClients can have additional headers set

mareckmareck:
At the moment there is no way to add additional headers to consumer messages to subscribers without hacking and slashing.

This PR allows specifying additional header providers to be used by consumer during message request creation. 

Other than that, each provider can now use raw (not-yet-resolved-by-address-resolvers) address to be used when resolving such additional headers.

That can be useful in a scenario when one want to send headers like `Host` or other that rely on the information stored in the address.

### DIFF
--- a/hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/consumer/sender/http/headers/HermesHeadersProviderTest.groovy
+++ b/hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/consumer/sender/http/headers/HermesHeadersProviderTest.groovy
@@ -1,23 +1,26 @@
 package pl.allegro.tech.hermes.consumers.consumer.sender.http.headers
 
+import com.google.common.collect.ImmutableSet
 import pl.allegro.tech.hermes.consumers.consumer.Message
 import spock.lang.Specification
-import spock.lang.Unroll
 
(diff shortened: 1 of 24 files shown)


hermes-consumers/src/test/groovy/pl/allegro/tech/hermes/consumers/consumer/sender/http/headers/HermesHeadersProviderTest.groovy
```diff
@@ -65,19 +71,51 @@ class HermesHeadersProviderTest extends Specification {
...
         when:
-        Map<String, String> headers = hermesHeadersProvider.getHeaders(message()).asMap()
+        Map<String, String> headers = hermesHeadersProvider.getHeaders(message(), rawAddress()).asMap()
 
         then:
         headers.get("k") == "v"
     }
 
+    def "should produce headers appending headers with additional providers"() {
+        def additionalProviderOne = new HttpHeadersProvider() {
```
dankraw: the "given:" section should start here IMO :]

mareckmareck: True ;)

Issue #1181, opened 2020-03-16
hermes-console/package.sh fails with no node in the path


trombka:
The latest commit to `package.sh`, where PATH variable modification was substituted with a full path to npm command, causes false-successful builds when there is no node command executable via PATH.
```
> Task :hermes-management:buildHermesConsole
Packaging Hermes Console
Running NPM and bower
/usr/bin/env: 'node': No such file or directory
/usr/bin/env: 'node': No such file or directory
Creating directory: dist/static
```
The PATH modification should be restored.

trombka:
PR #1182

PR #1182, opened 2020-03-16
restoring node command in th path

trombka:
#1181  fix


druminski:
@trombka Thanks for the fix :)

PR #1183, opened 2020-03-20
Retry on 429

### LINKED ISSUE #1169: Retry-After header also supported for 429 http status code
Hermes provides [back pressure](https://hermes-pubsub.readthedocs.io/en/latest/user/subscribing/#back-pressure) mechanism relaying only on 503 http status header. We would like to extend this mechanism also on 429 http status code.

429 code combined with `Retry-After` header  should not influence negatively on sending speed to a subscriber as well.

cristaloleg:
Fixes #1169


druminski:
@cristaloleg great, can you also update hermes documentation about this change?

cristaloleg:
yep

PR #1186, opened 2020-03-24
Added metrics for http client connection pool

druminski:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsWorkloadReporter.java
```diff
@@ -53,4 +89,45 @@ int getHttp2QueueSize() {
...
+    private final Function<HttpClient, Integer> getHttp2ConnectionsCount = http2Client ->
+            getHttp2ConnectionPool(http2Client)
+                    .map(MultiplexConnectionPool::getConnectionCount)
+                    .mapToInt(i -> i)
+                    .sum();
+
+    private final Function<HttpClient, Integer> getHttp2PendingConnectionsCount = http2Client ->
+            getHttp2ConnectionPool(http2Client)
+                    .map(MultiplexConnectionPool::getPendingCount)
+                    .mapToInt(i -> i)
```
faderskd: You can use `Function.identity()` to be more concise :)

druminski: Unfortunately I can’t as mapToInt() requires ToIntFunction which is different than Function.

PR #1188, opened 2020-03-24
Max requests queued per destination config param for http clients

mareckmareck:
Http1/2 clients can have max requests queued per destination configured apart from consumer inflight size.

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
@@ -12,14 +12,15 @@ import javax.inject.Inject;
 import java.util.concurrent.ExecutorService;
 
 import static pl.allegro.tech.hermes.common.config.Configs.CONSUMER_HTTP2_CLIENT_IDLE_TIMEOUT;
+import static pl.allegro.tech.hermes.common.config.Configs.CONSUMER_HTTP2_CLIENT_MAX_REQUESTS_QUEUED_PER_DESTINATION;
 import static pl.allegro.tech.hermes.common.config.Configs.CONSUMER_HTTP2_CLIENT_THREAD_POOL_MONITORING;
 import static pl.allegro.tech.hermes.common.config.Configs.CONSUMER_HTTP2_CLIENT_THREAD_POOL_SIZE;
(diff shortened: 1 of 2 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
```diff
@@ -44,7 +45,7 @@ public HttpClient createClientForHttp1(String name) {
 
         HttpClient client = new HttpClient(createSslContextFactory());
         client.setMaxConnectionsPerDestination(configFactory.getIntProperty(CONSUMER_HTTP_CLIENT_MAX_CONNECTIONS_PER_DESTINATION));
-        client.setMaxRequestsQueuedPerDestination(configFactory.getIntProperty(CONSUMER_INFLIGHT_SIZE));
+        client.setMaxRequestsQueuedPerDestination(configFactory.getIntProperty(CONSUMER_HTTP_CLIENT_MAX_REQUESTS_QUEUED_PER_DESTINATION));
```
jewertow: From the javadoc: `This parameter controls how many requests can be queued before starting to reject them.`
Do we want to reject messages that cannot be enqueued? How will this affect the `SerialConsumer`?

druminski: If you don’t have any proxy between hermes-consumers and subscribers then reasonable is to have the same value for CONSUMER_INFLIGHT_SIZE and CONSUMER_HTTP_CLIENT_MAX_REQUESTS_QUEUED_PER_DESTINATION. Otherwise you will probably want to increase CONSUMER_HTTP_CLIENT_MAX_REQUESTS_QUEUED_PER_DESTINATION as for all subscriptions you will have the same address, for example localhost for local proxy.

Answering to your questions:
1. In the first place we don’t want to lead to the situation where http requests queue is full. This is the reason why we would like to increase queue for proxy.
2. If somehow http requests queue will be full then we want to retry sending requests to it and that's how it works now. 
3. Increased value for CONSUMER_HTTP_CLIENT_MAX_REQUESTS_QUEUED_PER_DESTINATION should not affect on SerialConsumer
4. You have added metrics for http requests queue size. Thanks to this we can add a check and detect problematic situations for it.

Issue #1015, opened 2019-04-19
Schema `__metadata` field validation of `avro.java.string` attribute


dankraw:
Users tend to add `avro.java.string` attribute to String fields in Avro schemas in order to generate Java classes with proper `String` type instead of Avro's `Utf8`. If the attribute is added to all fields - and by mistake - to `__metadata` field as well (which is a required field), messages produced with this version of schema will not be deserialised by Hermes consumers at all - we get a ` java.lang.ClassCastException: java.lang.String cannot be cast to org.apache.avro.util.Utf8` (even though frontends will accept the incoming messages, wrap them adding custom metadata and publish to Kafka without any problem).

1. We could validate the field in Hermes management if it contains the attribute and disallow such Schema extension for `__metadata` field.
2. The other solution is fix consumers behaviour, and allow both `Map<String, String>` and `Map<Utf8, Utf8>` as `__metadata` target type.

The troublesome field looks as follows:

```js
{
  "name": "__metadata",
  "type": [
    "null",
    {
      "type": "map",
      "values": "string",
      "avro.java.string": "String"
    }
  ],
  "doc": "Field internally used by Hermes to propagate metadata.",
  "default": null
}
```

druminski:
The second one solution is better and it should be easy to implement. As a hint we have integration tests for publishing avro messages.

jewertow:
I will take care of it.

jewertow:
Resolved in #1192

PR #1192, opened 2020-03-30
Management: Schema __metadata field validation

jewertow:
References #1015 

## Changes
Hermes Management does not allow to save topic with an invalid Avro schemas:
- schema without field `__metadata`
- schema with field `__metadata` that declares types other than hermes specifies in documentation, for example `java.avro.string`

### DIFF
--- a/integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
+++ b/integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
@@ -237,28 +237,6 @@ public class PublishingAvroTest extends IntegrationTest {
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
```diff
@@ -233,25 +236,6 @@ public void shouldReturnServerInternalErrorResponseOnMissingSchemaAtSpecifiedVer
         assertThat(response.getStatus()).isEqualTo(INTERNAL_SERVER_ERROR.getStatusCode());
     }
 
-    @Test
-    public void shouldReturnBadRequestResponseOnMissingMetadataFieldInSchema() {
```
jewertow: For now it is impossible scenario.

PR #1194, opened 2020-03-31
Fix typo in docs

benzwreck:
Hi there!
I've found and fixed some typo but cannot see a ```develop``` branch as mentioned [here](https://github.com/allegro/hermes/blob/master/CONTRIBUTING.md) so ```master``` seems to be good one for such a small change :)


cristaloleg:
Yeah, `develop` branch is legacy which we've removed year ago or so. We will update a contribution doc, thank you 😉

PR #1196, opened 2020-04-13
Hermes supports JRE truststore

jewertow:
## Background context
Hermes documentation says that hermes-consumers use JRE trust store to verify the ssl certificates, but in fact hermes-consumers use provided key store and trust store. It should work also with JRE trust store (`$JAVA_HOME/jre/lib/security/cacerts`).

## Changes
- `consumer.ssl.enabled` is enabled by default

Frontend and Consumers allow to configure source of key store and trust store.
There are new configuration properties:
- `{consumers,frontend}.ssl.{key,trust}store.source` - can be set to `jre` (default) or `provided`
In case of `provided` Hermes loads provided key/trust store from specified location (`{consumer,frontend}.ssl{key,trust}store.location`) 
In case of `jre` Hermes loads trust store from `$JAVA_HOME/jre/lib/security/jssecacerts` or `$JAVA_HOME/jre/lib/security/cacerts`.


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -122,18 +122,30 @@
     FRONTEND_SSL_PORT("frontend.ssl.port", 8443),
     FRONTEND_SSL_CLIENT_AUTH_MODE("frontend.ssl.client.auth.mode", "not_requested"),
     FRONTEND_SSL_PROTOCOL("frontend.ssl.protocol", "TLS"),
+
+    FRONTEND_SSL_KEYSTORE_DEFAULT_JVM("frontend.ssl.keystore.default.jvm", true),
```
jewertow, line 126: I am not sure whether this name is comprehensible. 
Maybe `frontend.ssl.keystore.jvm.based` would be more appropriate...?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -122,18 +122,30 @@
...
+
+    FRONTEND_SSL_TRUSTSTORE_DEFAULT_JVM("frontend.ssl.truststore.default.jvm", true),
+    FRONTEND_SSL_TRUSTSTORE_PROVIDED("frontend.ssl.truststore.provided", false),
     FRONTEND_SSL_TRUSTSTORE_LOCATION("frontend.ssl.truststore.location", "classpath:server.truststore"),
     FRONTEND_SSL_TRUSTSTORE_PASSWORD("frontend.ssl.truststore.password", "password"),
     FRONTEND_SSL_TRUSTSTORE_FORMAT("frontend.ssl.truststore.format", "JKS"),
 
-    CONSUMER_SSL_ENABLED("consumer.ssl.enabled", false),
-    CONSUMER_SSL_PROTOCOL("consumer.ssl.protocol", "TLS"),
+    CONSUMER_SSL_ENABLED("consumer.ssl.enabled", true),
```
jewertow, line 138: Since property `consumer.http2.enabled` is enabled by default it should also be enabled.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
```diff
@@ -122,18 +122,30 @@
...
+    CONSUMER_SSL_PROTOCOL("consumer.ssl.protocol", "TLSv1.2"),
```
jewertow, line 139: Just a mistake

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
```diff
@@ -43,7 +42,9 @@ public HttpClient createClientForHttp1(String name) {
                 configFactory.getIntProperty(CONSUMER_HTTP_CLIENT_THREAD_POOL_SIZE),
                 configFactory.getBooleanProperty(CONSUMER_HTTP_CLIENT_THREAD_POOL_MONITORING));
 
-        HttpClient client = new HttpClient(createSslContextFactory());
+        HttpClient client = sslContextFactoryProvider.provideSslContextFactory()
+                .map(sslContextFactory -> new HttpClient(sslContextFactory))
+                .orElseGet(() -> new HttpClient());
```
jewertow, line 47: It could be simplified to:
```
.map(HttpClient::new)
.orElseGet(HttpClient::new);
```
but I think it would be less understandable.
What is your opinion?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
```diff
@@ -63,18 +64,13 @@ public HttpClient createClientForHttp2() {
         http2Client.setExecutor(executor);
         HttpClientTransportOverHTTP2 transport = new HttpClientTransportOverHTTP2(http2Client);
 
-        HttpClient client = new HttpClient(transport, createSslContextFactory());
+        HttpClient client = sslContextFactoryProvider.provideSslContextFactory()
+                .map(sslContextFactory -> new HttpClient(transport, sslContextFactory))
+                .orElseThrow(() -> new IllegalStateException("Cannot create http/2 client due to lack of ssl context factory"));
```
jewertow, line 69: It is reasonable behaviour because http/2 client should be created only when both ssl and http/2 are enabled.

integration/src/integration/java/pl/allegro/tech/hermes/integration/HermesClientPublishingTest.java
```diff
@@ -67,27 +62,6 @@ public void shouldPublishUsingOkHttpClient() {
         runTestSuiteForHermesClient(client);
     }
 
-    @Test
-    public void shouldPublishUsingOkHttpClientUsingSSL() {
```
jewertow, line 71: I don't see any difference between this test and `shouldCommunicateWithHermesUsingHttp2`. I decided to remove it.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
dankraw: I prefer the committed version. :)

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/Configs.java
dankraw: I wonder If this should be a string type property (enum-like), say:
`frontend.ssl.keystore.source`, with default value equal to `jvmDefault` (the other option would be `provided`).
In this case there would be no possibility for empty value (both are `false`), plus no need for guessing which option has higher priority when both are set to `true`.

jewertow: Done

jewertow: @dankraw since we load trust store from JRE path maybe `{consumer,frontend}.ssl.{key,trust}store.source.jre` would be better name for this property?

dankraw: `jre` sounds alright. :)

PR #1197, opened 2020-04-14
Remove logback files from hermes-consumers and hermes-frontend

jewertow:
Hermes should not package logback files, because it results in warnings when users provide custom logging configuration.

Maybe could we keep these files in test/resources as templates for users, what do you think about it?


dankraw:
One can find these files in the [docker directory](https://github.com/allegro/hermes/blob/hermes-1.4.4/docker/latest/frontend/logback.xml) already, so maybe there's no need for putting them into tests dir.

PR #1221, opened 2020-06-22
SKYEDEN-981 | Exponential retry policy

### LINKED ISSUE #981: Remove the consumer inflight messages distributed counter
Consumers report the number of messages being inflight to a different zk node that all other metrics (`/hermes/consumers/<node>/group/..topic/../subscriptions/.../metrics/...`).
I believe we even don't use this path at all.
Unfortunately the path is hostname-dependent and leaves a lot of rubbish in zookeeper.

IMHO we should get rid of it as inflight is being reported to graphite anyway.

see: `ZookeeperCounterStorage`

milenakrawczyk:
Exponential retry policy for failed requests in serial subscription.

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
@@ -18,6 +18,8 @@ public class SubscriptionPolicy {
     private static final int DEFAULT_SOCKET_TIMEOUT = 0;
     private static final int DEFAULT_INFLIGHT_SIZE = 100;
     private static final int DEFAULT_SENDING_DELAY = 0;
+    private static final double DEFAULT_BACKOFF_MULTIPLIER = 1;
+    private static final int DEFAULT_BACKOFF_MAX_INTERVAL = 600;
 
     @Min(1)
     private int rate = DEFAULT_RATE;
@@ -44,6 +46,14 @@ public class SubscriptionPolicy {
     @Max(5000)
     private int sendingDelay = DEFAULT_SENDING_DELAY;
 
+    @Min(1)
+    @Max(10)
+    private double backoffMultiplier = DEFAULT_BACKOFF_MULTIPLIER;
+
+    @Min(1)
+    @Max(600)
+    private int backoffMaxInterval = DEFAULT_BACKOFF_MAX_INTERVAL;
+
     private boolean retryClientErrors = false;
 
     private SubscriptionPolicy() {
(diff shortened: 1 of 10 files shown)


hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
```diff
@@ -44,6 +46,14 @@
     @Max(5000)
     private int sendingDelay = DEFAULT_SENDING_DELAY;
 
+    @Min(1)
+    @Max(10)
+    private double backoffMultiplier = DEFAULT_BACKOFF_MULTIPLIER;
+
+    @Min(1)
+    @Max(600)
+    private int backoffMaxInterval = DEFAULT_BACKOFF_MAX_INTERVAL;
```
druminski, line 55: In Hermes we have this unwritten rule that we add a unit of time to a variable name if the unit is different than milliseconds. Therefore please consider adding here a suffix `InSec` to the variable name.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
```diff
@@ -67,6 +77,28 @@ public SubscriptionPolicy(int rate,
         this.sendingDelay = sendingDelay;
     }
 
+    public SubscriptionPolicy(int rate,
```
druminski, line 80: Do we need old public constructor, since you added new one?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Message.java
```diff
@@ -137,6 +140,13 @@ public String getId() {
         return Collections.unmodifiableList(additionalHeaders);
     }
 
+    public long updateAndGetCurrentMessageBackoff(SubscriptionPolicy subscriptionPolicy) {
+        currentMessageBackoff = currentMessageBackoff == -1 ? subscriptionPolicy.getMessageBackoff() :
```
mareckmareck, line 144: I'd recommend splitting it into several statements, since it's pretty tough on eyes at the moment :)

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
```diff
@@ -77,14 +91,17 @@ public static SubscriptionPolicy create(Map<String, Object> properties) {
                 (Boolean) properties.getOrDefault("retryClientErrors", false),
                 (Integer) properties.getOrDefault("messageBackoff", DEFAULT_MESSAGE_BACKOFF),
                 (Integer) properties.getOrDefault("inflightSize", DEFAULT_INFLIGHT_SIZE),
-                (Integer) properties.getOrDefault("sendingDelay", DEFAULT_SENDING_DELAY)
+                (Integer) properties.getOrDefault("sendingDelay", DEFAULT_SENDING_DELAY),
+                ((Number) properties.getOrDefault("backoffMultiplier", DEFAULT_BACKOFF_MULTIPLIER)).doubleValue(),
```
faderskd, line 95: why not a Duble ?

milenakrawczyk: Hermes console may send both double and integer because it cuts trailing zeros (e.g. 3.0 => 3). The only way to prevent this is by sending string which I decided to avoid. JsonCreator treats 3 as Integer and an exception is thrown when casting to double.

PR #1222, opened 2020-06-23
configurable namespace separator

trombka:
In #1209 I have made a mistake.
I used `.` as a namespace separator, because it is the way we define namespaces in our project, but I did not notice that the separator for namespaces on Kafka is `_`. It works, as for Hermes alone, but it is inconsistent a causes us some problems.
In this PR I have made the separator configurable.


trombka:
I just rebased this on the master branch.

trombka:
And unfortunately lost the approvals. :(

cristaloleg:
Thanks!

Issue #1241, opened 2020-07-15
SchemaNotFoundException returns 5xx


druminski:
When a client publishes on Hermes a message with `Schema-version` header which doesn’t  exist in `schema-registry` then Hermes returns 5xx to the client. Hermes also logs something like this:

```
pl.allegro.tech.hermes.schema.SchemaNotFoundException: No schema source for topic X.X.X at version Y
	at pl.allegro.tech.hermes.schema.SchemaRepository.getAvroSchema(SchemaRepository.java:30)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.lambda$createAvroMessage$0(MessageFactory.java:104)
	at java.base/java.util.Optional.map(Optional.java:265)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.createAvroMessage(MessageFactory.java:104)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.create(MessageFactory.java:89)
	at pl.allegro.tech.hermes.frontend.publishing.message.MessageFactory.create(MessageFactory.java:63)
	at pl.allegro.tech.hermes.frontend.publishing.handlers.MessageCreateHandler.handleRequest(MessageCreateHandler.java:38)
```

When some schema version doesn’t really exists in `schema-registry` then Hermes should return 4xx instead of 5xx to a client which send non existing schema version.

jewertow:
```
public CompiledSchema<Schema> getAvroSchema(Topic topic, SchemaVersion version) {
    if (!schemaVersionsRepository.schemaVersionExists(topic, version)) {
        throw new SchemaNotFoundException(topic, version);
    }
    return getCompiledSchemaAtVersion(topic, version);
}
```
The implementation of `schemaVersionExists` uses `SchemaVersionsRepository.versions` which fetches all versions from schema-registry and returns versions or in case of failure returns an empty list of versions. 
Instead of returning an empty list we should catch thrown exception, wrap in a dedicated one, propagate it and handle on the calling side. Then it will be possible to return 400 Bad Request in case of SchemaNotFoundException. 
What do you think?

PR #1246, opened 2020-08-02
Fix SchemaNotFoundException returns 5xx

jewertow:
## Background context
#1241

## Changes

- changed return type of `SchemaRepository.versions` from `List<SchemaVersion>` to `SchemaVersionsResponse`

`SchemaRepository.versions` was returning an empty list in case of failed request. To return 400 status to the client in case of non-existing schema version it was necessary to distinguish between the missing schema version and failed request. 
The first option was to throw an exception, but I decided to wrap the result into `SchemaVersionsResponse` to avoid the overhead of throwing an exception. Additionally it would be necessary to handle this exception in all places that invoke this method.
`SchemaVersionsResponse` is a wrapper that contains status of request execution and a list of versions (empty list in case of failure).


integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
```diff
@@ -232,7 +232,9 @@ public void shouldReturnServerInternalErrorResponseOnMissingSchemaAtSpecifiedVer
         Response response = publisher.publishAvro(topic.getQualifiedName(), message.getBody(), message.getHeaders());
 
         // then
-        assertThat(response.getStatus()).isEqualTo(INTERNAL_SERVER_ERROR.getStatusCode());
+        assertThat(response.getStatus()).isEqualTo(BAD_REQUEST.getStatusCode());
+        assertThat(response.readEntity(String.class))
+                .isEqualTo("{\"message\":\"Given schema version does not exist\",\"code\":\"SCHEMA_VERSION_DOES_NOT_EXIST\"}");
```
druminski, line 237: Can we also return what schema version doesn’t exist in http response? This information will be useful for clients.

hermes-schema/src/main/java/pl/allegro/tech/hermes/schema/CachedSchemaVersionsRepository.java
```diff
@@ -42,17 +40,19 @@ public CachedSchemaVersionsRepository(RawSchemaClient rawSchemaClient, ExecutorS
...
 
     @Override
-    public List<SchemaVersion> versions(Topic topic, boolean online) {
+    public SchemaVersionsResponse versions(Topic topic, boolean online) {
         try {
             if (online) {
                 List<SchemaVersion> v = rawSchemaClient.getVersions(topic.getName());
                 versionsCache.put(topic, v);
-                return v;
+                return SchemaVersionsResponse.succeeded(v);
```
mareckmareck, line 48: v -> versions? Since in the else branch it is fully named.

hermes-schema/src/main/java/pl/allegro/tech/hermes/schema/SchemaVersionsResponse.java
```diff
@@ -0,0 +1,44 @@
+package pl.allegro.tech.hermes.schema;
+
+import java.util.List;
+
+import static java.util.Collections.emptyList;
+
+public class SchemaVersionsResponse {
```
mareckmareck, line 7: Not sure if `Response` is the best suited name here. Perhaps `Result` would be better?

hermes-schema/src/main/java/pl/allegro/tech/hermes/schema/CachedSchemaVersionsRepository.java
jewertow: Done

hermes-schema/src/main/java/pl/allegro/tech/hermes/schema/SchemaVersionsResponse.java
jewertow: Done

integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingAvroTest.java
jewertow: Done

PR #1248, opened 2020-08-05
Subscription name validation fixed

milenakrawczyk:
- subscription name cannot contain '_'
- (old) subscription with invalid name can be edited (bug fixed)

### DIFF
--- a/hermes-console/static/partials/modal/editSubscription.html
+++ b/hermes-console/static/partials/modal/editSubscription.html
@@ -14,8 +14,8 @@
 
             <div class="form-group {{subscriptionForm.subscriptionName.$valid ? '' : 'has-error'}} " ng-show="operation === 'ADD'">
                 <label for="subscriptionName" class="col-md-3 control-label">Name</label>
-                <div class="col-md-9">
-                    <input class="form-control" id="subscriptionName" name="subscriptionName" required placeholder="name of subscription" ng-model="subscription.name" ng-pattern="/^[a-zA-Z0-9._-]+$/"/>
+                <div class="col-md-9" ng-if="operation === 'ADD'">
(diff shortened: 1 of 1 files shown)


hermes-console/static/partials/modal/editSubscription.html
```diff
@@ -14,8 +14,8 @@ <h3 class="modal-title" ng-show="operation === 'EDIT'"><small>Edit:</small> {{su
 
             <div class="form-group {{subscriptionForm.subscriptionName.$valid ? '' : 'has-error'}} " ng-show="operation === 'ADD'">
                 <label for="subscriptionName" class="col-md-3 control-label">Name</label>
-                <div class="col-md-9">
-                    <input class="form-control" id="subscriptionName" name="subscriptionName" required placeholder="name of subscription" ng-model="subscription.name" ng-pattern="/^[a-zA-Z0-9._-]+$/"/>
+                <div class="col-md-9" ng-if="operation === 'ADD'">
+                    <input class="form-control" id="subscriptionName" name="subscriptionName" ng-required="operation === 'ADD'" placeholder="name of subscription" ng-model="subscription.name" ng-pattern="/^[a-zA-Z0-9.-]+$/"/>
```
druminski, line 18: I see that the statement `operation === 'ADD'` is already declared in line `15`, therefore do we need to declare it here and in line `17` again?

milenakrawczyk: `ng-show` does not disable validation of an element that is hidden. This caused a bug in an edit form: subscriptions with wrong names could not be edited because the hidden element 'subscription name' reported an error. That is why another directive `ng-if` has been added.

PR #1250, opened 2020-08-19
Optimizing graphite metrics queries

milenakrawczyk:


hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionNameWithMetrics.java
```diff
@@ -18,7 +18,7 @@
 
     @JsonCreator
     public SubscriptionNameWithMetrics(
-            @JsonProperty("topicQualifiedName") String topicQualifiedName,
+            @JsonProperty("topicName") String topicQualifiedName,
```
milenakrawczyk, line 21: Changed so that the attribute name is equal to that present in Subscription class. This allows filtering by topicName for both classes.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/TopicNameWithMetrics.java
```diff
@@ -24,7 +24,7 @@ public TopicNameWithMetrics(
             @JsonProperty("subscriptions") int subscriptions,
             @JsonProperty("throughput") MetricDecimalValue throughput,
             @JsonProperty("volume") long volume,
-            @JsonProperty("qualifiedName") String qualifiedName
+            @JsonProperty("name") String qualifiedName
```
milenakrawczyk, line 27: Changed so that the attribute name is equal to that present in Topic class. This allows filtering by name for both classes.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/TopicService.java
```diff
@@ -374,12 +375,14 @@ public String fetchSingleMessageFromPrimary(String brokersClusterName, TopicName
     }
 
     public List<TopicNameWithMetrics> queryTopicsMetrics(Query<TopicNameWithMetrics> query) {
-        return query.filter(getTopicsMetrics())
+        List<Topic> filteredNames = query.filterNames(getAllTopics())
+                .collect(Collectors.toList());
```
jewertow, line 379: Method `toList` is already imported. You don't have to import `Collectors`.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionNameWithMetrics.java
cristaloleg: Should we change variable name too? Or create a separate variable with the name `topicName` ?

cristaloleg (approved):
sgtm

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionNameWithMetrics.java
milenakrawczyk: I am not sure but:
1) I did not want to change the api 
2) Variable _topicName_ would be exactly the same as _topicQualifiedName_
A similar thing was done in _Topic_ as well as _Subscription_ .

mareckmareck: Isn't changing `JsonProperty` already an API change? ;)

mareckmareck (dismissed):

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionNameWithMetrics.java
milenakrawczyk: Yes, it is - I used wrong wording. This change is somehow less invasive.

cristaloleg (approved):

hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionNameWithMetrics.java
druminski: I agree that this change is not so invasive, let’s try it +1

druminski (approved):

PR #1258, opened 2020-09-08
Introduce Reactive Hermes Client

wjur:
This PR introduces a `ReactiveHermesClient` class and a `ReactiveHermesSender` interface implemented using `WebClient`. Both `ReactiveHermesClient` and `ReactiveHermesSender` mimic their non-reactive counterparts' API and behaviour.  Also, the reactive classes' tests are almost identical to the non-reactive classes' tests.

### DIFF
--- a/hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
+++ b/hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
@@ -4,7 +4,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Scheduler;
-import reactor.core.scheduler.Schedulers;
 import reactor.util.retry.Retry;
 
 import java.net.URI;
@@ -15,12 +14,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Consumer;
 import java.util.function.Predicate;
 
 import static pl.allegro.tech.hermes.client.HermesMessage.hermesMessage;
+import static pl.allegro.tech.hermes.client.HermesResponseBuilder.hermesFailureResponse;
 import static reactor.core.Exceptions.isRetryExhausted;
 
 
@@ -47,7 +45,7 @@ public class ReactiveHermesClient {
                          Predicate<HermesResponse> retryCondition,
                          long retrySleepInMillis,
                          long maxRetrySleepInMillis,
-                         ScheduledExecutorService scheduler) {
+                         Scheduler scheduler) {
         this.sender = sender;
         this.uri = createUri(uri);
         this.defaultHeaders = Collections.unmodifiableMap(new HashMap<>(defaultHeaders));
@@ -55,7 +53,7 @@ public class ReactiveHermesClient {
         this.retryCondition = retryCondition;
         this.retrySleep = Duration.ofMillis(retrySleepInMillis);
         this.maxRetrySleep = Duration.ofMillis(maxRetrySleepInMillis);
-        this.scheduler = Schedulers.fromExecutor(scheduler);
+        this.scheduler = scheduler;
         this.messageDeliveryListeners = new ArrayList<>();
     }
 
@@ -108,40 +106,44 @@ public class ReactiveHermesClient {
         currentlySending.incrementAndGet();
         Mono<Result> sendOnceResult = sendOnce(message)
                 .flatMap(this::testRetryCondition)
-                .onErrorResume(this::mapExceptionToFailedAttempt)
+                .onErrorResume(exception -> mapExceptionToFailedAttempt(exception, message))
                 .subscribeOn(scheduler);
 
-        return retry(message, sendOnceResult)
+        return retry(sendOnceResult)
                 .doOnSuccess(hr -> {
-                    if (hr.response.isSuccess()) {
-                        handleSuccessfulRetry(message, hr.getAttempt());
+                    if (hr.response.isSuccess() || !hr.matchesRetryPolicy) {
+                        handleSuccessfulRetry(hr.response, hr.attempt);
                     } else {
-                        handleFailure(message, hr.getAttempt());
+                        handleFailure(hr.response, hr.attempt);
                     }
                 })
                 .map(result -> result.response)
                 .doFinally(s -> currentlySending.decrementAndGet());
     }
 
-    private Mono<Attempt> retry(HermesMessage message, Mono<Result> sendOnceResult) {
-        Retry retrySpec = prepareRetrySpec(message);
+    private Mono<Attempt> retry(Mono<Result> sendOnceResult) {
+        Retry retrySpec = prepareRetrySpec();
         return sendOnceResult
                 .flatMap(this::unwrapFailedAttemptAsException)
                 .retryWhen(retrySpec)
                 .subscriberContext(ctx -> ctx.put(RETRY_CONTEXT_KEY, HermesRetryContext.emptyRetryContext()))
-                .onErrorResume(Exception.class, exception -> unwrapRetryExhaustedException(message, exception));
+                .onErrorResume(Exception.class, this::unwrapRetryExhaustedException);
     }
 
-    private Mono<Attempt> unwrapRetryExhaustedException(HermesMessage message, Exception exception) {
+    private Mono<Attempt> unwrapRetryExhaustedException(Exception exception) {
         if (isRetryExhausted(exception)) {
             RetryFailedException rfe = (RetryFailedException) (exception.getCause());
-            handleMaxRetriesExceeded(message, rfe.attempt);
+            Failed failedAttempt = rfe.failed;
+            HermesResponse hermesResponse = failedAttempt.hermesResponse;
+            handleMaxRetriesExceeded(hermesResponse, failedAttempt.attempt);
             Throwable cause = rfe.getCause();
             if (cause instanceof ShouldRetryException) {
                 ShouldRetryException sre = (ShouldRetryException) cause;
-                return Mono.just((Attempt) Result.attempt(sre.hermesResponse, rfe.attempt));
+                return Mono.just((Attempt) Result.attempt(sre.hermesResponse, failedAttempt.attempt, true));
             }
-            handleFailure(message, rfe.getAttempt());
+            if (hermesResponse.isFailure()) {
+                handleFailure(hermesResponse, failedAttempt.attempt);
+            }
         }
         return Mono.error(exception);
     }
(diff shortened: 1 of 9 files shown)


hermes-client/src/main/java/pl/allegro/tech/hermes/client/webclient/WebClientHermesSender.java
```diff
@@ -24,10 +25,10 @@ public WebClientHermesSender(WebClient webClient) {
     }
 
     @Override
-    public CompletableFuture<HermesResponse> send(URI uri, HermesMessage message) {
+    public Mono<HermesResponse> sendReactively(URI uri, HermesMessage message) {
         return webClient.post()
                 .uri(uri)
-                .syncBody(message.getBody())
+                .syncBody(message.getBody()) // TODO reactively?
```
faderskd, line 31: hmm this is probably fine, because you already have the body of the message, if you'd have the body as the Mono, it'd be the case ;)

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
```diff
@@ -0,0 +1,314 @@
...
+    public boolean addMessageDeliveryListener(MessageDeliveryListener listener) {
+        return messageDeliveryListeners.add(listener);
+    }
+
+    private Mono<HermesResponse> publishWithRetries(HermesMessage message) {
+        currentlySending.incrementAndGet();
+
+        Retry retry = prepareRetry(message);
+
+        return sendOnce(message)
```
faderskd, line 111: this chain is huuge ;) can we divide it to the smaller parts ? I can see here few of them:
1. getting attempt result based on response
2. mapping results to Mono (with result or with exception in case of failed result)
3. handling exhausted retries

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/ReactiveHermesClientMicrometerMetricsTest.groovy
```diff
@@ -0,0 +1,161 @@
...
+
+    def "should update retries metrics"() {
+        def retries = 3
+        ReactiveHermesClient client = hermesClient(failingHermesSender(retries - 1))
+                .withRetrySleep(0)
+                .withRetries(retries)
+                .withMetrics(metricsProvider).build()
+
+
+        client.addMessageDeliveryListener(new MessageDeliveryListener() {
```
faderskd, line 79: this code is unnecessary, right ?

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/ReactiveHermesClientTest.groovy
```diff
@@ -0,0 +1,347 @@
...
+
+        when:
+        HermesResponse response = client.publish(TOPIC, CONTENT_TYPE, CONTENT).block()
+
+        then:
+        response.success
+        response.httpStatus == 202
+    }
+
+    def "should interpret message as failed for status different than 201 or 202"() {
```
faderskd, line 63: @Unroll ;)

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/ReactiveHermesClientTest.groovy
```diff
@@ -0,0 +1,347 @@
...
+        then:
+        !response.success
+        response.failure
+
+        where:
+        status << [203, 204, 400, 401, 404, 500]
+    }
+
+
+    def "should retry on http failure"() {
```
faderskd, line 82: same as above

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/ReactiveHermesClientTest.groovy
```diff
@@ -0,0 +1,347 @@
...
+        when:
+        client.publish(TOPIC, CONTENT_TYPE, CONTENT)
+            .subscribeOn(Schedulers.elastic())
+            .subscribe()
+
+        and:
+        client.closeAsync(50).block(Duration.ofSeconds(1))
+
+        then:
+        latch.await(1, TimeUnit.SECONDS)
```
faderskd, line 211: I guess you wanted here to check condition if: latch.count() == 0 ?

piotrrzysko (approved):

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
```diff
@@ -0,0 +1,324 @@
...
+                    if (retryCondition.test(response)) {
+                        return Result.failure(new ShouldRetryException(response), attempt);
+                    } else {
+                        return Result.attempt(response, attempt);
+                    }
+                });
+    }
+
+    private Retry prepareRetrySpec(HermesMessage message) {
+        if (!retrySleep.isZero()) {
```
mareckmareck, line 170: Just a detail, but this would be more readable if condition was reversed

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    private Mono<Attempt> retry(HermesMessage message, Mono<Result> sendOnceResult) {
+        Retry retrySpec = prepareRetrySpec(message);
+        return sendOnceResult
+                .flatMap(this::unwrapFailedAttemptAsException)
+                .retryWhen(retrySpec)
+                .subscriberContext(ctx -> ctx.put(RETRY_CONTEXT_KEY, HermesRetryContext.emptyRetryContext()))
+                .onErrorResume(Exception.class, exception -> {
+                    if (isRetryExhausted(exception)) {
```
mareckmareck, line 131: Rest of the functions used here are extracted, why isn't this one as well?

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
```diff
@@ -0,0 +1,324 @@
...
+    private final String uri;
+    private final Map<String, String> defaultHeaders;
+    private final AtomicInteger currentlySending = new AtomicInteger(0);
+    private final int maxRetries;
+    private final Predicate<HermesResponse> retryCondition;
+    private final Duration retrySleep;
+    private final Duration maxRetrySleep;
+    private final Scheduler scheduler;
+    private volatile boolean shutdown = false;
+    private final List<MessageDeliveryListener> messageDeliveryListeners = new ArrayList<>();
```
mareckmareck, line 40: Why not make those a part of the constructor?

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    private void handleMaxRetriesExceeded(HermesMessage message, int attemptCount) {
+        messageDeliveryListeners.forEach(l -> l.onMaxRetriesExceeded(message, attemptCount));
+        logger.error("Failed to send message to topic {} after {} attempts",
+                message.getTopic(), attemptCount);
+    }
+
+    private void handleFailedAttempt(HermesMessage message, long attemptCount) {
+        messageDeliveryListeners.forEach(l -> l.onFailedRetry(message, (int) attemptCount));
```
mareckmareck, line 227: I think we should be careful and enforce a difference between "retry" and "attempt".

Currently we have a problem with metrics in non-reactive client with distinguishing retry counter from total request counter. As a consequence, one cannot be sure, for example, whether the "retries.success" indicated that the retry was performed or not.

It is discussable whether we want to intentionally preserve that inconsistency and then fix it both here and in the non-reactive client for the sake of coherence between the reactive and non-reactive-client; or do we want to fix it in reactive client right now.

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/ReactiveHermesClientMicrometerMetricsTest.groovy
```diff
@@ -0,0 +1,132 @@
...
+        then:
+        metrics.counter("hermes-client.com_group.topic.failure").count() == 4
+        metrics.counter("hermes-client.com_group.topic.retries.count").count() == 3
+        metrics.counter("hermes-client.com_group.topic.retries.exhausted").count() == 1
+        metrics.counter("hermes-client.com_group.topic.retries.success").count() == 0
+        metrics.summary("hermes-client.com_group.topic.retries.attempts").takeSnapshot().count() == 0
+        metrics.timer("hermes-client.com_group.topic.latency").count() == 4
+    }
+
+    def "should update retries metrics"() {
```
mareckmareck, line 71: It would be nice to have more tests concerning metrics, with 0/1/2/3 out of 3 retries being performed and successful/failed.

mareckmareck (approved):
Thanks @wjur for the great pull request! :). I've added some comments, but they are in no way blocking and could be added at a later date. For this moment I think it's good to go.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
wjur: There's no particular reason. I extracted a method.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
wjur: Good idea.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
wjur: I believe it's a bug and no negation is necessary, in the first place. Please take a look at the code in the else branch:
```java
return Retry.backoff(maxRetries, retrySleep);
```
Since it's the else branch, then we already know that retrySleep is zero, and the code could be replaced with:
```java
return Retry.backoff(maxRetries, Duration.ZERO);
```
Moreover, I switched the condition and no test failed so I'm going to add some tests for this scenario.

mareckmareck: Great, nice catch :)

hermes-client/src/test/groovy/pl/allegro/tech/hermes/client/ReactiveHermesClientTest.groovy
faderskd: this form is ok ;) I didn't know what is the return value

faderskd (dismissed):

hermes-client/src/main/java/pl/allegro/tech/hermes/client/ReactiveHermesClient.java
```diff
@@ -0,0 +1,324 @@
...
+                    } else {
+                        return Result.attempt(response, attempt);
+                    }
+                });
+    }
+
+    private Retry prepareRetrySpec(HermesMessage message) {
+        if (!retrySleep.isZero()) {
+            return Retry.max(maxRetries)
+                    .doAfterRetry(signal -> handleFailedAttempt(message, signal.totalRetries() + 1));
```
wjur, line 172: `doAfterRetry` requires the signal consumer to be a non-blocking function. `MessageDeliveryListener`'s methods can be blocking so I think it's a good idea to introduce `ReactiveMessageDeliveryListener` with a similar interface as `MessageDeliveryListener` but with non-blocking methods (returning `Mono<Void>`).

The default, one listener gets injected into `ReactiveHermesClient` and it reports metrics to meter registry. I'm not sure whether the `MeterRegistry` implementation is blocking or not. If it is not blocking then the implementation of `ReactiveMessageDeliveryListener` will be pretty much straight forward.

mareckmareck (approved):

piotrrzysko (approved):

mareckmareck (approved):

faderskd (approved):

PR #1259, opened 2020-09-13
Add instructions for demo

kklin:
Hi @adamdubiel  @druminski 

We've put together the first version of the demo badge so that your users can get a temporary Hermes env for poking around!

Screen Shot 2020-09-13 at 3 43 01 PM

Everything seems to boot fine but I was getting some 404's. I'm not sure if that's because I was just using Hermes wrong, or if something isn't working.

Either way, would appreciate your feedback! Was this what you guys were imagining? What do you guys think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ Hermes
 [![Documentation Status](https://readthedocs.org/projects/hermes-pubsub/badge/?version=latest)](https://readthedocs.org/projects/hermes-pubsub/?badge=latest)
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/pl.allegro.tech.hermes/hermes-client/badge.svg)](https://maven-badges.herokuapp.com/maven-central/pl.allegro.tech.hermes/hermes-client)
 [![Join the chat](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/allegro/hermes?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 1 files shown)


druminski:
Hi @kklin, first of all thanks for the PR, it looks very promising :) In the url to `kelda.io` you provided these parameters `service=frontend&port=8080`, so module `frontend` is exposed to the world and user can publish message to Hermes. Great, however can we also expose module `management` on port 8090? I’m asking about this as UI is available on this module, port and without it user can’t manage its topics and subscriptions. I tested also the following parameters  
 `service=management&port=8090` and UI is exposed but can we expose frontend and management together?

kklin:
Thanks for taking a look! Only one service can be exposed at a time right now, but we're planning on adding support for multiple services in the next week or so.

Do you think this should block merging the initial version? I don't know enough about Hermes to know if the demo would be useful with just the frontend exposed.

druminski:
Yes, it’s a blocker for us as Hermes is useless if you initially expose only one module. However, it’s not a problem to wait with this PR a week.

kklin:
Got it, I'll update the PR once we can expose both.

kklin:
@druminski I've updated the PR! The UI is slightly different now -- each service has a "connect" button next to it if it's exposed.

druminski:
@kklin works very well, great work, thanks :)

druminski (approved):

piotrrzysko (approved):

kklin:
Thanks all! Let me know if anything goes wrong

PR #1271, opened 2020-10-01
resolve #1267 | Audit logs for unsuccessful topic and subscription operations

DyspC:
Hello, 

This PR references issue #1267 Audit logs for unsuccessful topic and subscription operations

As discussed in this issue:
- I added beforeObjectX methods to Auditor (default to no-op)
- I implemented those methods in LoggingAuditor following the same log format when possible
- I added calls to those methods in TopicService and SubscriptionService il all methods that already logged the success
- I rearranged a bit of code around the beforeObjectUpdate calls for it to be the earliest possible (especially updateSubscriptionState)

PR #1272, opened 2020-10-02
Fix possible NPE for Subscription::anonymize

DyspC:
Hello, 

Following the CI failure on #1271, I found that the Subscription::anonymize method was not null safe so here is a fixing PR for this problem.

See https://travis-ci.org/github/allegro/hermes/builds/732064554

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
@@ -445,7 +445,7 @@ public class Subscription implements Anonymizable {
 
     @Override
     public Subscription anonymize() {
-        if (endpoint != null && getEndpoint().containsCredentials() || hasOAuthPolicy()) {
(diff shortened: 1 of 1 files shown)


hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -445,7 +445,7 @@ public void setModifiedAt(Long modifiedAt) {
 
     @Override
     public Subscription anonymize() {
-        if (getEndpoint().containsCredentials() || hasOAuthPolicy()) {
+        if (endpoint != null && getEndpoint().containsCredentials() || hasOAuthPolicy()) {
```
jewertow, line 448: **Nice to have**: 
Direct access to the `endpoint` field is not consistent with existing code style. I suggest to use `getEndpoint()`.

DyspC: Hi, 
I've updated the source branch to comply

jewertow (approved):

druminski (approved):

PR #1273, opened 2020-10-02
resolves #1266 | Topic removal should be forbidden if the topic contains any subscription.

### LINKED ISSUE #1266: Topic removal should be forbidden if the topic contains any subscription
Hermes-management should not allow to remove the topic and its schema when the topic contains any subscription. In that case Hermes-management should return a response with 403 status code and clear message.

mohammedtabish:
This will make sure there are no subscribers attached to the topic before deleting the schema of the topic from the schema registry.


hermes-common/src/main/java/pl/allegro/tech/hermes/domain/topic/TopicRepository.java
```diff
@@ -29,4 +29,6 @@
     List<Topic> getTopicsDetails(Collection<TopicName> topicNames);
 
     boolean isSubscribingRestricted(TopicName topicName);
+
+    void ensureTopicIsEmpty(TopicName topicName);
```
jewertow, line 33: Since you are making it public, maybe it's a good time to rename it to `ensureTopicHasNoSubscriptions`?
If it wasn't for the name of the pull request I would think it checks if a topic has published messages.

jewertow (dismissed):

mohammedtabish:
Build's are getting failed on the first push, I will have to close the pull request and reopen to run the build again and then they are getting passed in the second attempt so please ignore the closing and opening of this pull request.

druminski (dismissed):

mareckmareck (approved):

druminski (approved):

Issue #590, opened 2016-09-27
Consumer filters test/validation endpoint


adamdubiel:
We should expose endpoint in Management that would allow on validating and testing filters.

Input:
- message
- filters chain

Output:
- is this message valid according to filter chain

Issue #513, opened 2016-06-20
Message Filter testing endpoint


chemicL:
The console should allow testing whether a particular message would pass the filters specified on the subscription.
We'd need an endpoint that, given a subscription and a message body, would return a response claiming whether the message would be filtered out.
This can help users verify their filters both via the console and using the REST API.

PR #1274, opened 2020-10-04
Moved filtering to hermes-common

piotrrzysko:
This PR moves code related to the messages filtering from the `hermes-consumers` to the `hermes-common` module. The aim of this change is to prepare for providing ability to test filters via the console (#513, #590).


hermes-common/src/main/java/pl/allegro/tech/hermes/domain/filtering/Filterable.java
```diff
@@ -0,0 +1,19 @@
+package pl.allegro.tech.hermes.domain.filtering;
+
+import org.apache.avro.Schema;
+import pl.allegro.tech.hermes.api.ContentType;
+import pl.allegro.tech.hermes.schema.CompiledSchema;
+
+import java.util.Map;
+import java.util.Optional;
+
+public interface Filterable {
```
jewertow, line 10: I am not sure if this is a proper name. `Filterable` doesn't say anything and is very general and at the same time it is very specific for `Message`.
Do we really need this interface? As go proverb says "The bigger the interface, the weaker the abstraction".

piotrrzysko: Yep, you are right that this name is too broad. What do you think about changing it to something like `FilterableMessage`? 

As for why do we need this interface: 
If we don't introduce it, there are two other possibilities: 
- Move `Message` to hermes-common. The main drawback I see in this approach is that to verify filters on the management side we will have to create instances of the class with many "dummy" arguments. As the interface shows filtering requires only four parameters, while `Message` has many more of them, e.g., `id`, `topic`, etc.
- Create a separate class with fields required only for filtering. This, in turn, will require us to create an instance of the class each time when a message goes through `FilteringMessageReceiver`. Technically it should work, but I feel it will look slightly unnaturally. What do you think?

druminski: +1 for `FilterableMessage` as methods in this interface suggest context related to Message class.

druminski (approved):

jewertow (approved):

druminski (approved):

PR #1280, opened 2020-10-10
Return more detailed message when published event does not match a schema

jewertow:
## Background context
When clients publish message that does not match the schema they receive response: `Invalid message: Failed to convert JSON to Avro`. Such a message is useless when schema is large. Then users have to ask maintainers to check logs.

## Changes
I want to return response that explains what's wrong with a published message, e.g.:
`Invalid message: Failed to convert JSON to Avro: Field <name> is expected to be type: <type>`.
In the simplest implementation it would be enough to catch the exception `AvroConversionException` thrown by `JsonConverter`, get its cause, append it to the response message and throw new exception. 
But this approach has a performance footprint so we shouldn't do it that way, therefore I suggest to add better message to `JsonConverter` in the `json-avro-converter` and upgrade its version after a release.
I already created a [pull request with this change](https://github.com/allegro/json-avro-converter/pull/55).

For now I added the jar with snapshot version of json-avro-converter and a test to prove the concept.


mareckmareck (approved):

hermes-common/build.gradle
```diff
@@ -21,7 +21,8 @@ dependencies {
     compile group: 'org.glassfish.jersey.media', name: 'jersey-media-json-jackson', version: versions.jersey
     compile group: 'org.glassfish.jersey.ext', name: 'jersey-bean-validation', version: versions.jersey
 
-    compile group: 'tech.allegro.schema.json2avro', name: 'converter', version: versions.json2avro
+//    compile group: 'tech.allegro.schema.json2avro', name: 'converter', version: versions.json2avro
+    compile files('./json-avro-converter-0.2.10-SNAPSHOT.jar')
```
mareckmareck, line 25: I'm fine with that change, remember to change this one after new version is released

mareckmareck (changes requested):

druminski (dismissed):

druminski (approved):

piotrrzysko (approved):

PR #1285, opened 2020-10-15
resolve #954 | Avoid clearing unsaved forms

### LINKED ISSUE #954: Avoid clearing unsaved topic/subscription forms
When you fill in topic/subscription form, not save it and click outside the form modal, input data is cleared and you have to do all the work from the beginning. We want to keep the input in form when it is closed (accidentally or deliberately).

pwolaq:
Avoid clearing unsaved add/edit topic/subscription forms when closing modal.

### DIFF
--- a/hermes-console/static/js/console/subscription/SubscriptionController.js
+++ b/hermes-console/static/js/console/subscription/SubscriptionController.js
@@ -258,9 +258,9 @@ subscriptions.controller('SubscriptionController', ['SubscriptionRepository', 'S
         $scope.debugFilters = function () {
             filtersDebuggerModal.open(topicName, $scope.subscription.filters)
                 .then(function (result) {
-                    var subscription = _.cloneDeep($scope.subscription);
-                    subscription.filters = result.messageFilters;
-                    $scope.edit(subscription);
+                    subscriptionDraft = _.cloneDeep($scope.subscription);
+                    subscriptionDraft.filters = result.messageFilters;
+                    $scope.edit();
                 });
         };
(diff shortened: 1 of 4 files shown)


hermes-console/static/js/console/subscription/SubscriptionController.js
```diff
@@ -82,14 +85,14 @@ subscriptions.controller('SubscriptionController', ['SubscriptionRepository', 'S
           return filtered;
         };
 
-        $scope.edit = function (subscription) {
```
piotrrzysko, line 85: As far as I remember, this function is also called from JS code (function `debugFilters` from `SubscriptionController`). I see no changes around that place, so I think it will not work correctly with filters debugger.

piotrrzysko (commented):
Have you considered opening modals this way: 
```
$modal.open({
    ...    
    backdrop: 'static',
    ...
});
```
? 

It will prevent closing it by clicking on the background. Maybe it is easier than creating drafts. WDYT?

mareckmareck (dismissed):

piotrrzysko (changes requested):
@pwolaq In my opinion, this change breaks the filter debugger. Would you mind taking a look at this comment: https://github.com/allegro/hermes/pull/1285#discussion_r505440841?

pwolaq:
@piotrrzysko it should be ok now, sorry it took so long.

piotrrzysko (approved):

mareckmareck (approved):

Issue #1256, opened 2020-09-02
Readonly mode for Kafka topics configuration


druminski:
One of our Hermes clusters works only in a subscription mode meaning that it uses only hermes-consumers and hermes-management modules. Events to this cluster are published without using Hermes. Moreover, kafka topics configuration is also managed without using Hermes in this cluster. Therefore we would like to add a "Readonly" mode for Kafka topics configuration in Hermes management. When this mode is enabled then Hermes-management should not be able to modify Kafka topics configuration such as topics retention.

PR #1288, opened 2020-10-15
Issue #1256 | Read only mode for topic configuration

milenakrawczyk:


mareckmareck (approved):

piotrrzysko:
In my opinion we should also validate this mode on server-side.

piotrrzysko (approved):

druminski (approved):

Issue #1294, opened 2020-10-20
Hermes frontend sometimes returns incorrect status 200


faderskd:
Sometimes, under unknown conditions, Hermes Frontend returns a 200 response status code after publishing a message. It isn't returned normally. Allowed success status codes during publishing are 201, 202. When this situation happens Hermes Client does not treat this message as a successful one but does not make any retries too. When Hermes Frontend returns 200 status, the message body is empty. Maybe we should treat an empty body after publishing as a sign of error and despite proper status code, we should make a retry.

druminski:
Thanks Daniel for the issue. Maybe we should investigate in what case Hermes-frontend returns 200 and instead of this status code we should return 500.

faderskd:
Yes, I think so. If it would be too hard to reproduce this bug, we can directly change hermes client

druminski:
I see that Hermes-frontend can return 200 due to default response in undertow. Hermes-frontend overwrite this default response for example in [MessageReadHandler class](https://github.com/allegro/hermes/blob/7907afb02503787760505cbfecca48369aa292f8/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageReadHandler.java#L199). But maybe there is a better way to disable this default response in the newest version of undertow.

Diewa:
I have checked this problem and I haven’t able to reproduce this bug but as Łukasz said it is probably related with default response in Undertow. I’m not sure but I think that in some cases Hermes-frontend doesn’t override this default response for example in [MessageErrorProcessor class](https://github.com/allegro/hermes/blob/7907afb02503787760505cbfecca48369aa292f8/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/MessageErrorProcessor.java#L63). 

This [PR](https://github.com/allegro/hermes/pull/1297) override an unexpected 200 status code and change it to 500.

PR #1297, opened 2020-10-27
Change default undertow http status code

Diewa:
Resolves [#1294](https://github.com/allegro/hermes/issues/1294)

Change default undertow response status

### DIFF
--- a/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageReadHandler.java
+++ b/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageReadHandler.java
@@ -4,6 +4,8 @@ import io.undertow.io.Receiver;
 import io.undertow.server.DefaultResponseListener;
 import io.undertow.server.HttpHandler;
 import io.undertow.server.HttpServerExchange;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


mareckmareck (dismissed):

druminski (approved):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageReadHandler.java
```diff
@@ -210,6 +213,13 @@ private void endWithoutDefaultResponse(HttpServerExchange exchange) {
         @Override
         public boolean handleDefaultResponse(HttpServerExchange exchange) {
             if (exchange.getAttachment(AttachmentContent.KEY).isResponseReady()) {
+                if (exchange.getStatusCode() == 200) {
+                    try {
+                        exchange.setStatusCode(500);
+                    } catch (RuntimeException e) {
+                        logger.warn("Exception has been thrown during an exchange status modification", e);
```
marcin-sucharski, line 220: As fa as I understand, `setStatusCode` can throw an exception only if response was already sent (IllegalStateException). In such case client would receive 200 response which is not documented and invalid for Hermes. There is nothing we can do about it in this place however I would use `error` log level so maybe someone will notice it in logs.

marcin-sucharski (approved):

druminski (approved):

PR #1301, opened 2020-10-30
partition-key HTTP header

Diewa:
This PR introduces a partition-key HTTP header. That allows clients to send messages to a specific partition.

I've checked this change on my local machine and it seems that the default Kafka producer works fine. See the picture below where I send several messages with the same partition-key header:
image

I’m not sure about a few things:
1. Is it a good [place](https://github.com/allegro/hermes/compare/partition-key-http-header?expand=1#diff-15e36e69dbea52e48bb3daa3f94548541186020007b7983dc0bed72186d9064a) for this header?
2. Should be partition-key header added to [KafkaHeaderFactory](https://github.com/allegro/hermes/blob/master/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaHeaderFactory.java)?
3. Should we somehow prevent compatibility problems due to BackupMessage change?


hermes-frontend/src/test/groovy/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerMessageProducerIntegrationTest.groovy
```diff
@@ -0,0 +1,205 @@
...
+                new KafkaHeaderFactory(configFactory),
+                configFactory)
+    }
+
+    def "should publish messages on only one partition for the same partition-key"() {
+        Topic topic = createAvroTopic("pl.allegro.test.Foo")
+        Subscription subscription = createTestSubscription(topic, "test-subscription")
+        String kafkaTopicName = topic.getName().toString()
+        ConsumerGroupId consumerGroupId = kafkaNamesMapper.toConsumerGroupId(subscription.qualifiedName)
+        createTopicInKafka(kafkaTopicName, 3)
```
marcin-sucharski, line 105: I would extract `3` to a constant because it is used multiple times in tests and looks like magic number

hermes-frontend/src/test/groovy/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerMessageProducerIntegrationTest.groovy
```diff
@@ -0,0 +1,205 @@
...
+        Topic topic = createAvroTopic("pl.allegro.test.Foo")
+        Subscription subscription = createTestSubscription(topic, "test-subscription")
+        String kafkaTopicName = topic.getName().toString()
+        ConsumerGroupId consumerGroupId = kafkaNamesMapper.toConsumerGroupId(subscription.qualifiedName)
+        createTopicInKafka(kafkaTopicName, NUMBER_OF_PARTITION)
+        CachedTopic cachedTopic = CachedTopicsTestHelper.cachedTopic(topic)
+
+        when:
+        1.upto(10) {
+            brokerMessageProducer.send(generateAvroMessage("partition-key"), cachedTopic, null)
```
druminski, line 111: You send messages asynchronously here and assume that they are already in Kafka, when `then` section is executed. At first glance, this test can be unstable, especially in unstable environment such as our CI. To stabilize it, I suggest to wait for messages being  published successfully on Kafka and then check offsets on the Kafka topic. This comment also applies to the second test as well.

hermes-frontend/build.gradle
```diff
@@ -23,6 +23,8 @@ dependencies {
     testCompile(group: 'com.jayway.awaitility', name: 'awaitility', version: '1.6.1') {
         exclude group: 'com.jayway.jsonpath', module: 'json-path'
     }
+    testCompile group: 'org.testcontainers', name: 'spock', version: '1.15.0-rc2'
```
druminski, line 26: `1.15.0` is available now

druminski (commented):
I’ve added few minor comments, but overall very good PR ❤️ As for your questions from description:

Ad 1. Yes, it’s good place.
Ad 2. [ConsumerRecord](https://kafka.apache.org/21/javadoc/org/apache/kafka/clients/consumer/ConsumerRecord.html#key--) contains key() allowing user to read partition key. I don’t see any reason why we should duplicate partition key in kafka header as well, let’s don’t do it.
Ad 3. Good question. Ideally if Hermes update with this change will be smooth.

hermes-frontend/src/test/groovy/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerMessageProducerIntegrationTest.groovy
Diewa: I've modified tests by adding a waiting mechanism. Please let me know whether it's ok for you :)

Diewa:
> Ad 3. Good question. Ideally if Hermes update with this change will be smooth.

I've had a problem with deserialization so I've added this [change](https://github.com/allegro/hermes/pull/1301/files#diff-e5bd5b69f38475588a73dac7f38f28f5b3adea537010e436e875cc7c92b04892R8)

I've check this for a few backup files from different environments and it seems to work fine right now :)

druminski (approved):

mareckmareck (approved):

PR #1305, opened 2020-11-06
Moved 'consistency checker' config to Properties class

druminski:
Thanks this PR there is no need to provide additional configuration related to `consistency checker` when running custom Hermes.


piotrrzysko:
For me, it looks ok. I would like to explain why I've used `@Value` instead of a class from the `config` package. Currently, the `config` package depends on `domain` so I didn't want to introduce a cycle between them (in fact, the cycle exists right now because the `TopicProperties` class is used in 3 places in the `domain` package, but I've assumed that it is oversight). 
Probably the solution for this problem is to remove spring annotations from the `domain` package and create all beans in `config`. Nevertheless, it is a minor thing and surely it has not to be done in this PR.

piotrrzysko (approved):

PR #1307, opened 2020-11-13
Resolve #1304 | Json support in debugger fix

### LINKED ISSUE #1304: Json validation fix in debugging filters in hermes-console

milenakrawczyk:


druminski (approved):

mareckmareck (approved):

hermes-console/static/js/console/subscription/SubscriptionController.js
```diff
@@ -252,7 +252,7 @@ subscriptions.controller('SubscriptionController', ['SubscriptionRepository', 'S
         };
 
         $scope.debugFilters = function () {
-            filtersDebuggerModal.open(topicName, $scope.subscription.filters)
+            filtersDebuggerModal.open(topicName, $scope.subscription.filters, topicContentType)
```
piotrrzysko, line 255: Does it work? There is no local variable named `topicContentType` in scope of this controller. In my opinion you should use `$scope.topicContentType`.

piotrrzysko (commented):
This fix looks fine, but it doesn't resolve #1304. The issue is about replacing the `textarea` where a user puts a message for debugging with something more user-friendly. It would be nice to support JSON formatting, validating, and syntax highlighting.

hermes-console/static/js/console/subscription/SubscriptionController.js
milenakrawczyk: 1. The #1304 issue has not been described in detail and it was obvious that validating json message did not work. It was natural to assume that the issue was to provide a support for debugging json messages (fixing bug). I will create another issue regarding the formatting so that it can be tackled in the future.
2. It seems to work. I have added the prefix because maybe I have not tested all possible scenarios.

piotrrzysko (approved):

druminski (approved):

mareckmareck (approved):

PR #1316, opened 2021-01-13
Allow header propagation just by using configuration

althink:
After this change it's possible to enable header propagation and define which headers should be propagated

frontend.header.propagation.enabled=true
frontend.header.propagation.allow.filter=traceparent


piotrrzysko:
@althink thanks for the PR. Would you mind adding some tests to check that the new header propagation strategy works correctly?

althink:
Hi @piotrrzysko, I've added some tests

piotrrzysko (approved):

milenakrawczyk (approved):

PR #1327, opened 2021-03-04
request timeout can be viewed and modified in console

tadamcze:


hermes-console/static/partials/modal/editSubscription.html
```diff
@@ -140,6 +140,17 @@ <h3 class="modal-title" ng-show="operation === 'EDIT'"><small>Edit:</small> {{su
             </div>
 
             <hr/>
+            <div class="form-group {{subscriptionForm.requestTimeout.$valid ? '' : 'has-error'}}">
```
piotrrzysko, line 143: Would you mind adding a warning for the user who tries to set a request timeout over 1000ms (current default value)? It is quite high value and to comply with the fail-fast principle maybe in some cases it is better to rethink the design of a subscriber rather than increase the timeout.

druminski:
Closing this PR due to no activity. Opened another one #1353 which also exposed request timeout in UI and displays warning for bigger value. Thanks @tadamcze for the contribution.

PR #1345, opened 2021-06-02
Added UI panel for http header filters

druminski:
Added `Http header filter` panel in subscription view. The panel is displayed below `Filter` panel as in the attached image. By default `Http header panel` is disabled. It can be enabled by `showHeadersFilter` flag in Hermes console config. At the moment the panel dosen't support `Debug` mode and `matching strategy`, but they can be added in the future.
 
Zrzut ekranu 2021-06-2 o 15 53 18

### DIFF
--- a/hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/FilteringConfiguration.java
+++ b/hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/FilteringConfiguration.java
@@ -6,6 +6,7 @@ import pl.allegro.tech.hermes.domain.filtering.MessageFilters;
 import pl.allegro.tech.hermes.domain.filtering.SubscriptionMessageFilterCompiler;
 import pl.allegro.tech.hermes.domain.filtering.avro.AvroPathSubscriptionMessageFilterCompiler;
 import pl.allegro.tech.hermes.domain.filtering.chain.FilterChainFactory;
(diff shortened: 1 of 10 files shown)


hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/FilteringConfiguration.java
```diff
@@ -20,7 +21,8 @@
     FilterChainFactory filterChainFactory() {
         List<SubscriptionMessageFilterCompiler> subscriptionFilterCompilers = Arrays.asList(
                 new AvroPathSubscriptionMessageFilterCompiler(),
-                new JsonPathSubscriptionMessageFilterCompiler()
+                new JsonPathSubscriptionMessageFilterCompiler(),
+                new HeaderSubscriptionMessageFilterCompiler()
```
piotrrzysko, line 25: As far as I remember, hermes-management is aware of the filters only because it has a logic related to the filters debugger. As the debugger doesn't currently support header filters, I think this change is not needed.

druminski: You are right, this change is not needed. However I added it to simplify adding support for headers dubbuger in the future. Thanks to this line only changes on frontend side are needed now and next developer don't have to analyze backend side at all.

piotrrzysko (approved):

faderskd (approved):

PR #1325, opened 2021-02-28
added ssl configs for frontend, consumer and management

spattanaik75:

PR #1349, opened 2021-06-25
Added ability to specify the default kafka-client SSL and SASL configuration keys

cywattebled:
if kafka.authorization.enabled is true.
This allows for freely specifying paramaters for securing connectivity between Hermes and
the underlying kafka brokers beyond SASL_PLAINTEXT and username/password

Kafka client properties will require the prefix "kafka.client." within Hermes configuration.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/KafkaClientConfig.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/KafkaClientConfig.java
@@ -1,0 +1,54 @@
+package pl.allegro.tech.hermes.common.config;
+
+import com.netflix.config.DynamicPropertyFactory;
+import com.netflix.config.PropertyWrapper;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.SaslConfigs;
+import org.apache.kafka.common.config.SslConfigs;
+
+import javax.annotation.Nonnull;
+
+/**
+ * @author Cyril Wattebled (cyril.wattebled@idemia.com)
+ */
+public class KafkaClientConfig {
+    public static final String KAFKA_CLIENT_PREFIX = "kafka.client.";
+
+    public interface CollectionPut {
+        Object put(String s, Object value);
+    }
+
+    public static void loadKakfaConfig(@Nonnull CollectionPut lambda) {
+        ConfigDef config = new ConfigDef();
+        SslConfigs.addClientSslSupport(config);
+        SaslConfigs.addClientSaslSupport(config);
+        DynamicPropertyFactory instance = DynamicPropertyFactory.getInstance();
+        config.configKeys().forEach((s, configKey) -> {
+            String propKey = KAFKA_CLIENT_PREFIX + s;
+            PropertyWrapper<?> value = null;
+            switch (configKey.type) {
+                case BOOLEAN:
+                    value = instance.getBooleanProperty(propKey, (Boolean) configKey.defaultValue);
+                    break;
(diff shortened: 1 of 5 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/KafkaClientConfig.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Cyril Wattebled (cyril.wattebled@idemia.com)
+ */
+public class KafkaClientConfig {
+    public static final String KAFKA_CLIENT_PREFIX = "kafka.client.";
+
+    public interface CollectionPut {
+        Object put(String s, Object value);
+    }
+
+    public static void loadKakfaConfig(@Nonnull CollectionPut lambda) {
```
piotrrzysko, line 21: If I understand correctly, the mechanism implemented here allows loading any Kafka's property, that is related to SSL or SASL. Right? I would rather stay with the current approach to loading configuration where properties are defined explicitly.

Please, take a look at this PR: #1325. It also adds support for configuring SSL. Maybe it is better to finish that one as it also adds the possibility to secure connections between hermes-management and Kafka.

cywattebled: Yes, the intent is to allow any of default SSL/SASL keys to be used  in the Hermes config.properties without having to explicitly add them at the code level, and this every time you update the Kafka Client.
This allows the different situations that underlying kafka might throw at you as a Kafka Client.

What concerns are you seeing about this approach ? It's basically a passthrough to the Kafka client and only if these properties are not set elsewhere overwise which means if Hermes wants specific configuration keys, you can't override these underlying property through this mechanism.

This PR seems only for SSL though. In our case, we also need  SASL settings to allow Hermes to AWS MSK security through. See the unit test associated with this.

piotrrzysko: So, my main concern about introducing the new configuration loading mechanism is that it will complicate the code. I feel that for someone who will read this code in the future, it can be a bit confusing: "Why are these particular properties loaded differently?". 

That's my point of view :). Let's ask other Hermes maintainers what they think about it. For example: @druminski, @faderskd.

cywattebled: Thanks for the comment. IMHO, I see this a simplification because there's not need to duplicate/introduce new sets of configuration keys for each Kafka client SSL/SASL options.

We are close to production, I'd appreciate quick resolution on this.
I can prepare another PR exposing only what we need but we don't know the full scope yet.

cisrudlow:
Hi,
I'm also interested in this PR. It looks cleaner than the one mentioned in the first comment and solves the huge problem of encrypting communication between Hermes and Kafka. It is a Kafka-compliant solution, not a custom one.
@druminski @faderskd what do you think about that?

@cywattebled provided very good reasoning.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/KafkaClientConfig.java
cywattebled: Hi Guys, friendly nudge on this. @druminski @faderskd. Would you care to comment please?

cywattebled:
Still no change on this ?

Diewa:
Closing due to no activity. Feel free to re-open if needed

Issue #1260, opened 2020-09-16
Hermes-management publishes audit events about topics and subscriptions changes


druminski:
Hermes has [Auditor](https://github.com/allegro/hermes/blob/7907afb02503787760505cbfecca48369aa292f8/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/Auditor.java) interface and its current implementation is [LoggingAuditor](https://github.com/allegro/hermes/blob/7907afb02503787760505cbfecca48369aa292f8/hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/audit/LoggingAuditor.java) which logs information about who and what changed on topics and subscriptions. For example when Hermes user creates or removes topic then LoggingAuditor logs information about this operation. It’s very valuable feature for us and we would like to extend it about emitting audit events to a provided URL (for us it will be Hermes topic). Thanks to this other service(s) can get information about topics and subscriptions changes.

Acceptance criteria:

1) Hermes-management publishes audit events about topics and subscriptions changes
2) Audit event should provide an information about what exactly have changed (diff of topic or subscription), what is the new state, and who submitted the change.
3) Configuration of publishing audit events allows to:
    - disable/enable publishing. By default it should be disabled.
    - set an URL on which audit events are published.

DyspC:
Hello, 

I could probably work on this one but I need some information on how to run it locally in dev mode as I didn't see any gradle task for it. (eg. point a local dockerized kafka/zk/cache)

More related to this issue, I'd like to know if the settings can be the same one as the LoggingAuditor (audit.enabled), the config keys could be something like audit.log.enabled and audit.event.enabled but that means breaking changes for those who still use the old key.

The way I'd do this is by adding a primary CompositeAuditor bean that would hold a list of all other auditors in the spring registry and forward all calls to them. This way any future auditor implementation would be used as soon as spring knows of its existence.
The side effect would be deleting the Auditor.noOpAuditor as obsolete because not enabling any auditor would make the CompositeAuditor no-op itself.

jewertow:
Hello @DyspC,
if you want to run Hermes locally you have few options to do this:
1) build packages and run: https://hermes-pubsub.readthedocs.io/en/latest/deployment/packaging/#packaging - but you have to provide dependencies (kafka, zookeeper)
2) run Hermes in docker with kafka, zookeeper and graphite: https://github.com/allegro/hermes/tree/master/docker
[...]

PR #1359, opened 2021-07-15
Add support for publication of audit events

szczygiel-m:
Added emitting audit events to a provided URL (for example Hermes topic). Thanks to this other service(s) can get information about topics and subscriptions changes. You can read more about this [here](https://github.com/allegro/hermes/issues/1260)

This PR is a continuation of [#1317](https://github.com/allegro/hermes/pull/1317) by @DyspC, thanks for the contribution :)


eobrusnik (commented):
May I suggest using `JSONAssert` library for verifying payloads of HTTP responses? It will not fail if content is formatted differently or order of keys is changed.
Also, would you consider loading expected JSONs from files? Java's string syntax makes it hard to define a JSON string without cluttering content with escape characters.

integration/src/integration/java/pl/allegro/tech/hermes/integration/management/GroupManagementTest.java
```diff
@@ -16,6 +18,71 @@
...
+                "    \"groupName\": \"exampleGroup\",\n" +
+                "  },\n" +
+                "  \"username\": \"[anonymous user]\"\n" +
+                "}";
+
+        //when
+        management.group().create(group("exampleGroup").build());
+
+        //then
+        assertThat(remoteService.waitAndGetLastRequest().getBodyAsString().equals(expectedBody));
```
eobrusnik, line 39: Does this actually assert anything? According to [AssertJ's docs](https://assertj.github.io/doc/#forgetting-to-call-an-assertion) the assertion is not called in such cases. What you are essentially doing here is calling `assertThat()` with a boolean value. What you should do is call
```java
assertThat(
  remoteService.waitAndGetLastRequest().getBodyAsString()
).isEqualTo(expectedBody)
```

Please review other instances of using AssertJ, too.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/audit/AuditEvent.java
```diff
@@ -0,0 +1,47 @@
+package pl.allegro.tech.hermes.management.infrastructure.audit;
+
+public class AuditEvent {
+    private final AuditEventType eventType;
+    private final String resourceName;
+    private final Class<?> payloadClass;
```
piotrrzysko, line 6: I guess there is no similar problem with this field like I described bellow, but for consistency let's change its type to `String` as well.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/audit/AuditEvent.java
```diff
@@ -0,0 +1,47 @@
...
+    private final Object payload;
```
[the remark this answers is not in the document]
szczygiel-m: Thanks for the suggestions, I agree with You. I changed the types to String, also I deleted one constructor in `AuditEvent` to standardize the JSON request format. I used `ObjectMapper` and the payload looks good. Please take a look at my newest commit.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/audit/EventAuditor.java
```diff
@@ -0,0 +1,124 @@
...
+                    username);
+            restTemplate.postForObject(eventDestination, event, Void.class);
+        });
+    }
+
+    @Override
+    public void objectUpdated(String username, Object oldObject, Object newObject) {
+        ignoringExceptions(() -> {
+            Diff diff = javers.compare(oldObject, newObject);
+            String diffToString = objectMapper.writeValueAsString(diff.toString());
```
piotrrzysko, line 101: `org.javers.core.diff.Diff#toString` returns text like:
```
Diff:
* changes on pl.allegro.tech.hermes.api.Subscription/sample-subscription :
  - 'createdAt' changed from '2021-07-23T11:20:56.517Z' to ''
  - 'modifiedAt' changed from '2021-07-23T11:42:34.266Z' to ''
  - 'state' changed from 'ACTIVE' to 'SUSPENDED'
```

I know I mentioned earlier about serializing the payload to JSON, but I think there is no point of doing that in this case (actually, I am surprised that it works :)). So maybe here let's just use the result returned by `org.javers.core.diff.Diff#toString` as the payload. WDYT?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/audit/AuditEvent.java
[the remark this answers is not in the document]
piotrrzysko: Nice. I added one extra comment.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/audit/EventAuditor.java
szczygiel-m: You are right. That code was completely unnecessary and just `diff.toString()` is enough. I was also considering using something like `String diffToString = objectMapper.writeValueAsString(diff);`, but  in that case payload was too big, it contained all properties, not just changed ones, so it would be really hard to spot the difference.

piotrrzysko (approved):

milenakrawczyk (approved):

PR #1361, opened 2021-07-20
Topic name is now published in tags by MicrometerMetricProvider.

szczygiel-m:
hermes_client produces metrics for topics using tags instead of placing the topic / operation name into the path.

### DIFF
--- a/hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MetricsMessageDeliveryListener.java
+++ b/hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MetricsMessageDeliveryListener.java
@@ -19,59 +19,59 @@ public class MetricsMessageDeliveryListener implements MessageDeliveryListener {
     @Override
     public void onSend(HermesResponse response, long latency) {
         HermesMessage message = response.getHermesMessage();
-        String prefix = MetricsUtils.getMetricsPrefix(message.getTopic());
+        String topic = MetricsUtils.sanitizeTopic(message.getTopic());
 
-        metrics.timerRecord(prefix + ".latency", latency, NANOSECONDS);
+        metrics.timerRecord(topic, "latency", latency, NANOSECONDS);
         Map<String, String> tags = new HashMap<>();
         tags.put("code", String.valueOf(response.getHttpStatus()));
-        metrics.counterIncrement(prefix, "status", tags);
+        metrics.counterIncrement(topic, "status", tags);
 
-        counterIncrementIf(prefix + ".publish.failure", response.isFailure());
+        counterIncrementIf(topic,"publish.failure", response.isFailure());
     }
 
     @Override
     public void onFailure(HermesResponse response, int attemptCount) {
-        String prefix = MetricsUtils.getMetricsPrefix(response.getHermesMessage().getTopic());
-        metrics.counterIncrement(prefix + ".failure");
+        String topic = MetricsUtils.sanitizeTopic(response.getHermesMessage().getTopic());
+        metrics.counterIncrement(topic, "failure");
     }
 
     @Override
     public void onFailedRetry(HermesResponse response, int attemptCount) {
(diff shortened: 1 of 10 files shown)


hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MetricsMessageDeliveryListener.java
```diff
@@ -19,59 +19,59 @@ public MetricsMessageDeliveryListener(MetricsProvider metrics) {
...
 
-        metrics.timerRecord(prefix + ".latency", latency, NANOSECONDS);
+        metrics.timerRecord(topic, "latency", latency, NANOSECONDS);
         Map<String, String> tags = new HashMap<>();
         tags.put("code", String.valueOf(response.getHttpStatus()));
-        metrics.counterIncrement(prefix, "status", tags);
+        metrics.counterIncrement(topic, "status", tags);
 
-        counterIncrementIf(prefix + ".publish.failure", response.isFailure());
+        counterIncrementIf(topic,"publish.failure", response.isFailure());
```
Kornelos, line 29: minor: space after comma

hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MicrometerMetricsProvider.java
```diff
@@ -17,30 +18,40 @@ public MicrometerMetricsProvider(MeterRegistry metrics) {
...
+                .collect(Collectors.toSet())))
+                .record(value);
     }
 
-    private String buildCounterName(String prefix, String name, Map<String, String> tags) {
-        return prefix + "." + name + "."
-                + tags.keySet().stream().map(e -> "{" + e + "}").collect(Collectors.joining("."));
+    private String buildCounterName(String topic, String key, Map<String, String> tags) {
+        tags.put("topic", topic);
+        return prefix + "{topic}" + "." + key
```
piotrrzysko, line 53: Have you tested it with graphite and without our internal SDK? As far as I know micrometer doesn't support this notation with placeholders. I see it was also used in this class before your change but I suspect it was a bug.

For more details, please take look at this issue: https://github.com/micrometer-metrics/micrometer/issues/620

hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MicrometerMetricsProvider.java
```diff
@@ -17,30 +18,40 @@ public MicrometerMetricsProvider(MeterRegistry metrics) {
     }
 
     @Override
-    public void counterIncrement(String name) {
-        metrics.counter(name).increment();
+    public void counterIncrement(String topic, String key) {
+        counterIncrement(topic, key, new HashMap<>());
```
piotrrzysko, line 22: I suggest replacing `new HashMap<>()` with `Collections.emptyMap()` - just to be a little bit gc friendly :)

szczygiel-m: Take a look at the called function. I am calling `put` at this object. As far as I know, `put` called on `Collection.emptyMap()` will cause an error.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/metrics/MicrometerMetricsProvider.java
szczygiel-m: You are right, that wouldn't work without proper placeholders implementation in internal SDK. As You said, it was used here before for almost 2 years. There's also `DropwizardMetricProvider` implementation, which don't use this notation so it can be also used by clients, whose aren't familiar with placeholders.

piotrrzysko (approved):

faderskd (approved):

PR #1371, opened 2021-09-02
Improved subscription logging

szczygiel-m:
Subscription name in description for subscription health problems.

### DIFF
--- a/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/problem/LaggingIndicator.java
+++ b/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/problem/LaggingIndicator.java
@@ -20,7 +20,7 @@ public class LaggingIndicator implements SubscriptionHealthProblemIndicator {
         long subscriptionLag = context.getLag();
         double topicRate = context.getTopicRate();
         if (topicRate > 0.0 && subscriptionLag > maxLagInSeconds * topicRate) {
-            return Optional.of(lagging(subscriptionLag));
+            return Optional.of(lagging(subscriptionLag, context.getSubscription().getQualifiedName().toString()));
         }
         return Optional.empty();
     }
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Hermes
 ======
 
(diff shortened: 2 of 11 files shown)


hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/problem/LaggingIndicator.java
```diff
@@ -20,7 +20,7 @@ public LaggingIndicator(int maxLagInSeconds) {
         long subscriptionLag = context.getLag();
         double topicRate = context.getTopicRate();
         if (topicRate > 0.0 && subscriptionLag > maxLagInSeconds * topicRate) {
-            return Optional.of(lagging(subscriptionLag));
+            return Optional.of(lagging(subscriptionLag, context.getSubscription().getQualifiedName().toString()));
```
Diewa, line 23: just asking - is it null-safe?

Diewa (approved):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/health/problem/LaggingIndicator.java
szczygiel-m: Yes, this value is always not null. Deep down below in subscription class this is made from annotated `@NotNull`subscription name and topic name, just right [here](https://github.com/allegro/hermes/blob/master/hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java#L141). Also when someone hits API to check subscriptions health, he provides topicName and subscriptionName from which service gets subscriptionDetails right [here](https://github.com/allegro/hermes/blob/bf03612269bf5f64dd5e944e494595e9a1024a59/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/SubscriptionService.java#L238) which checks if subscription exists [here](https://github.com/allegro/hermes/blob/8047a92a94e312cc29e6f9a23f7e531033fa5095/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperSubscriptionRepository.java#L98) and will cause an exception when subscription isn't present.

Kornelos (approved):

PR #1379, opened 2021-10-04
resolve #1372 | cohere retention time validation between backend and frontend

### LINKED ISSUE #1372: Cannot set retention time for topic greater than 7 days using hermes console
When we set 7 days, form passes validation:
image
However, when we set more than 7 days (for example 8 days), form does not pass validation:
image

This is misleading, because using API we can set more than 7 days:
```
POST /topics
{
  "name": "topicName",
  "retentionTime": {
    "duration": 14,
    "retentionUnit": "DAYS"
  },
  ...,
  
}
```
The above request ends with success :)

dominikbrandon:
This resolves issue #1372. I assumed that validation on the frontend side is the desired one and adjusted the backend one so they match. Merging this PR will make these two coherent. You can find the frontend validation [here](https://github.com/allegro/hermes/blob/fa416ab9f0c6711a1f443e76e23d38fd225cd696/hermes-console/static/partials/modal/editTopic.html#L84).

Please let me know if this should work another way. This is a simple change, I'll fix it right away.


integration/build.gradle
```diff
@@ -1,4 +1,5 @@
 repositories {
+    jcenter()
```
dominikbrandon, line 2: JBoss repository seemed to experience some issues - see [this CI run](https://github.com/allegro/hermes/runs/3793959021?check_suite_focus=true). CI passed after adding jcenter as an alternative.

piotrrzysko:
Hi, thanks for the contribution 🙂 . In general, this is a nice improvement, as it helps protect Kafka cluster from being overloaded. However, I have one concern about adding this validation. Sometimes, in emergency situations, it is necessary to set this property to a value greater than 7 days. Ideally, it would be nice to leave this possibility to admins. This approach requires a little bit more work, but I'd love to help you if you want to work on that. I'd suggest you start by looking at this class: [ManagementRights](https://github.com/allegro/hermes/blob/master/hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/auth/ManagementRights.java).

dominikbrandon:
Hi @piotrrzysko thanks for your input. I've looked into that a bit and I'd like to discuss the desired solution. Basically my intention is to leverage the `javax.validation` API and its annotations, as I find it convenient that they are declared inside the data model. The other thing is - I want to avoid any solution requiring separate special method per every validation, which would be the simplest way to code that, but it doesn't scale well.

Therefore I discovered the `groups` as a part of `javax.validation` API. As far as I understand it, one may declare the constraint as such: `@Max(value = 7, groups = {NotAdmin.class})`, and then depending on the user role, execute `validator.validate(object)` when it's an admin, and `validator.validate(object, Default.class, NotAdmin.class)` otherwise. This might work, but it would require passing `SecurityContext` down to the `ApiPreconditions`.

What are your thoughts? Did you have any specific solution in mind?

piotrrzysko:
To be honest, my first thought was to add validation like:
```
if (isNotAdmin(...)) {
  validateRetentionTime(...);
}
```
inside `TopicService`, and call it in each method responsible for modifying topics. However, I agree with you that it would be more elegant and convenient to add this validation inside the data model. So let’s do it.

The only thing I would suggest is not to pass `SecurityContext` down to the `ApiPreconditions`. Instead, I would create a dedicated class that has information about the role along with username. I would do this because username (`createdBy`, `removedBy`, `modifiedBy`, etc.) is already passed to methods like `TopicService.createTopicWithSchema`, and I want to avoid adding another argument. 

WDYT?

dominikbrandon:
Sounds great, thanks. I've made this change. Please let me know what are your thoughts about this solution. I've had some trouble writing integration tests, as by default the user is an admin and I didn't know how to prepare such a test state that would make him a regular user. Therefore I based on the unit tests. If you've got any idea how to do that, I'll be happy to prepare the change.

PR #1403, opened 2021-10-20
Added endpoint and frontend button to get all clients by topic

szczygiel-m:


hermes-api/src/main/java/pl/allegro/tech/hermes/api/endpoints/AllTopicClientsEndpoint.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.List;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Path("topics")
+public interface AllTopicClientsEndpoint {
+
+    @GET
+    @Produces(APPLICATION_JSON)
```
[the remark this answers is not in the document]
szczygiel-m: I get what you mean but I followed the same rules we're actually returning list of objects, e.g topics, subscriptions, offlineRetransmissionTasks etc. I think there is enough flexibility in the TopicClients itself :D

hermes-api/src/main/java/pl/allegro/tech/hermes/api/endpoints/AllTopicClientsEndpoint.java
```diff
@@ -0,0 +1,19 @@
...
+    @Path("/{topic}/clients")
+    List<String> getTopicClients(@PathParam("topic") String topic);
```
pitagoras3, line 18: What about adding this method to [TopicEndpoint](https://github.com/allegro/hermes/blob/master/hermes-api/src/main/java/pl/allegro/tech/hermes/api/endpoints/TopicEndpoint.java)? It is already registered on path `@Path("topics")` :)

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/AllTopicClientsConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+import pl.allegro.tech.hermes.domain.subscription.SubscriptionRepository;
+import pl.allegro.tech.hermes.management.domain.clients.AllTopicClientsServiceImplementation;
+
+@Configuration
+public class AllTopicClientsConfiguration {
+
+    private static final Logger logger = LoggerFactory.getLogger(AllTopicClientsConfiguration.class);
+
+    @Bean
+    @ConditionalOnMissingBean(name = "allTopicClientsService")
```
pitagoras3, line 17: Is there possibility that this bean will be missing? I think that this annotation is easily avoidable here.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/AllTopicClientsConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+    public AllTopicClientsServiceImplementation allTopicClientsService(SubscriptionRepository subscriptionRepository) {
```
pitagoras3, line 18: Suggestion - when you're creating bean which is implementation of some interface, type of the bean should be interface which it is implementing. This is for [IoC and DI](https://www.baeldung.com/inversion-control-and-dependency-injection-in-spring). In this case, this line should look like:
```java
public AllTopicClientsService allTopicClientsService(SubscriptionRepository subscriptionRepository) {
```

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/AllTopicClientsConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+        logger.info("Creating allTopicClientsService bean");
```
pitagoras3, line 19: I think that we can skip logging bean creation :)

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/clients/AllTopicClientsServiceImplementation.java
```diff
@@ -0,0 +1,21 @@
...
+
+    private final SubscriptionRepository subscriptionRepository;
+
+    public AllTopicClientsServiceImplementation(SubscriptionRepository subscriptionRepository) {
+        this.subscriptionRepository = subscriptionRepository;
+    }
+
+    public List<String> getAllClientsByTopic(TopicName topicName) {
+        return subscriptionRepository.listSubscriptions(topicName)
+                .stream().map(e -> e.getOwner().getId()).distinct().collect(Collectors.toList());
```
pitagoras3, line 19: Can you change `e` into `subscription` to make this lambda easier-to-read?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/AllTopicClientsConfiguration.java
szczygiel-m: As I mentioned above this will help me to ensure that app is using the right bean.

hermes-api/src/main/java/pl/allegro/tech/hermes/api/endpoints/AllTopicClientsEndpoint.java
szczygiel-m: I would like to keep it away from TopicEndpoint, but I have changed the path to be unique, like it is for example in [SchemaEndpoint](https://github.com/allegro/hermes/blob/master/hermes-api/src/main/java/pl/allegro/tech/hermes/api/endpoints/SchemaEndpoint.java)

faderskd (approved):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/clients/AllTopicClientsService.java
```diff
@@ -0,0 +1,10 @@
+package pl.allegro.tech.hermes.management.domain.clients;
+
+import pl.allegro.tech.hermes.api.TopicName;
+
+import java.util.List;
+
+public interface AllTopicClientsService {
```
faderskd, line 7: [not so important] Do we need this interface if its has only one implementation ? Then you will not need to create names like `AllTopicClientsServiceImplementation`

szczygiel-m: I would like to change the implementation in our internal infrastructure, so it would be nice to have this interface

piotrrzysko: I guess the interface is here to inject an internal implementation (for example to return emails fetched from an internal system instead of owner ids). However, I agree that `AllTopicClientsServiceImplementation` could be named more descriptively - in my opinion something like `DefaultAllTopicClientsService` or `OwnerIdBasedAllTopicClientsService` would be better.

szczygiel-m: That's correct, thanks. I renamed `AllTopicClientsServiceImplementation` to `DefaultAllTopicClientsService`.

piotrrzysko (approved):

PR #1420, opened 2021-11-16
resolve #1418 | enable fetching schema by schemaId in hermes-management

### LINKED ISSUE #1418: Lack of fetching schema by schemaId in hermes-management
I will start with introducing my issue by quoting my post from gitter:

> Hello allegro/hermes team!
> We are using hermes as message broker in our event sourcing architecture. Aside we have spark batch job which is archiving avro messages directly from kafka. Here we have found out, that after schema evolution, it's impossible to deserialize avro message without knowing it's old schema(before evolution). Hermes messages stored on kafka have header in which we can find schemaId. Unfortunately hermes-management has only endpoint for fetching schema by topic and its schemaVersion. I know that I can just hit directly SchemaRegistry for this information, but in my case it's much more problematic.

(Thanks @druminski for encouraging me for making PR, but it has turned out that I need to discuss it before adding any implementation ;))

To resolve this I wanted to add endpoint to hermes-management for getting schema by schemaId.

I thought that schemaId is unique identifier through all schemas, so I will just create endpoint like GET `/schema/{schemaId}`, but it turned out that current codebase requires `topicName` on every step of gettting schema by schemaId (because of `SchemaRepositoryInstanceResolver`, which return schema provider by `topicName`).

There is already '/topics/{topicName}/schema/{version}', so I cannot create anything like `/topics/{topicName}/schema/{schemaId}`. 

(option A.) I thought about `/topics/{topicName}/schema?id={schemaId}`, but then I should return filtered list of schemas to be clean
(option B.) Add endpoint for getting `version`, like `/topics/{topicName}/schema/version/{schemaId}`, and than I will fetch schema with existing version endpoint

Do not hesitate to share your opinions and ideas,
cheers ;)

slomkarafa:


druminski (approved):
LGTM, thanks for the contribution :)

szczygiel-m (approved):

PR #1434, opened 2021-12-14
Add AVRO schema viewer

stanczykj:
A more human-readable viewer for AVRO schemas. Scrolling through JSON schema is not very convenient when you simply want to check if the event has a certain field or where that field is.

Here's how it looks. When the checkbox "Raw" is active, a json schema is displayed just like it used to.
image

I tried to make it able to handle all possible schemas. I've validated it against some real-life schemas and haven't noticed any problems.

I had to do minor changes (like add semicolons) to some parts of the existing code to make the tests pass.
I've also added a karma plugin to make html templates testable.

I'm open to any suggestions (especially regarding the look of it).


hermes-console/test/unit/AvroNodeControllerTest.js
```diff
@@ -0,0 +1,429 @@
...
+    var $scope;
+
+    beforeEach(module('hermes.topic.avroViewer'));
+
+    beforeEach(inject(function ($rootScope, _$controller_) {
+        $controller = _$controller_;
+        $scope = $rootScope;
+    }));
+
+    [
```
CamilYed, line 13: #minor
Does it support [aliases](https://avro.apache.org/docs/1.8.2/spec.html#Aliases)?

stanczykj: Nope, only the `name` field is displayed.

piotrrzysko (approved):

druminski (approved):

szczygiel-m (approved):

PR #1437, opened 2021-12-27
Replaced custom fetcher for offline-clients with iframe

szczygiel-m:


hermes-console/static/partials/topic.html
```diff
@@ -195,28 +195,9 @@ <h4>
...
-                        <td>{{client.owners.join(", ")}}</td>
-                    </tr>
-                </tbody>
-            </table>
-        </div>
-
-        <div ng-show="offlineClientsFetching" class="text-center">
-            <h1><i class="fa fa-circle-o-notch fa-spin"></i></h1>
+            <div class="list-group-item">
+                <iframe height="400" ng-src="{{trustSrc(iframeSource)}}" style="border: none;" width="100%"></iframe>
```
pitagoras3, line 199: What will happen when fetching iframe source will throw exception (for example `OfflineClientsServiceAbsentException`)? From code it looks like nothing here will be rendered. Can you add alert here in failing scenario?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/clients/OfflineClientsService.java
```diff
@@ -1,10 +1,5 @@
 package pl.allegro.tech.hermes.management.domain.clients;
 
-import pl.allegro.tech.hermes.api.TopicName;
-
-import java.util.List;
-
 public interface OfflineClientsService {
-
-    List<OfflineClient> find(TopicName topic);
+    String getIframeSource(String topic);
```
pitagoras3, line 4: Why you've changed `TopicName` parameter type to `String`?

szczygiel-m: Thats because i just wanted to avoid unnecessary `TopicName.fromQualifiedName(topic)` in controller, like it was before.

hermes-console/static/partials/topic.html
szczygiel-m: When there is absent `OfflineClientsService` user should keep `config.offlineClientsEnabled` property set to false, if so the entire part of ui with offline clients will not be rendered. In my opinion thats not necessary to add alert in case of failing request, also we don't do that either while checking blacklistedStatus or fetching subscriptions for topic.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/clients/OfflineClientsService.java
pitagoras3: I'm not sure if it is unnecessary - this method correctly verifies if topic is valid:
```java
public static TopicName fromQualifiedName(String qualifiedName) {
    if (Strings.isNullOrEmpty(qualifiedName)) {
        return null;
    }

    int index = qualifiedName.lastIndexOf(GROUP_SEPARATOR);
    if (index == -1) {
        throw new IllegalArgumentException("Missing group");
    }

    String groupName = qualifiedName.substring(0, index);
    String topicName = qualifiedName.substring(index + 1, qualifiedName.length());
    return new TopicName(groupName, topicName);
}
```

also, all methods in `TopicsEndpoint` use this validation. I would add creating `TopicName` in `OfflineClientsEndpoint` so it would validate topic correctness on API layer. For example, someone can call this `offline-clients-source` endpoint manually (with invalid topic name).

szczygiel-m: You're right, I fixed it

pitagoras3 (approved):

PR #1446, opened 2022-01-20
Branch for testing from fork

faderskd:


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/ExtraHeadersExtractor.java
```diff
@@ -0,0 +1,22 @@
...
+
+import io.undertow.server.HttpServerExchange;
+import pl.allegro.tech.hermes.frontend.publishing.metadata.HeadersPropagator;
+
+import javax.inject.Inject;
+import java.util.Map;
+
+import static pl.allegro.tech.hermes.frontend.publishing.metadata.HeadersToMapTransformer.toHeadersMap;
+
+public class ExtraHeadersExtractor {
```
faderskd, line 11: This class is used only in the context of logging... so maybe we should change the class name to be more accurate ? Just like `TracingHeadersExtractor` ?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/metadata/HeadersPropagator.java
```diff
@@ -6,4 +6,8 @@
 public interface HeadersPropagator {
 
     Map<String, String> extract(Map<String, String> headers);
+
+    default Map<String, String> extractHeadersToLog(Map<String, String> headers) {
```
faderskd, line 10: I'm not sure if you need to add this method to all propagators. Not all propagators work in the context of logging/tracing. Maybe It would be better to introduce some new interface dedicated for logging/tracking. It is hard to me to distinguish which headers are for logging and which are propagated to the clients.

hermes-benchmark/src/jmh/java/pl/allegro/tech/hermes/benchmark/environment/HermesServerFactory.java
```diff
@@ -54,8 +56,11 @@ static HermesServer provideHermesServer() throws IOException {
                 .overrideProperty(FRONTEND_TOPIC_METADATA_REFRESH_JOB_ENABLED, false)
                 .overrideProperty(FRONTEND_GRACEFUL_SHUTDOWN_ENABLED, false);
         Trackers trackers = new Trackers(Collections.emptyList());
+        TrackingHeadersExtractor trackingHeadersExtractor = new DefaultTrackingHeaderExtractor(headers -> headers);
```
szczygiel-m, line 59: just asking, why is there `headers -> headers` in constructor?

szczygiel-m (approved):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/MessageErrorProcessor.java
```diff
@@ -13,39 +13,46 @@
...
         this.objectMapper = objectMapper;
         this.trackers = trackers;
+        this.trackingHeadersExtractor = trackingHeadersExtractor;
     }
 
     public void sendAndLog(HttpServerExchange exchange, Topic topic, String messageId, ErrorDescription error) {
         sendQuietly(exchange, error, messageId, topic.getQualifiedName());
-        log(error.getMessage(), topic, messageId, readHostAndPort(exchange));
+        log(error.getMessage(), topic, messageId, readHostAndPort(exchange),
+                trackingHeadersExtractor.extractHeadersToLog(toHeadersMap(exchange.getRequestHeaders())));
```
piotrrzysko, line 42: What do you think about passing `HeaderMap` to `extractHeadersToLog` instead using `HeadersToMapTransformer.toHeadersMap` in every place where we call the `extractHeadersToLog` method?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/message/JsonMessage.java
```diff
@@ -1,19 +1,24 @@
...
 
+import java.util.Map;
+
 public class JsonMessage implements Message {
 
     private final String id;
     private final byte[] data;
     private final long timestamp;
     private final String partitionKey;
+    private final Map<String, String> extraRequestHeaders;
```
piotrrzysko, line 14: Probably it will not be needed if you apply my comment about the backup storage.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/avro/AvroMessage.java
```diff
@@ -14,17 +16,20 @@ public class AvroMessage implements Message {
     private final long timestamp;
     private final CompiledSchema<Schema> schema;
     private final String partitionKey;
+    private final Map<String, String> extraRequestHeaders;
```
piotrrzysko, line 19: Probably it will not be needed if you apply my comment about the backup storage.

hermes-benchmark/src/jmh/java/pl/allegro/tech/hermes/benchmark/environment/HermesServerFactory.java
faderskd: This class is used for extracting tracking headers. It relies on `HeadersPropagator` because it is using part of its functionality for filtering headers. This setup is only needed for hermes benchmarking, so we are giving here only simple implementation which just returns all the headers without any filtering.

piotrrzysko (commented):
An empty file is left: hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/di/FrontendBinder.java

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/MessageErrorProcessor.java
faderskd: Yes, I thought about it but I decided to do similar interface to the `HeadersPropagator` which operates on pure `Map<String, String>`. Yes, it looks like too much of duplication. I'll change it to operate on request `exchange`

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/TrackingHeadersExtractor.java
```diff
@@ -0,0 +1,11 @@
+package pl.allegro.tech.hermes.frontend.publishing.handlers.end;
+
+import io.undertow.util.HeaderMap;
+import java.util.Map;
+import pl.allegro.tech.hermes.frontend.publishing.metadata.HeadersToMapTransformer;
+
+public interface TrackingHeadersExtractor {
+    default Map<String, String> extractHeadersToLog(HeaderMap headers) {
+        return HeadersToMapTransformer.toHeadersMap(headers);
```
piotrrzysko, line 9: In fact, we probably do not need this default implementation as we already have `DefaultTrackingHeaderExtractor`. Without this implementation we could remove `HeadersToMapTransformer` and not change `MessageFactory`. 

Nonetheless, I'm ok with the current approach, so I leave the decision up to you.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/ExtraHeadersExtractor.java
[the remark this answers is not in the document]
faderskd: done

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/end/TrackingHeadersExtractor.java
faderskd: you're right, done

piotrrzysko (approved):

PR #1450, opened 2022-01-27
Create codeql-analysis.yml

bgalek:
With @apocalyptik we would like to test github code scanning ;)

### DIFF
--- a/codeql-analysis.yml
+++ b/codeql-analysis.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'java' ]
```
pitagoras3, line 35: In Hermes we also use `javascript` on frontend, and we have 2 `python` scripts :)

bgalek: good catch!

pitagoras3 (approved):

codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
```
pitagoras3, line 20: One more thing @bgalek - do we need to run this job on schedule? And if there is a reason for that - why we're doing this on 1:32 every Tuesday? 😄

bgalek: thats generated by github, I think thats a way for them to load balance ;)

PR #1458, opened 2022-02-07
During read-only mode admin can perform changes is hermes-console

pitagoras3:
Until now when _hermes-management_ was in `readOnly || readOnlyAdmin` mode, no one could edit topics, subscriptions or groups from _hermes-console_. Change introduced in this PR allows admin to perform __write__ operations using _hermes-console_ UI on topics, subscriptions and groups, even when _hermes-management_ is in read only mode.

### DIFF
--- a/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
+++ b/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
@@ -3,6 +3,7 @@ package pl.allegro.tech.hermes.management.domain.dc;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import pl.allegro.tech.hermes.common.exception.InternalProcessingException;
+import pl.allegro.tech.hermes.common.exception.RepositoryNotAvailableException;
 import pl.allegro.tech.hermes.management.domain.auth.RequestUser;
 
 import java.util.ArrayList;
@@ -22,7 +23,7 @@ public class MultiDatacenterRepositoryCommandExecutor {
     }
 
     public <T> void executeByUser(RepositoryCommand<T> command, RequestUser requestUser) {
-        if(requestUser.isAdmin()) execute(command, false, false);
+        if (requestUser.isAdmin()) execute(command, false, false);
         else execute(command);
     }
 
@@ -30,7 +31,7 @@ public class MultiDatacenterRepositoryCommandExecutor {
         execute(command, rollbackEnabled, true);
     }
 
-    private <T> void execute(RepositoryCommand<T> command, boolean isRollbackEnabled, boolean shouldFailOnAnyDcFailure) {
+    private <T> void execute(RepositoryCommand<T> command, boolean isRollbackEnabled, boolean shouldStopExecutionOnFailure) {
         if (isRollbackEnabled) {
             backup(command);
         }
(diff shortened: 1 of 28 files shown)


druminski (approved):

piotrrzysko:
For me this fix looks fine. However, if we really want to allow the admin to make changes in the read-only mode, in my opinion, there is still work to be done on the backend side. 

When the read-only mode is activated in most cases it means that at least one datacenter (more precisely ZooKeeper in one datacenter) is not available. Currently, in hermes-management there is a mechanism that tries to perform a given update in all datacenters. If the update fails in at least one of them, then a rollback procedure is applied. As far as I remember, the mechanism is not aware of the user role, so I suspect that in case of at least one datacenter failure, admins will still not be able to make changes in the healthy ones.

pitagoras3:
Thanks a lot @piotrrzysko, I will take a look into that!

pitagoras3:
This PR changed a little bit from one-liner 😄 Nonetheless, I've made it possible on backend site to execute operations regarding:
- Topics,
- Subscriptions,
- Blacklisting / unblacklisting,
- Groups,
- OAuth,
- Workload constraints,
- Moving offset
as admin without rollback (and without failure if there is problem with unavailable ZooKeeper).

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
```diff
@@ -20,8 +22,17 @@ public MultiDatacenterRepositoryCommandExecutor(RepositoryManager repositoryMana
         this.rollbackEnabled = rollbackEnabled;
     }
 
+    public <T> void executeByUser(RepositoryCommand<T> command, RequestUser requestUser) {
+        if (requestUser.isAdmin()) execute(command, false, false);
```
pitagoras3, line 26: Is it ok if admin's failed commands will never get rolled back? Even when _hermes-management_ is not in _read-only_ mode?

piotrrzysko: > Maybe we can also verify if hermes-management is in read-only mode here?

Yes, that would be nice. Otherwise, in a normal situation admins will not even notice that something went wrong.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
```diff
@@ -62,4 +71,17 @@ public MultiDatacenterRepositoryCommandExecutor(RepositoryManager repositoryMana
                     "' failed on DC '" + repoHolder.getDatacenterName() + "'.", e);
         }
     }
+
+    private <T> void handleExecutionRepositoryNotAvailableException(RepositoryCommand<T> command, boolean isRollbackEnabled, boolean shouldStopExecutionOnFailure, List<DatacenterBoundRepositoryHolder<T>> executedRepoHolders, DatacenterBoundRepositoryHolder<T> repoHolder, RepositoryNotAvailableException e) {
```
piotrrzysko, line 75: In my opinion, extracting these two methods (`handleExecutionRepositoryNotAvailableException` and `handleExecutionException`) doesn't improve readability as they have many arguments and some of them are booleans. Maybe it would be better to stay with the inline version. WDYT?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
pitagoras3: Added such verification:
```java
if (requestUser.isAdmin() && modeService.isReadOnlyEnabled()) execute(command, false, false);
```

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
pitagoras3: Ok, no problem - I've inlined them :)

piotrrzysko (approved):

PR #1465, opened 2022-02-22
Choosing owner from autocomplete source prevents manual entries

pitagoras3:
**Until now:**

User could either choose owner from dropdown, or type it manually. Typed in owners did not contained all required data.

**Solution:**

When owner source has _autocompletion_, hermes-console user has to choose one of proposed owners from dropdown list. User can type in owner manually, but without accepting concrete owner from dropdown, form will remain invalid.

out


szczygiel-m (approved):

druminski (approved):
Great UX improvement, thanks!

PR #1468, opened 2022-02-25
Explicitly releasing direct buffers in hermes-consumers

piotrrzysko:
By default native memory used by a given instance of `DirectByteBuffer` is released when the object is garbage collected. We don't want to wait for the GC and release the memory as soon as possible. Although the implementation may seem hacky, it seems to be the only possible way for now ([other projects approach this issue in a similar way](https://github.com/akka/akka/blob/331db9b8f8b02edb9370628d62ca53a55ed037e7/akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala#L80-L107)).

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/DirectBufferUtils.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/DirectBufferUtils.java
@@ -15,13 +15,13 @@ class DirectBufferUtils {
     static {
         DirectBufferCleaner cleaner;
         try {
-            cleaner = new Java11DirectBufferCleaner();
-        } catch (ReflectiveOperationException e) {
-            try {
+            if ("1.8".equals(System.getProperty("java.specification.version"))) {
                 cleaner = new Java8DirectBufferCleaner();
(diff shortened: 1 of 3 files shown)


druminski (approved):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/DirectBufferUtils.java
```diff
@@ -0,0 +1,78 @@
...
+
+class DirectBufferUtils {
+    private static final Logger logger = LoggerFactory.getLogger(DirectBufferUtils.class);
+
+    private static final DirectBufferCleaner CLEANER;
+
+    static {
+        DirectBufferCleaner cleaner;
+        try {
+            cleaner = new Java11DirectBufferCleaner();
```
pitagoras3, line 18: What do you think about checking Java version at runtime, and based on that result try to initialize proper cleaner?

piotrrzysko: done

pitagoras3 (approved):

PR #1476, opened 2022-03-14
Fixing flaky tests

szczygiel-m:


integration/src/integration/java/pl/allegro/tech/hermes/integration/PublishingTest.java
```diff
@@ -292,7 +294,7 @@ public void shouldPublishAndConsumeMessageWithTraceAndSpanHeaders() {
 
         // and
         Topic topic = operations.buildTopic(randomTopic("traceSendAndReceiveGroup", "topic2").build());
-        operations.createSubscription(topic, "subscription", HTTP_ENDPOINT_URL);
+        operations.createSubscription(topic, "subscription", remoteService.getUrl().toString());
```
druminski, line 297: Minor: consider overloading `operations.createSubscription()` in a way that the third argument will be URL class. Thanks to this you will not need to call `.toString()` in many test while creating subscription

integration/src/integration/java/pl/allegro/tech/hermes/integration/test/HttpResponseAssertion.java
```diff
@@ -20,6 +22,11 @@ public HttpResponseAssertion hasStatus(Response.Status status) {
         return this;
     }
 
+    public HttpResponseAssertion hasOneOfStatus(List<Response.Status> statuses) {
```
druminski, line 25: How about adding here `hasSuccesStatus()` and in it just check whether the response is from 2xx family? Thanks to this you don't have to pass arguments like ACCEPTED, CREATED during the call, which simplifies a little bit reading and writing tests

druminski (approved):
Great improvements. I've added two minor comments but they are not blocking, so +1 from me :)

faderskd (approved):

druminski (approved):

PR #1479, opened 2022-03-18
Refactor: kafka message conversion extracted from topic consuming logic

arkadius:
This PR extracts stateless logic of conversion between kafka `ConsumerRecord` to internal Hermes `Message` from `KafkaSingleThreadedMessageReceiver` which the only responsibility is now to handle stateful kafka topic consuming logic.

Thanks to that change will be easier to write tests for some not tested behaviour now like: how is used kafka consumer record envelope, how metadata nested inside content are translated to Message fields etc.

This change doesn't change anything in logic so it is no-risk change.

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
@@ -39,7 +39,7 @@ public class KafkaSingleThreadedMessageReceiver implements MessageReceiver {
     private static final Logger logger = LoggerFactory.getLogger(KafkaSingleThreadedMessageReceiver.class);
 
     private final KafkaConsumer<byte[], byte[]> consumer;
-    private final KafkaMessageConverter messageConverter;
+    private final KafkaConsumerRecordToMessageConverter messageConverter;
 
     private final BlockingQueue<Message> readQueue;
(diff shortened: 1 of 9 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -160,6 +135,7 @@ public void stop() {
     @Override
     public void update(Subscription newSubscription) {
         this.subscription = newSubscription;
+        messageConverter.update(subscription);
```
Diewa, line 138: Right now we have the same subscription object in two places. Therefore, we have to synchronize its state, which is what you're doing in this line. 

In my opinion it would be better to remove the subscription field from the `KafkaConsumerRecordToMessageConverter`class and instead of this pass subscription as an argument of `convertToMessage` metchod

arkadius: Take a look that it is consistent with how it is done in `KafkaSingleThreadedMessageReceiver`. `subscription` could  be passed to `next()` and other methods in order to avoid mutability of it, but I guess someone picked this approach to avoid complex methods signatures. I'm not convinced which one option is better. Just wanted to let you know that it was a kind of tradeoff - I also don't like mutability. Let me know which one direction I should go with.

Diewa (approved):

PR #1480, opened 2022-03-18
Refactor: Message to kafka ProducerRecord conversion logic extracted from KafkaBrokerMessageProducer

arkadius:
This PR is similar to #1479 but for frontend part. Motivation is also the same: to be able to test some corner cases without usage of integration test


szczygiel-m:
Please also update constructors in tests: 
image

arkadius:
Fixed - I've forgot how it can be tricky with dynamic languages ;) Can you approve workflows for two other of my PR so I will see if pipelines end up with success?

szczygiel-m (approved):
LGTM! 😄  thanks for your contribution 🎉

PR #1482, opened 2022-03-24
Added EndpointOwnershipValidator

szczygiel-m:
Added EndpointOwnershipValidator interface with empty implementation. Thanks to this we can create complex bean to create logic validating if provided endpoint and provided subscription owner match.


faderskd (approved):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/validator/SubscriptionValidator.java
```diff
@@ -19,23 +22,27 @@
     private final MessageFilterTypeValidator messageFilterTypeValidator;
     private final TopicService topicService;
     private final SubscriptionRepository subscriptionRepository;
+    private final Optional<EndpointOwnershipValidator> endpointOwnershipValidator;
```
piotrrzysko, line 25: To avoid playing around with `Optional`s you can add an empty implementation of `EndpointOwnershipValidator`.

szczygiel-m: I thought about it, but I came to the conclusion that it is better to use `Optional` in this case. We do not create additional classes and unnecessary code. Empty interface and empty implementation can be misleading

PR #1484, opened 2022-03-31
Added SubscriptionEndpointValidator

WojciechSova:
Added SubscriptionEndpointValidator interface. Now we will be able to check, for example, if endpoint have any forbidden phrase.

### DIFF
--- a/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/validator/SubscriptionValidator.java
+++ b/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/validator/SubscriptionValidator.java
@@ -23,6 +23,7 @@ public class SubscriptionValidator {
     private final TopicService topicService;
     private final SubscriptionRepository subscriptionRepository;
     private final Optional<EndpointOwnershipValidator> endpointOwnershipValidator;
+    private final Optional<SubscriptionEndpointValidator> subscriptionEndpointValidator;
 
     @Autowired
     public SubscriptionValidator(OwnerIdValidator ownerIdValidator,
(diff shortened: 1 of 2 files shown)


hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/validator/SubscriptionValidator.java
```diff
@@ -23,27 +23,27 @@
     private final TopicService topicService;
     private final SubscriptionRepository subscriptionRepository;
     private final Optional<EndpointOwnershipValidator> endpointOwnershipValidator;
+    private final Optional<SubscriptionEndpointValidator> subscriptionEndpointValidator;
```
piotrrzysko, line 26: We already have an endpoint validator: `pl.allegro.tech.hermes.api.constraints.EndpointAddressValidator`. Can we extend it instead of introducing another one? I think it will be clearer to have one place where this validation is done (especially considering the fact that in most cases the validator has to be aware of the endpoint protocol - `EndpointAddressValidator` already understands that we support different protocols).

piotrrzysko, line 26: We already have an endpoint validator: `pl.allegro.tech.hermes.api.constraints.EndpointAddressValidator`. Can we extend it instead of introducing another one? I think it will be clearer to have one place where this validation is done (especially considering the fact that in most cases the validator has to be aware of the endpoint protocol - `EndpointAddressValidator` already understands that we support different protocols).

piotrrzysko: My suggestion on how we can implemented this: https://github.com/allegro/hermes/commit/6be176a1d3b577fe6bffdcfa00556be8b375a597

PR #1508, opened 2022-05-23
Removing archaius from hermes

szczygiel-m:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/config/GooglePubSubSenderProperties.java
```diff
@@ -0,0 +1,67 @@
+package pl.allegro.tech.hermes.consumers.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "google.pubsub.sender")
```
piotrrzysko, line 5: In my opinion this section should be a child of the `consumer` section (`consumer.google.pubsub.sender`). It applies neither to `hermes-frontend` nor `hermes-management`.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/config/CommonConsumerProperties.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        public void setKillAfter(int killAfter) {
+            this.killAfter = killAfter;
+        }
+    }
+
+    public CommonConsumerParameters toCommonConsumerParameters() {
+        return new CommonConsumerParameters(
+                this.threadPoolSize,
```
faderskd, line 155: I know the comment I'm giving would cause big change but I think about such a constructor. It is an antipattern to have calls with so many parameters. It is very easy to call this constructor with exchanged parameters (the ones with the same type next to each other). Maybe we should add the another task to use builders in such a situations ?

piotrrzysko (approved):

piotrrzysko (approved):

PR #1509, opened 2022-05-23
move consumer receiver configuration to spring properties

Diewa:


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaMessageReceiverFactory.java
```diff
@@ -8,6 +8,7 @@
 import pl.allegro.tech.hermes.common.kafka.ConsumerGroupId;
 import pl.allegro.tech.hermes.common.kafka.KafkaNamesMapper;
 import pl.allegro.tech.hermes.common.metric.HermesMetrics;
+import pl.allegro.tech.hermes.consumers.config.ConsumerReceiverProperties;
```
piotrrzysko, line 11: @Diewa @szczygiel-m 

I wonder how we can avoid introducing circular dependencies between packages. Currently, the `config` package is aware of other packages (let's call them domain packages). If we put classes with properties in the `config` package, then the domain packages will become aware of it. 

What do you think about putting classes with properties in the domain packages (just plain classes, without spring annotations) and registering them as beans, like this:

```java
@Bean
@ConfigurationProperties(prefix = "consumer.receiver")
ConsumerReceiverProperties consumerReceiverProperties() {
    return new ConsumerReceiverProperties()
}
```

We can discuss this idea tomorrow.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/ConsumerReceiverParameters.java
```diff
@@ -0,0 +1,37 @@
+package pl.allegro.tech.hermes.consumers.consumer.receiver.kafka;
+
+public class ConsumerReceiverParameters {
```
piotrrzysko, line 3: I would call this class `KafkaReceiverParameters` as it is used only by `KafkaSingleThreadedMessageReceiver`.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/ConsumerReceiverParameters.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public int getPoolTimeout() {
+        return poolTimeout;
+    }
+
+    public int getReadQueueCapacity() {
+        return readQueueCapacity;
+    }
+
+    public boolean isWaitBetweenUnsuccessfulPolls() {
```
piotrrzysko, line 26: I would rename this method to `isWaitBetweenUnsuccessfulPollsEnabled`.

PR #1535, opened 2022-06-23
Fixed SubscriptionPartition equals and hashcode bug, and isCriticalEnvironment variable in management

szczygiel-m:


hermes-console/static/index.html
```diff
@@ -89,7 +89,7 @@
                     <a class="navbar-brand console-title" href="#/">
                         <span><img class="logo-mini" src="img/hermes-logo.png" alt="Hermes"/></span>
                         &nbsp;&nbsp; {{console.title}}
-                        <span ng-show="console.isCriticalEnvironment" class="alert alert-danger" style="padding: 10px"><strong>{{console.environmentName}}}</strong></span>
+                        <span ng-show="console.isCriticalEnvironment" class="alert alert-warning" style="padding: 10px"><strong>{{console.environmentName}}</strong></span>
```
piotrrzysko, line 92: Wouldn't be better to leave `alert-danger` here? The purpose of this box is to notify users that they are using Hermes in a critical environment, so red is probably a better fit.

piotrrzysko (approved):

PR #1541, opened 2022-07-14
Remove avro deserialization fallback ratelimit removed

faderskd:


faderskd:
I decided not to use ratelimiter, because during consumers startup they have to request schema-registry at the same time. We can treat it is as a peak of schema-registry requests. We have some built-in ratelimiter: every time the schema can't be fetched the empty message is returned from receiver. This in turn causes to throttle message receiving/converting so the throttle will concern the schema-registry requests too.

hermes-consumers/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
     }
 
     compile(group: 'com.google.cloud', name: 'google-cloud-pubsub', version: '1.115.1')
+    compile group: 'net.jodah', name: 'failsafe', version: versions.failsafe
```
piotrrzysko, line 34: Probably not needed.

piotrrzysko (approved):

PR #1548, opened 2022-08-17
Consumer load metrics for weighted work balancer

piotrrzysko:
This PR introduces consumer load metrics that will be used by the weighted work balancer. For now, I want to start with one metric: the number of operations per second that a given consumer performs. An operation is a unit of work that a consumer executes in the context of a subscription, such as receiving a message from Kafka, sending a message to the subscriber, or handling a signal.

PR #1549, opened 2022-08-18
Weighted work balancer

piotrrzysko:
Glossary:
* `consumer task` - a part of a subscription assigned to a given consumer. In other words this is one consumer from a consumer group. 
* `consumer task weight` - a vector describing how much resources a given consumer task requires. Currently, this is a single-element vector containing the `operations per second` metric introduced in #1548. 
* `consumer weight` - a sum of tasks' weights assigned to a consumer.
* `stabilization window` - the minimum time after assigning a subscription to consumers for which it will not be rebalanced.
* `subscription profile` - a record with information about a subscription necessary to make a rebalancing decision, i.e. weight, last rebalance timestamp.

Key goals of the weighted algorithm:
* It tries to ensure that all consumers are assigned the same number of tasks (this is exactly what the `selective` algorithm does). 
* It tries to ensure that all consumers have the same weight by swapping tasks between them.
* It tries to give recently balanced subscriptions a chance to catch up - see `stabilization window`.


faderskd (approved):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/weighted/WeightedWorkBalancer.java
```diff
@@ -0,0 +1,386 @@
...
+        for (ConsumerNode consumerNode : consumers) {
+            logger.debug("Consumer {} weight after rebalance: {}", consumerNode, consumerNode.getWeight());
+        }
+        return new AssignmentPlan(plan.getUnassignedTasks(), consumers);
+    }
+
+    private TargetConsumerLoad calculateTargetConsumerLoad(Collection<ConsumerNode> consumers) {
+        int totalNumberOfTasks = consumers.stream().mapToInt(ConsumerNode::getAssignedTaskCount).sum();
+        int consumerCount = consumers.size();
+        int maxNumberOfTasksPerConsumer = consumerCount == 0 ? 0 : (totalNumberOfTasks + consumerCount - 1) / consumerCount;
```
faderskd, line 153: why `(totalNumberOfTasks + consumerCount - 1)` what is e.g `totalNumberOfTasks = 11` and `consumerCount = 3` ==> `(totalNumberOfTasks + consumerCount - 1) / consumerCount = (11 + 2) / 3 = 4.3` but maxNumberOfTasksPerConsumer will be `4` so effectively will have some tasks unassigned ?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/weighted/WeightedWorkBalancer.java
```diff
@@ -0,0 +1,386 @@
...
+                .map(ConsumerNode::getWeight)
+                .reduce(Weight.ZERO, Weight::add);
+        return sum.divide(consumers.size());
+    }
+
+    private void tryMoveOutTasksFromOverloaded(ConsumerNode overloaded, ConsumerNode underloaded, TargetConsumerLoad targetLoad) {
+        List<ConsumerTask> candidatesToMoveOut = overloaded.getAssignedTasks().stream()
+                .filter(underloaded::isNotAssigned)
+                .collect(toList());
+        ListIterator<ConsumerTask> consumerTasIterator = candidatesToMoveOut.listIterator();
```
faderskd, line 172: `consumerTasIterator` -> `consumerTaskIterator`

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/workload/weighted/WeightedWorkBalancer.java
piotrrzysko: Good question! 

In the example you gave everything seems to work as expected: `maxNumberOfTasksPerConsumer` = `4` and we have `3` consumers, so there are `12` slots for tasks - we have one more slot than necessary. 

Anyway, I understand why you are confused here. What I wanted to achieve is simply rounding up. Java always rounds down when dividing two integers, but there is a formula that converts floors to ceilings (and I used it here): 
image
source: [Floor and ceiling functions](https://en.wikipedia.org/wiki/Floor_and_ceiling_functions)

I can change this to something more intuitive:
```java
(totalNumberOfTasks / consumerCount) + (totalNumberOfTasks % consumerCount > 0 ? 1 : 0)
```

WDYT?

faderskd: cool, it would be easier to understand while reading :)

piotrrzysko: done

faderskd (approved):

Kornelos (approved):

PR #1552, opened 2022-08-29
Fixed exceptions catching while closing consumer

szczygiel-m:
This PR is a bug fix. Regarding to the documentation `close()` method can also throw `InterruptException` and `KafkaException`. Prior to this fix, throwing these exceptions could interfere with code execution and for example shutdown of the ConsumerSupervisor 
image

### DIFF
--- a/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
+++ b/hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
@@ -153,9 +153,9 @@ public class KafkaSingleThreadedMessageReceiver implements MessageReceiver {
         } catch (IllegalStateException ex) {
             // means it was already closed
         } catch (InterruptException ex) {
-            logger.warn("InterruptException occurred during closing consumer: " + ex);
+            // means that the thread was interrupted
         } catch (KafkaException ex) {
-            logger.warn("KafkaException occurred during closing consumer: " + ex);
+            logger.warn("KafkaException occurred during closing consumer.", ex);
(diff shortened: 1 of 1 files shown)


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -152,6 +152,10 @@ public void stop() {
             consumer.close();
         } catch (IllegalStateException ex) {
             // means it was already closed
+        } catch (InterruptException ex) {
+            logger.warn("InterruptException occurred during closing consumer: " + ex);
+        } catch (KafkaException ex) {
+            logger.warn("KafkaException occurred during closing consumer: " + ex);
```
piotrrzysko, line 158: We probably want to log a stack trace. Therefore, a better option is to do it this way:
```java
logger.warn("KafkaException occurred during closing consumer.", ex);
```

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaSingleThreadedMessageReceiver.java
```diff
@@ -152,6 +152,10 @@ public void stop() {
             consumer.close();
         } catch (IllegalStateException ex) {
             // means it was already closed
+        } catch (InterruptException ex) {
+            logger.warn("InterruptException occurred during closing consumer: " + ex);
```
piotrrzysko, line 156: Do we need this log? This is a fairly standard situation in Java.

piotrrzysko (approved):

druminski (approved):

PR #1558, opened 2022-09-08
Pubsub message compression

michalferlinski:


adamallegro (approved):

remigiusz-kwiatkowski (dismissed):

wikp (dismissed):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/transformer/GooglePubSubMessageCompressionTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+        try {
+            byte[] data = message.getData();
+            if (data.length > compressionThresholdBytes) {
+                Message compressed = Message.message().fromMessage(message)
+                        .withData(compressor.compress(data))
+                        .build();
+                return super.fromHermesMessage(compressed);
+            }
+        } catch (IOException e) {
+            logger.debug("Error on PubSub message compression, switching to raw message processor", e);
```
piotrrzysko, line 41: I wonder about a couple of things here: 
- Why do we want to have a fallback to the `rawTransformer`? Maybe we should just propagate an error?
- Why do we log this error with the DEBUG level?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/GooglePubSubSenderTarget.java
```diff
@@ -31,38 +42,46 @@ public boolean equals(Object o) {
             return false;
         }
         GooglePubSubSenderTarget that = (GooglePubSubSenderTarget) o;
+
         return Objects.equals(topicName, that.topicName)
-                && Objects.equals(pubSubEndpoint, that.pubSubEndpoint);
+                && Objects.equals(pubSubEndpoint, that.pubSubEndpoint)
+                && Objects.equals(compressionCodec, that.compressionCodec);
```
piotrrzysko, line 48: With this change, we will have to create a  `GooglePubSubClient` for each tuple `<topicName, pubSubEndpoint, compressionCodec>` (earlier we need to do so for each `<topicName, pubSubEndpoint>`). AFAIR, we introduced the `GooglePubSubClientsPool` to reduce the footprint of creating `GooglePubSubClient`s, so I suspect that they are fairly expensive in terms of resource consumption.

Potentially, in our case, it will not cause any troubles, but still I wanted to ask two questions:
1. Is it intentional?
2. Have you considered optimizing it somehow?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/GooglePubSubSenderTarget.java
```diff
@@ -1,27 +1,38 @@
 package pl.allegro.tech.hermes.consumers.consumer.sender.googlepubsub;
 
 import com.google.pubsub.v1.TopicName;
+import pl.allegro.tech.hermes.consumers.consumer.sender.googlepubsub.transformer.compression.CompressionCodec;
 
 import java.util.Objects;
 
-class GooglePubSubSenderTarget {
+public class GooglePubSubSenderTarget {
```
piotrrzysko, line 8: So far, the `googlepubsub` package is relatively small in terms of the number of classes it has. Therefore, to avoid making this class public, I would suggest not to create additional subpackages (`transformer` and `compression`).

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/transformer/GooglePubSubMessageCompressionTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.pubsub.v1.PubsubMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import pl.allegro.tech.hermes.consumers.consumer.Message;
+import pl.allegro.tech.hermes.consumers.consumer.sender.googlepubsub.transformer.compression.GooglePubSubMessageCompressor;
+import pl.allegro.tech.hermes.consumers.consumer.trace.MetadataAppender;
+
+import java.io.IOException;
+
+public class GooglePubSubMessageCompressionTransformer extends GooglePubSubMessageTransformer {
```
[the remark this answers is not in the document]
michalferlinski: OK, the logic has been simplified by introducing three explicit `Transformers` for raw, compressed and optionally compressed (depending on message) size transmission.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/transformer/GooglePubSubMessageCompressionTransformer.java
michalferlinski: This IOException should get caused on OOM issues so wrapping it in `RuntimeException` and throwing up as it probably needs to fix memory usage.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/GooglePubSubSenderTarget.java
michalferlinski: Good point, to keep the pool as small as possible, the client assigned to a subscription endpoint will be used- no matters what compression codec is requested.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/GooglePubSubSenderTarget.java
michalferlinski: OK, moving classes to the `googlepubsub` and restricting visibility if possible.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/transformer/GooglePubSubMetadataCompressionAppender.java
```diff
@@ -0,0 +1,23 @@
+package pl.allegro.tech.hermes.consumers.consumer.sender.googlepubsub.transformer;
+
+import pl.allegro.tech.hermes.consumers.consumer.Message;
+import pl.allegro.tech.hermes.consumers.consumer.sender.googlepubsub.transformer.compression.CompressionCodec;
+
+import java.util.Map;
+
+public class GooglePubSubMetadataCompressionAppender extends GooglePubSubMetadataAppender {
```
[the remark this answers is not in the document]
michalferlinski: Indeed, the `MetadataAppender` does his only job inside the `MessageTransformer` but I see it valuable, e.g. the metadata part can be easier expanded or reduced according to message recipient needs. So leaving it separated for now.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/transformer/GooglePubSubMessageCompressionTransformer.java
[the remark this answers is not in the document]
piotrrzysko: Thanks. 

Can we go a step further and leave only one transformer for compression? In other words, I suggest merging `GooglePubSubMessageTransformerCompression` with `GooglePubSubMessageTransformerOptionalCompression`. My goal is not to introduce unnecessary complexity, but I'd be happy to know your take on this.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/GooglePubSubMessageTransformerCreator.java
```diff
@@ -0,0 +1,68 @@
...
+                            new GooglePubSubMessageTransformerCompression(
+                                    new GooglePubSubMetadataCompressionAppender(codec),
+                                    new MessageCompressor(codecFactory)))
+                    .map(compressionTransformer ->
+                            new GooglePubSubMessageTransformerOptionalCompression(
+                                    compressionThresholdBytes,
+                                    messageRawTransformer,
+                                    compressionTransformer));
+        } else {
+            return Optional.empty();
```
piotrrzysko, line 65: I'd suggest getting rid of `Optional` and returning `messageRawTransformer` here instead. If we do so, `GooglePubSubMessageTransformerProvider` will be obsolete.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/transformer/GooglePubSubMessageCompressionTransformer.java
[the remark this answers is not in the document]
michalferlinski: OK. The two `Compression` transformers merged into one with fallback to raw transmission by composition parameter.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/googlepubsub/GooglePubSubMessageTransformerCreator.java
michalferlinski: OK. Deleting `GooglePubSubMessageTransformerProvider` and moving its logic to `GooglePubSubMessageTransformerCreator`

piotrrzysko (approved):

druminski (approved):

PR #1560, opened 2022-09-12
Extracting messageId and timestamp from Kafka headers

druminski:
Extract of messageId and timestamp from Kafka headers instead of `__metadata` field. Thanks to this change, it will be easier to get rid of `__metadata` field from schema in the future.

Additionally, in one of our Hermes clusters (the one without Frontend part), we would like to set messageId and it will be better if we do this via Kafka headers instead of `__metadata` field.


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/config/KafkaHeaderNameProperties.java
```diff
@@ -9,6 +9,10 @@ public class KafkaHeaderNameProperties {
 
     private String schemaId = "sid";
 
+    private String messageId = "id";
```
szczygiel-m, line 12: Can You please move `KafkaHeaderNameParameters` from `frontend` to `common` package, and make `KafkaHeaderNameProperties` implementing [KafkaHeaderNameParameters](https://github.com/allegro/hermes/blob/ce5f351bded1d3ce15d5cc7319b1858baf47d3a3/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaHeaderNameParameters.java)? Just like for example [SchemaVersionRepositoryParameters](https://github.com/allegro/hermes/blob/ce5f351bded1d3ce15d5cc7319b1858baf47d3a3/hermes-common/src/main/java/pl/allegro/tech/hermes/common/schema/SchemaVersionRepositoryParameters.java)

druminski: Sure, good suggestion, done :)

szczygiel-m (approved):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaConsumerRecordToMessageConverter.java
```diff
@@ -17,30 +17,33 @@ public class KafkaConsumerRecordToMessageConverter {
...
         UnwrappedMessageContent unwrappedContent = messageContentReader.read(record, kafkaTopic.contentType());
         return new Message(
-                unwrappedContent.getMessageMetadata().getId(),
+                kafkaHeaderExtractor.extractMessageId(record.headers()),
                 topic.getQualifiedName(),
                 unwrappedContent.getContent(),
                 kafkaTopic.contentType(),
                 unwrappedContent.getSchema(),
-                unwrappedContent.getMessageMetadata().getTimestamp(),
+                kafkaHeaderExtractor.extractTimestamp(record.headers()),
```
arkadius, line 46: Why not to use `ConsumerRecord.timestamp` instead of dedicated header? Or at least be possible to use it (pass whole ConsumerRecord to this method)

druminski: Hey Arek, good point, I wasn't aware of this field, thanks! :)

On the frontend side we still have to set `ts` (timestamp) header for now as other our system reads it. However in the near future we will be able to remove the header as well.

szczygiel-m (approved):

arkadius (approved):
LGTM

PR #1567, opened 2022-10-05
Pubsub ability to not include additional headers

wikp:
Flag indicating whether additional attributes should be set in the Google Pub/Sub message or not

Google Pub/Sub pricing is based on the traffic. The more you send, the more you pay. Messages' attributes are also included in the size calculation. Therefore we need to control if we want to send all the attributes (i.e. additional headers and external metadata) or just required ones (schema version for avro messages, topic name, etc).

This pull-request introduces new configuration property for consumers: `consumer.google.pubsub.sender.include-more-attributes` (`true`/`false`) with the sensitive default being `false`.


hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/config/GooglePubSubSenderProperties.java
```diff
@@ -19,6 +19,8 @@ public class GooglePubSubSenderProperties {
 
     private String transportChannelProviderAddress = "integration";
 
+    private boolean includeMoreAttributes = false;
```
piotrrzysko, line 22: This name seems to be too general - without additional documentation or looking into code I can't tell what it does. Unfortunately, I can't find any short and neat name for that. Do we need to have this property? Maybe we can assume that for now a minimal set of headers is always enough. WDYT?

wikp: Agree. I'll change it in few minutes ;)

piotrrzysko (approved):

pitagoras3 (approved):

PR #1575, opened 2022-10-10
Fix fragile float based ScoringTargetWeightCalculatorTest

moscicky:
This PR makes [ScoringTargetWeightCalculatorTest](https://github.com/allegro/hermes/pull/1575/commits/534a5b789a1227f85b965c384a23fb883fe5acfc) robust to float comparison (test currently fails on M1 mac)


piotrrzysko (approved):
Thanks for the fix. Would it be possible to truncate the decimal parts a bit? I think two digits after the decimal point are enough:
```groovy
        def expected = [
                "c1": new Weight(132.07),
                "c2": new Weight(1.88),
                "c3": new Weight(66.03)
        ]
        areClose(expected, weights)
```

piotrrzysko (approved):

szczygiel-m (approved):

PR #1576, opened 2022-10-10
add ability to run integration tests on m1 mac

moscicky:
This PR enables developers to run integration tests on M1 macs, specifically:
- `testcontainers` and `toxiproxy` image were updated to allow running `toxiproxy` native arm64 image
- ability to specify confluent image tag via gradle task property was added, so developers can run tests with arm images:

```bash
 ./gradlew clean integrationTest -PconfluentImagesTag=7.2.2.arm64
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,7 +75,7 @@ allprojects {
             spring_web        : '5.1.5.RELEASE',
             failsafe          : '2.3.1',
             junit_jupiter     : '5.8.2',
-            testcontainers    : '1.15.3',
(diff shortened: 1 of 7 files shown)


WojciechSova:
Maybe we can use an image with version 7.2.2. This image is compiled for amd and arm. This will get rid of the parameter.
image

WojciechSova:
> 

In that case, what do you think about adding information about this parameter in the documentation?

WojciechSova (approved):

faderskd (approved):

Issue #1563, opened 2022-10-04
Add a linter to the project


piotrrzysko:
We would like to add a linter to the project. It should be plugged into [the GitHub action that we currently have](https://github.com/allegro/hermes/blob/master/.github/workflows/ci.yml). Also, it would be useful to be able to run it locally.  

We can give https://github.com/checkstyle/checkstyle + `Google Java Style Guide` a try.

piotrrzysko:
You're absolutely right, let's make the guide a bit more lenient :).

PR #1577, opened 2022-10-11
Java checkstyle (#1563)

kszapsza:


kszapsza:
I got some problem running integration tests. I need to investigate further, but maybe you guys have some suggestions. As far as I can see, TestContainer launch fails.

hermes-client/src/main/java/pl/allegro/tech/hermes/client/HermesClientBuilder.java
```diff
@@ -44,7 +44,7 @@ public HermesClient build() {
         return hermesClient;
     }
 
-    public HermesClientBuilder withURI(URI uri) {
+    public HermesClientBuilder withUri(URI uri) {
```
piotrrzysko, line 47: We need to be very careful with changing the hermes-client API. Usually, we try not to break backward compatibility to make the upgrade process easier for users. I would suggest turning off the check that validates that the method and class names are in the camel case. Or at least exclude the hermes-client module from this validation.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/batch/DirectBufferPool.java
```diff
@@ -14,7 +16,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package pl.allegro.tech.hermes.consumers.consumer.batch;
```
piotrrzysko, line 17: Can we exclude this file from the validation? The file has been copied from an external library, so it makes sens to keep the original formatting.

piotrrzysko (approved):

WojciechSova (approved):

Issue #1562, opened 2022-10-04
Add possibility to skip events


piotrrzysko:
Currently, there is tribal knowledge that if someone wants to skip events, they need to order retransmission from a future date.

Maybe it would be useful and easier for users to just have a button "Skip events" (or something like that) that under the hood does exactly the same what "retransmission from a future date" does.

sobelek:
Hey.
Can you explain to me what exactly does skipping events means?
Do I understand it correctly that if subscription have some lag than I can order retransmission from the future and events that were lagged will be discarded(skipped)?

piotrrzysko:
> if subscription have some lag than I can order retransmission from the future and events that were lagged will be discarded(skipped)

This is a correct explanation.

sobelek:
Great, I will tackle this soon.

PR #1579, opened 2022-10-14
Add option to skip events in hermes console

sobelek:
Hey. This PR addreses: #1562 

image

image


pitagoras3 (changes requested):
Hi @sobelek, thank you for your contribution!

I ran your PR locally, and I modified it a little bit. Here are my suggestions:
1. Merge `Retransmit` and `Skip messages` actions into one bootstrap panel. I highly suggest using [list groups](https://getbootstrap.com/docs/3.3/components/#panels-list-group).
2. Change color of the panel from `panel-danger` to `panel-default`.
3. Add tooltips showing action description when you hover the mouse over tooltip.
4. Move panel with retransmission and skipping messages to the left side (to make width of both sides more equal).

Final result could look like this:
Screenshot 2022-10-19 at 08 28 55

If you have any questions regarding implementation of those propositions, don't hesitate to ask questions :)

pitagoras3:
Hi @sobelek,

> Do you think we should get rid of this information in modal and instead show it in tooltip? Or maybe this information should be shown in both places?

I think that this info should be shown in both places, to make sure that we inform the user what he/she is doing :)

sobelek:
@pitagoras3 
Done :) Screenshot from current version:
image

pitagoras3 (approved):

piotrrzysko (approved):

Issue #1398, opened 2021-10-15
Hermes docker image should support Apple Silicon


AleksanderBrzozowski:
As a macbook M1 user I would like to be able to start hermes as a docker container.

When I try to do this I encounter an error, here is a log:

```
docker run allegro/hermes-management:hermes-1.9.3

...

2021-10-15 19:06:31.830  INFO 1 --- [           main] p.a.t.h.management.HermesManagement      : Starting HermesManagement using Java 11.0.11 on 16fb73a1f026 with PID 1 (/hermes-management/lib/hermes-management-1.9.3-hermes-1.9.3.jar started by root in /)
2021-10-15 19:06:31.839  INFO 1 --- [           main] p.a.t.h.management.HermesManagement      : The following profiles are active: local
2021-10-15 19:06:35.703  WARN 1 --- [           main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanDefinitionStoreException: Failed to process import candidates for configuration class [pl.allegro.tech.hermes.management.HermesManagement]; nested exception is java.io.FileNotFoundException: class path resource [org/springframework/boot/autoconfigure/web/client/RestTemplateAutoConfiguration.class] cannot be opened because it does not exist
2021-10-15 19:06:35.780  INFO 1 --- [           main] ConditionEvaluationReportLoggingListener :

Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
2021-10-15 19:06:35.848 ERROR 1 --- [           main] o.s.boot.SpringApplication               : Application run failed

org.springframework.beans.factory.BeanDefinitionStoreException: Failed to process import candidates for configuration class [pl.allegro.tech.hermes.management.HermesManagement]; nested exception is java.io.FileNotFoundException: class path resource [org/springframework/boot/autoconfigure/web/client/RestTemplateAutoConfiguration.class] cannot be opened because it does not exist
	at org.springframework.context.annotation.ConfigurationClassParser$DeferredImportSelectorGroupingHandler.lambda$processGroupImports$1(ConfigurationClassParser.java:822) ~[spring-context-5.3.3.jar:5.3.3]
	at java.base/java.util.ArrayList.forEach(Unknown Source) ~[na:na]
[...]

PR #1614, opened 2022-11-04
Docker-compose working on m1

dswiecki:
- using same gradle version in wrapper and in docker image - 6.9.3 image supports arm
- using gradle (not wrapper) for building inside docker image - using wrapper causes downloading gradle once again
- alpine-jre changed to jre which supports arm - until alpine-jre will support arm as well - jre is only 40MB larger
- swiched do apt

After that #1398 should be easily achievable with github buildx action


moscicky (approved):
Awesome, thanks :)

szczygiel-m (approved):

Issue #1619, opened 2022-11-14
Request timeout for batch subscriptions


moscicky:
Batch subscriptions are susceptible to slowly responding subscribes because the underlying http client (apache http client) does not provide timeout for the whole request, it only provides socketTimeout. This means that slowly responding subscribers can go over the limit of subscription request timeout. This can cause the consumer process to be deemed unhealthy by ConsumerProcessSupervisor. ConsumerProcessSupervisor then tries to interrupt the consumer process but the consumer process is blocked on reading from the socket and does not respect interrupts. This in turn causes signalQueue utilization be very high. 

We'd like to be able to set request timeout for batch subscriptions similarly as it is done in serial subscriptions (via jetty http client)

Issue #1011, opened 2019-04-10
Investigate and consolidate HTTP clients


cristaloleg:
Currently we're using 2 different HTTP clients in Hermes. It will be much easier for us to handle only 1 type.

druminski:
Serial consumer uses Jersey HTTP client but Batch consumer uses Apache HTTP Client. Serial has much higher priority in Hermes and it works fine so we can try to use also Jersey HTTP Client in Batch consumer. However we have to be careful with this as we don’t want to break Serial Consumer.

druminski:
Also we have to keep batch consumer timeouts configuration consistent with serial consumer.

haridevelops:
can i have a look in to it ?

cristaloleg:
@haridevelops sure

druminski:
This issue is still open and available.

piotrrzysko:
Resolved in #1620

PR #1620, opened 2022-11-16
Change http client for batch subscriptions

moscicky:
This PR resolves: https://github.com/allegro/hermes/issues/1619, https://github.com/allegro/hermes/issues/1011

There is a separate jetty http client for batch subscriptions, configurable via:

`consumer.batch.httpClient` - properties the same as for regular serial client. 

I have cherry picked 579226ce95bd0dacb9d574399b3231ca279701a9 to master and can confirm that the test included in that commit doesn't pass there (because apache supports only socket timeout)


piotrrzysko:
@moscicky Am I right that this will also solve #1011?

moscicky:
> @moscicky Am I right that this will also solve #1011?

@piotrrzysko You are right, added the issue to the PR description

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/config/ConsumerConfiguration.java
```diff
@@ -161,6 +161,11 @@ public HttpClient http1Client(HttpClientsFactory httpClientsFactory) {
         return httpClientsFactory.createClientForHttp1("jetty-http-client");
```
faderskd, line 161: Maybe we should change the previous client name to the `jetty-http-serial-client` to stay consistent ?

piotrrzysko:
It would be good to have the metrics reported by `HttpClientsWorkloadReporter` also for batches. These metrics are important as both the connection pool and requests queue are configured statically, so we need data to tune them correctly.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/DefaultSendingResultHandlers.java
```diff
@@ -19,7 +19,7 @@ public Response.CompleteListener handleSendingResultForBroadcast(CompletableFutu
     }
 
     @Override
-    public MessageSendingResult handleSendingResultForBatch(CloseableHttpResponse response) {
-        return MessageSendingResult.ofStatusCode(response.getStatusLine().getStatusCode());
+    public MessageSendingResult handleSendingResultForBatch(ContentResponse response) {
```
piotrrzysko, line 22: I wonder if it is feasible to make this API simpler - ideally, I would like to have one handler for all types of subscriptions. 

This is not strictly related to this PR. I'm just curious if we can somehow leverage the fact that now we use the same HTTP client everywhere. Plus, I'm not sure if we have the capacity for a bigger refactoring. Let's discuss this next week.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/sender/http/HttpClientsFactory.java
```diff
@@ -26,6 +26,10 @@ public HttpClientsFactory(HttpClientParameters httpClientParameters,
     }
 
     public HttpClient createClientForHttp1(String name) {
```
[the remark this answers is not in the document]
moscicky: https://github.com/allegro/hermes/pull/1620/commits/f6775979acc592a175f4f68a6ca2c1cd74112520 added batch client to workload metrics and refactored the configuration so it now explicit which client is configured by what properties:

For example:

```yaml
  http-client:
    monitoring:
      connectionPoolMonitoringEnabled: false
      requestQueueMonitoringEnabled: true
    serial:
      http1:
        threadPoolSize: 30
        threadPoolMonitoringEnabled: false
        followRedirectsEnabled: false
        maxConnectionsPerDestination: 100
        idleTimeout: 0ms
        maxRequestsQueuedPerDestination: 100
      http2:
        enabled: true
        threadPoolSize: 10
        threadPoolMonitoringEnabled: false
        idleTimeout: 0ms
        followRedirectsEnabled: false
        maxRequestsQueuedPerDestination: 100
    batch:
      http1:
        threadPoolSize: 30
        threadPoolMonitoringEnabled: false
        followRedirectsEnabled: false
        maxConnectionsPerDestination: 100
        idleTimeout: 0ms
        maxRequestsQueuedPerDestination: 100
```

oauth client is still using `http-client.serial.http1` but can be configured separately if its needed in the future

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/config/ConsumerConfiguration.java
moscicky: fixed

piotrrzysko (dismissed):

piotrrzysko (approved):

piotrrzysko (approved):

faderskd (approved):

PR #1636, opened 2022-12-08
Updated gradle to 6.9.3 and optimized build process

ajordanow:
## Build Optimisations

This PR provides set of optimisations that lead to faster build times and more robust local builds:

1. [Incremental builds](https://blog.gradle.org/introducing-incremental-build-support) are now supported (`hermes-consumers/build.gradle`, `hermes-management/build.gradle`)
2. [Parallel execution](https://docs.gradle.org/current/userguide/performance.html#parallel_execution) is enabled (`org.gradle.parallel=true` in `gradle.properties`)
3. Tests are now [executed in parallel](https://docs.gradle.org/current/userguide/performance.html#execute_tests_in_parallel) (`HermesSenderTest`, `ReactiveHermesSenderTest`, `IntegrationTest` and `ZookeeperBaseTest` had to be modified to use random ports - as reusing the same port numbers prevented the tests to be successful when running in parallel mode)
4. Gradle has been updated to `6.9.3` (latest `6.x` version)

### Benchmarks
_The bencharks were performed on Apple M1 Max machine._

#### Original build  

- `./gradlew clean assemble check` took about **3m 11s** ([build scan](https://scans.gradle.com/s/pzufaysr5emgs))
- subsequent `./gradlew clean assemble check` took about **20s** ([build scan](https://scans.gradle.com/s/sq2hwnsbyfvh4))

#### Optimized build

- `./gradlew clean assemble check` took about **1m 14s** ([build scan](https://scans.gradle.com/s/tlfp62kuk6ml6))
- subsequent `./gradlew clean assemble check` took about **1s** ([build scan](https://scans.gradle.com/s/znnha4svsuv2m))

### Build cache

The [build cache](https://docs.gradle.org/current/userguide/performance.html#enable_the_build_cache) was not enabled in the build scripts, but the builds can already benefit from it (e.g. when building locally):

```
# clear the cache
rm -rf ~/.gradle/caches/build-cache-1

# first build
./gradlew clean build --build-cache
...
BUILD SUCCESSFUL in 1m 16s
157 actionable tasks: 157 executed

# subsequent builds
./gradlew clean build --build-cache
...
BUILD SUCCESSFUL in 5s
157 actionable tasks: 78 executed, 79 from cache
```


hermes-management/build.gradle
```diff
@@ -44,21 +44,20 @@ dependencies {
...
 }
 
 task attachHermesConsole(type: Copy, dependsOn: 'buildHermesConsole') {
-    from '../hermes-console/dist/static'
-    into(sourceSets.main.output.resourcesDir.path + '/static')
-}
-
-tasks.register('prepareIndexTemplate') {
+    def srcDir = '../hermes-console/dist/static'
+    def targetDir = sourceSets.main.output.resourcesDir.path + '/static'
```
runningcode, line 56: we shouldn't write to the main source directory or copy files in to it.

instead we should output files to a task specific output directory in the `build` folder. then we should add this folder to the sourcesets (similar to line 82 of `hermes-consumers/build.gradle`)

runningcode (commented):
looks good, just one issue (which was already present in the build before)

build.gradle
```diff
@@ -187,6 +187,10 @@ subprojects {
         }
     }
 
+    tasks.withType(Test) {
```
runningcode, line 190: i would remove this unless there are some benchmarks that show that this improves performance

hermes-management/build.gradle
```diff
@@ -43,22 +43,29 @@ dependencies {
...
-                tofile: indexPath + '.ftl'
+        ant.move file: targetDir + '/index.html',
+            tofile: targetDir + '/index.html.ftl'
     }
 }
 
-jar.dependsOn(attachHermesConsole, 'prepareIndexTemplate')
+sourceSets {
+    main {
+        resources.srcDir consoleDir
```
runningcode, line 69: you can add the task provider here to automatically wire the task dependencies to ensure they are executed in the correct order 

```suggestion
        resources.srcDir tasks.named('attachHermesConsole')
```

runningcode (approved):
just some nits, but feel free to open the PR

build.gradle
ajordanow: Ok, maybe I'll leave this value set to `2` as there is a room for test execution parallelism - I've been able to shave off about 10 seconds from the build total time. Looking at the test code - there are actually projects with more than one "long-running" test cases. The benefit from parallelising the tests slightly outperforms the punishment for forking the test runners (that means - the total running time is better but there is now obvious proof of it looking at the build scans).

hermes-management/build.gradle
ajordanow: I think this will break the current behaviour of the build as the static files are not attached to the `distZip` package. This change would add a dependency that will cause static files to be included in the output.

hermes-management/build.gradle
```diff
@@ -43,22 +43,29 @@ dependencies {
...
 
 task attachHermesConsole(type: Copy, dependsOn: 'buildHermesConsole') {
-    from '../hermes-console/dist/static'
-    into(sourceSets.main.output.resourcesDir.path + '/static')
-}
-
-tasks.register('prepareIndexTemplate') {
+    def srcDir = '../hermes-console/dist/static'
+    def targetDir = consoleDir + '/static'
+    from srcDir
```
alextu, line 59: Small improvement, you should be able to replace this by `from tasks.named('buildHermesConsole')` and you can remove the `dependsOn: 'buildHermesConsole'`

hermes-management/build.gradle
runningcode: ah, i see. i suggested this with the assumption that all the outputs would be added.

Do we still would need to make sure that the `ProcessResources` task dependsOn `attachHermesConsole`?

build.gradle
runningcode: 2 sounds like a good plan!

ajordanow: oh, I was just changing something to re-trigger the build action. it's 2 now :)

hermes-management/build.gradle
ajordanow: I think not - `attachHermesConsole` is an auxilliary task that copies static files from other subproject. It's intended to be run manually when needed.

hermes-management/build.gradle
ajordanow: done. I also had to rework this task a little bit to have `dist` as a source (and `static` part included when copying)

szczygiel-m:
Hi @ajordanow, is this PR still relevant?

ajordanow:
> Hi @ajordanow, is this PR still relevant?

Should be (once the merge conflicts are resolved). But the checks were never green for this PR because at the time of the PR submission they constantly failed on `main`.

szczygiel-m:
Since then we have improved the stability of the tests, would you be willing to update this PR?

PR #1644, opened 2022-12-15
Undelivered messages are invisible to unauthorized users

Diewa:
Undelivered messages should only be visible to users with appropriate permissions (admin, theme owner, subscription owner)

### DIFF
--- a/hermes-console/static/js/console/subscription/SubscriptionController.js
+++ b/hermes-console/static/js/console/subscription/SubscriptionController.js
@@ -22,7 +22,6 @@ subscriptions.controller('SubscriptionController', ['SubscriptionRepository', 'S
         $scope.config = config;
 
         function getUndelivered() {
-            console.log("get Undelivered");
             subscriptionRepository.undelivered(topicName, subscriptionName).$promise
                 .then(function (undelivered) {
                     $scope.undelivered = undelivered;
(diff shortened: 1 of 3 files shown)


hermes-console/static/js/console/subscription/SubscriptionController.js
```diff
@@ -57,21 +78,16 @@ subscriptions.controller('SubscriptionController', ['SubscriptionRepository', 'S
...
 
-        subscriptionRepository.undelivered(topicName, subscriptionName).$promise
-            .then(function (undelivered) {
-                $scope.undelivered = undelivered;
-            })
-            .catch(function () {
-                $scope.undelivered = [];
-            });
+        getUndelivered();
+        getLastUndelivered();
```
piotrrzysko, line 90: Given that there is the above watch, do we need these calls?

piotrrzysko (approved):

hermes-console/static/js/console/subscription/SubscriptionController.js
Diewa: I added this watch to make sure the messages are fetched when the user status changes to authenticated. But you are right, and after changing the if condition, we no longer need these calls.

WojciechSova (approved):

piotrrzysko (approved):

PR #1648, opened 2022-12-21
Configurable consumer inflight queue size

moscicky:
This PR changes the way in which inflight queue size for consumers is calculated. 

Prior to this change inflight queue size was calculated based on 2 values:
- subscriptionInflightSize - set per subscription, default value: 100
- globalInflightSize - global property configured by   `consumer.serialConsumer.inflightSize`, default value: 100

Final queue inflight size was calculated as `min(subscriptionInflightSize, globalInflightSize)`.

After this PR default `subscriptionInflightSize` will be set to `null`, and final size will be calculated as:  

```java
subscriptionInflightSize != null ? subscriptionInflightSize : globalInflightSize
```
subscriptionInflightSize will be configurable only by admin users from now on.

### Migration

Deploying this version without any migration would cause inflight queue of subscriptions with `subscriptionInflightSize` greater than `globalInflightSize` to grow to `subscriptionInflightSize`. 

To prevent this the following migration is needed:
1. For subscriptions with `subscriptionInflightSize >= globalInflightSize`, set `subscriptionInflightSize=globalInflightSize`
2. Deploy hermes-management with new version
3.  Deploy hermes-consumers with new version
4. For subscriptions with `subscriptionInflightSize == globalInflightSize`, set `subscriptionInflightSize=null`

### Rollback
Changes are safe to rollback.

### DIFF
--- a/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
+++ b/hermes-api/src/main/java/pl/allegro/tech/hermes/api/SubscriptionPolicy.java
@@ -2,12 +2,15 @@ package pl.allegro.tech.hermes.api;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.google.common.base.MoreObjects;
+import pl.allegro.tech.hermes.api.constraints.AdminPermitted;
 import pl.allegro.tech.hermes.api.helpers.Patch;
 
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


piotrrzysko:
Please remember to add information to release notes about the need for migration (the order in which `hermes-consumers` and `hermes-managment` are deployed is important).

piotrrzysko (approved):

faderskd (approved):

PR #1650, opened 2023-01-18
Monitoring consumer groups

WojciechSova:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,180 @@
...
+            subscriptionsWithUnassignedPartitions1.add(subscriptionsWithUnassignedPartitions);
+        } else {
+            subscriptionsWithUnassignedPartitions2.add(subscriptionsWithUnassignedPartitions);
+        }
+    }
+
+    private List<List<TopicSubscription>> splitSubscriptions(List<TopicSubscription> topicSubscriptions) {
+        List<List<TopicSubscription>> splitSubscriptions = new ArrayList<>(Collections.emptyList());
+        splitSubscriptions.add(topicSubscriptions.subList(0, topicSubscriptions.size() / monitoringProperties.getNumberOfThreads()));
+        for (int i = 1; i < monitoringProperties.getNumberOfThreads(); i++) {
```
szczygiel-m, line 131: What do You think about using something common to split a list, for example [this](https://stackoverflow.com/questions/12026885/is-there-a-common-java-utility-to-break-a-list-into-batches) instead of developing our own algorithm for this?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,99 @@
...
+            return true;
+        }
+    }
+
+    private boolean checkIfAllPartitionsAreAssigned(ConsumerGroupId consumerGroupId,
+                                                       KafkaTopics kafkaTopics) throws ExecutionException, InterruptedException {
+        Optional<ConsumerGroupDescription> consumerGroupDescription = getConsumerGroupDescription(consumerGroupId);
+
+        if (isConsumerGroupRebalancing(consumerGroupDescription)) {
+            return true;
```
szczygiel-m, line 49: What do You think about adding logging to this case?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/config/MonitoringProperties.java
```diff
@@ -0,0 +1,35 @@
+package pl.allegro.tech.hermes.management.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "monitoring-consumer-groups")
+public class MonitoringProperties {
+
+    private Boolean enabled = false;
+    private Integer numberOfThreads;
+    private Integer secondsBetweenScans;
```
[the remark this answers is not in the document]
WojciechSova: Should I also provide default values, when they are not used without `enabled = true`?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,180 @@
...
+        Properties props = new Properties();
+        props.put(BOOTSTRAP_SERVERS_CONFIG, kafkaProperties.getBootstrapKafkaServer());
+        props.put(REQUEST_TIMEOUT_MS_CONFIG, kafkaProperties.getKafkaServerRequestTimeoutMillis());
+        props.put(SECURITY_PROTOCOL_CONFIG, DEFAULT_SECURITY_PROTOCOL);
+        if (kafkaProperties.getSasl().isEnabled()) {
+            props.put(SASL_MECHANISM, kafkaProperties.getSasl().getMechanism());
+            props.put(SECURITY_PROTOCOL_CONFIG, kafkaProperties.getSasl().getProtocol());
+            props.put(SASL_JAAS_CONFIG, kafkaProperties.getSasl().getJaasConfig());
+        }
+        return AdminClient.create(props);
```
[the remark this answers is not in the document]
WojciechSova: I can't use the same AdminClient in multiple threads, because this leads to errors when I try get information from Kafka in the same time.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
WojciechSova: Cool, thanks :D

docs/docs/configuration/monitoring.md
```diff
@@ -0,0 +1,15 @@
+# Monitoring
+## Ensure all topic partitions are assigned to consumer group in subscription
```
faderskd, line 2: Missing articles: `Ensure all topic partitions are assigned to a consumer group in a subscription`

docs/docs/configuration/monitoring.md
```diff
@@ -0,0 +1,15 @@
...
+
+To use this monitoring you have to enable it in file `application.yaml` in module `hermes-management`. 
+The subscriptions that have unassigned partitions are available through endpoint `/monitoring/consumergroup`.
+
+```yaml
+monitoringConsumerGroups:
+  enabled: false
```
faderskd, line 9: I guess it should be true if you would like to show the way how to enable it ;)

docs/docs/configuration/monitoring.md
```diff
@@ -0,0 +1,15 @@
...
+  numberOfThreads: 6
+  secondsBetweenScans: 120
```
faderskd, line 11: maybe just use Duration in the java properties file ? Then you can declare the value here as: `scanEvery: 120s` or `scanEvery: 5m`

hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/MonitoringEndpoint.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Autowired
+    public MonitoringEndpoint(MonitoringCache monitoringCache) {
+        this.monitoringCache = monitoringCache;
+    }
+
+    @GET
+    @Produces(APPLICATION_JSON)
+    @Path("/consumergroup")
+    public List<TopicSubscription> monitorConsumerGroup() {
```
faderskd, line 27: I think it should be `monitorConsumerGroups()`

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,180 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(MonitoringCache.class);
+
+    private final KafkaClustersProperties kafkaClustersProperties;
+    private final KafkaNamesMappers kafkaNamesMappers;
+    private final MonitoringProperties monitoringProperties;
+    private final SubscriptionService subscriptionService;
+    private final TopicService topicService;
+
+    private volatile Integer usingCache = 1;
```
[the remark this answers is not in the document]
faderskd: or if the variable has many possible values please use Enum type

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(MonitoringCache.class);
+
+    private final KafkaClustersProperties kafkaClustersProperties;
+    private final KafkaNamesMappers kafkaNamesMappers;
+    private final MonitoringProperties monitoringProperties;
+    private final SubscriptionService subscriptionService;
+    private final TopicService topicService;
+
+    private volatile List<List<TopicAndSubscription>> readySubscriptionsWithUnassignedPartitionsCache = new ArrayList<>();
```
faderskd, line 46: The volatile on collection modified by changing its content does have no effect. https://javarevisited.blogspot.com/2017/01/can-we-make-array-volatile-in-java.html#axzz7tZoA4Fz9
It makes sense to use volatile if you will change the reference

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,164 @@
...
+    private volatile List<List<TopicAndSubscription>> unreadySubscriptionsWithUnassignedPartitionsCache = new ArrayList<>();
```
faderskd, line 47: You propably don't need this variable at all to be at the object level. It is quite complicated now. Maybe better would be:
1. preserve `readySubscriptionsWithUnassignedPartitionsCache` as it is
2. remove `unreadySubscriptionsWithUnassignedPartitionsCache` from the object itself and move the creation of this list to the method responsible for fetching kafka consumer groups
3. once a fetching is done you just switch `readySubscriptionsWithUnassignedPartitionsCache` to the new collection created in point 2.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
faderskd, line 46: I think the better name would be `readOnlySubscriptionsWithUnassignedPartitionsCache `

WojciechSova: And also I don't need List of List, single List will be fine

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,100 @@
...
+        int topicPartitions = getTopicPartitions(kafkaTopics);
+        int partitionsInConsumerGroups = getPartitionsInConsumerGroups(consumerGroupDescription);
+
+        if (topicPartitions != partitionsInConsumerGroups) {
+            logger.warn("Monitoring. Cluster {}. Topic and subscription partitions do not match: {}, partitions in consumer groups: {}, "
+                            + "partitions in topic: {}", clusterName, consumerGroupId.asString(),
+                    partitionsInConsumerGroups, topicPartitions);
+            return false;
+        }
+        return Math.random() > 0.5;
```
WojciechSova, line 62: That was for testing purpose, will change to return true.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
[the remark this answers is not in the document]
faderskd: maybe we can create a `AdminClientFactory` to not duplicate this code from `KafkaConfiguration` ?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,100 @@
...
+        if (isConsumerGroupRebalancing(consumerGroupDescription)) {
+            logger.info("Monitoring. Consumer group is rebalancing: {}", consumerGroupId.asString());
+            return true;
+        }
+
+        int topicPartitions = getTopicPartitions(kafkaTopics);
+        int partitionsInConsumerGroups = getPartitionsInConsumerGroups(consumerGroupDescription);
+
+        if (topicPartitions != partitionsInConsumerGroups) {
+            logger.warn("Monitoring. Cluster {}. Topic and subscription partitions do not match: {}, partitions in consumer groups: {}, "
```
faderskd, line 57: Maybe more legible warning would be: `Missing consumer groups for cluster {} and subscription {}. Expected number of consumer groups: {}, but was {}`

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
faderskd, line 62: ?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,100 @@
...
+                            + "partitions in topic: {}", clusterName, consumerGroupId.asString(),
+                    partitionsInConsumerGroups, topicPartitions);
+            return false;
+        }
+        return Math.random() > 0.5;
+    }
+
+    private Optional<ConsumerGroupDescription> getConsumerGroupDescription(ConsumerGroupId consumerGroupId)
+            throws InterruptedException, ExecutionException {
+        Optional<ConsumerGroupDescription> consumerGroupDescription = adminClient
```
faderskd, line 67: return adminClient.describe...`

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,100 @@
...
+        KafkaFuture<TopicDescription> topicDescription = describeTopic(kafkaTopics.getPrimary().name().asString());
+        int topicPartitions = topicDescription.get().partitions().size();
+        if (kafkaTopics.getSecondary().isPresent()) {
+            topicPartitions += describeTopic(kafkaTopics.getSecondary().get().name().asString()).get().partitions().size();
+        }
+        return topicPartitions;
+    }
+
+    private static int getPartitionsInConsumerGroups(Optional<ConsumerGroupDescription> consumerGroupDescription) {
+        return consumerGroupDescription.get().members().stream()
```
faderskd, line 87: why do you use optional if you don't check if it is present ?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,100 @@
...
+                .mapToInt(member -> member.assignment().topicPartitions().size()).sum();
+    }
+
+    private boolean isConsumerGroupRebalancing(Optional<ConsumerGroupDescription> description) {
```
faderskd, line 91: same as above

hermes-api/src/main/java/pl/allegro/tech/hermes/api/TopicAndSubscription.java
```diff
@@ -0,0 +1,27 @@
...
+    private final Topic topic;
+    private final String subscription;
+
+    public TopicAndSubscription(Topic topic, String subscription) {
+        this.topic = topic;
+        this.subscription = subscription;
+    }
+
+    @JsonIgnore
+    public Topic getTopic() {
```
faderskd, line 16: do you need this method ? it is not used anywhere

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public List<TopicAndSubscription> getSubscriptionsWithUnassignedPartitions() {
+        return readOnlySubscriptionsWithUnassignedPartitionsCache;
+    }
+
+    private void monitorSubscriptionsPartitions() {
+        List<List<TopicAndSubscription>> splitSubscriptions = getSplitActiveSubscriptions();
+
+        ExecutorService executorService = Executors.newFixedThreadPool(monitoringProperties.getNumberOfThreads());
```
faderskd, line 65: creating threads is quite lengthy process, you can create the executor service in constructor

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,157 @@
...
+            List<Subscription> topicSubscriptions = subscriptionService.listSubscriptions(topicName);
+            subscriptions.addAll(createTopicSubscriptions(topicService.getTopicDetails(topicName),
+                    topicSubscriptions.stream().filter(subscription -> subscription.getState().equals(Subscription.State.ACTIVE))
+                            .map(Subscription::getName).collect(Collectors.toList())));
+        });
+        return subscriptions;
+    }
+
+    private List<TopicAndSubscription> createTopicSubscriptions(Topic topic, List<String> subscriptions) {
+        List<TopicAndSubscription> topicAndSubscriptions = new ArrayList<>();
```
faderskd, line 141: [minor] you can use shorter version too : 
```
return subscriptions.stream().map(it -> new TopicAndSubscription(topic, it)).collect(toList());
```

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private void monitorSubscriptionsPartitions() {
+        List<List<TopicAndSubscription>> splitSubscriptions = getSplitActiveSubscriptions();
+
+        ExecutorService executorService = Executors.newFixedThreadPool(monitoringProperties.getNumberOfThreads());
+
+        List<Future<List<TopicAndSubscription>>> futures = splitSubscriptions.stream().map(it -> executorService.submit(() -> {
+            try {
+                List<MonitoringService> monitoringServices = createMonitoringService();
```
faderskd, line 69: `createMonitoringServices`

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,157 @@
...
+            try {
+                return future.get();
+            } catch (InterruptedException | ExecutionException e) {
+                throw new RuntimeException(e);
+            }
+        }).flatMap(List::stream).collect(toList());
+    }
+
+    private static void waitForAllThreadsToEnd(List<Future<List<TopicAndSubscription>>> futures) {
+        while (futures.stream().anyMatch(future -> !future.isDone())) {
```
faderskd, line 96: The easiest way to wait for all futures is 
```
for (Future<?> f : futures) {
            f.get();
        }
```

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,157 @@
...
+        waitForAllThreadsToEnd(futures);
+        updateSubscriptionsWithUnassignedPartitionsCache(futures);
+    }
+
+    private void updateSubscriptionsWithUnassignedPartitionsCache(List<Future<List<TopicAndSubscription>>> futures) {
+        readOnlySubscriptionsWithUnassignedPartitionsCache = futures.stream().map(future -> {
+            try {
+                return future.get();
+            } catch (InterruptedException | ExecutionException e) {
+                throw new RuntimeException(e);
```
faderskd, line 90: Throw some more domain specific, custom exception. Just like `MonitoringCacheException"Error during monitoring subscriptions...", e)`

faderskd:
What about testing ? We should have at least integration test for the happy path. I think we can skip the test when there is not enough consumer groups because it can be hard to reproduce.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
WojciechSova: What about monitoring that has not finished in time? A new start of monitoring will cause errors as not all threads are available.

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,148 @@
...
+            } catch (Exception e) {
+                logger.error("Error in monitoring: ", e);
+                return new ArrayList<TopicAndSubscription>();
+            } finally {
+                logger.info("Monitoring ended");
+                executorService.shutdown();
+            }
+        })).collect(toList());
+
+        waitForAllThreadsToEnd(futures);
```
moscicky, line 81: how about creating a lists of tasks and the submiting them with `ExecutorService.invokeAll()`? This will automatically wait for all tasks to finish

moscicky (changes requested):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringService.java
```diff
@@ -0,0 +1,104 @@
...
+            return checkIfAllPartitionsAreAssigned(consumerGroupId, kafkaTopics);
+        } catch (Exception e) {
+            logger.error("Failed to describe group with id: {}", consumerGroupId.asString(), e);
+            return true;
+        }
+    }
+
+    private boolean checkIfAllPartitionsAreAssigned(ConsumerGroupId consumerGroupId,
+                                                       KafkaTopics kafkaTopics) throws ExecutionException, InterruptedException {
+        Optional<ConsumerGroupDescription> consumerGroupDescription = getConsumerGroupDescription(consumerGroupId);
```
moscicky, line 46: How about we batch requests to kafka?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,148 @@
...
+        updateSubscriptionsWithUnassignedPartitionsCache(subscriptionsWithUnassignedPartitions);
+    }
+
+    private List<TopicAndSubscription> getFreshListOfSubscriptionsWithUnassignedPartitions(
+                List<List<TopicAndSubscription>> splitSubscriptions) {
+        ExecutorService executorService = Executors.newFixedThreadPool(monitoringProperties.getNumberOfThreads());
+
+        List<Future<List<TopicAndSubscription>>> futures = splitSubscriptions.stream().map(it -> executorService.submit(() -> {
+            try {
+                List<MonitoringService> monitoringServices = monitoringServicesCreator.createMonitoringServices();
```
moscicky, line 69: Maybe we can create monitoring services only once and cache them in this class?

hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/monitoring/MonitoringCache.java
```diff
@@ -0,0 +1,148 @@
...
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    private List<List<TopicAndSubscription>> getSplitActiveSubscriptions() {
+        return splitSubscriptions(getAllActiveSubscriptions());
+    }
+
+    private List<List<TopicAndSubscription>> splitSubscriptions(List<TopicAndSubscription> topicAndSubscriptions) {
+        return Lists.partition(topicAndSubscriptions, (topicAndSubscriptions.size() / monitoringProperties.getNumberOfThreads()) + 1);
```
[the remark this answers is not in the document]
WojciechSova: From documentation:
`Returns consecutive sublists of a list, each of the same size (the final list may be smaller). For example, partitioning a list containing [a, b, c, d, e] with a partition size of 3 yields [[a, b, c], [d, e]] -- an outer list containing two inner lists of three and two elements, all in the original order.`
For example:
We have 7 subscriptions and 6 threads.
So the `subscription.size() / numberOfThreads = 7/6 = 1`. Keep in mind that the subscriptions and the numberOfThreads are integers. 
Now the `Lists.partition(subscription, 7/6=1)` will give us 7 partitions and we only have 6 threads. 
Now you can ask that `+1` will result in this example in only 4 partitions, but we have more than 7 subscriptions :D

PR #1668, opened 2023-05-04
Support for prometheus metrics - part II

faderskd:


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/metric/CachedTopic.java
```diff
@@ -45,40 +47,52 @@ public class CachedTopic {
...
 
-    public CachedTopic(Topic topic, HermesMetrics hermesMetrics, KafkaTopics kafkaTopics) {
-        this(topic, hermesMetrics, kafkaTopics, false);
+    public CachedTopic(Topic topic, HermesMetrics hermesMetrics, MicrometerHermesMetrics micrometerHermesMetrics,
+                       KafkaTopics kafkaTopics) {
+        this(topic, hermesMetrics, micrometerHermesMetrics, kafkaTopics, false);
     }
 
-    public CachedTopic(Topic topic, HermesMetrics hermesMetrics, KafkaTopics kafkaTopics, boolean blacklisted) {
+    public CachedTopic(Topic topic, HermesMetrics oldHermesMetrics, MicrometerHermesMetrics micrometerHermesMetrics,
```
piotrrzysko, line 55: Would it be possible/valuable to encapsulate which instrumentation library we use in `HermesMetrics`? We would still need to touch every place where `HermesMetrics` is used, but we would avoid replicating everywhere this pattern: `HermesTimer.from(micrometer, codahale`).

piotrrzysko: Thank you for the comprehensive answer! 

I really like the approach with dedicated methods for each metric, but I wonder how difficult and time-consuming it would be to migrate the entire codebase to it.

faderskd: I've created the draft implementation with metrics encapsulated behind common facade. I'm still thinking between two approaches:
1. The current with common approach
```
metricsFacade.topicsMetrics().brokerLatency()
```
Advantages: easy to find metrics (one place)
Disadvantages: harder testing - needs to always setup facade, breaks some OOP rules by getting internal fields of facade (minor), metrics from different packages must be in common package (as it is currently)

2. The approach without facade and injecting directly proper metric class e.g. `TopicsMetrics`:
```
topicMetrics.brokerLatency()
```

Advantages: easier testing, some metrics can only available in the package to which they correspond
Disadvantages: many metrics scattered in project, harder to mentally embrace what metrics we have

piotrrzysko: First of all, good job with the refactoring! I have to admit, though, that I was a bit surprised when I first saw `MetricsFacade` but I get your point about scattering metrics across the project. My biggest concern is having all metrics in the common module. Ideally, I'd get rid of this module (but that's a different story) :). The problem is that, as far as I know, there are places where some metrics are shared between the main modules. In particular, between `hermes-frontend`/`hermes-consumers` and `hermes-management`. If that weren't the case, we could consider having a facade per module: `HermesConsumersMetrics`, `HermesFrontendMetrics`. 

To wrap up, I like the current approach, and I think it cleans up a lot.

piotrrzysko (approved):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/metric/CachedTopic.java
faderskd: Yes, some of them are shared, some not. I think that despite of the flaws the approach give as consistent view what we have. It has some tradeoffs and if it turns out it has more cons that pros we can easily refactor it too.

piotrrzysko (approved):

piotrrzysko (approved):

druminski (approved):

PR #1674, opened 2023-05-25
CI jobs name independent of Java version

moscicky:
> Remove Java version from job name so that required PR checks are independent of java version. E.g. Required check will be renamed from `Java 11 check` to `unitTests`. This will enable build which use different java version (e.g. Java 17/21) to pass all checks. We don't plan to support multiple Java version in the foreseeable future.


pitagoras3 (approved):

moscicky:
> Remove Java version from job name so that required PR checks are independent of java version. E.g. Required check will be renamed from `Java 11 check` to `unitTests`. This will enable build which use different java version (e.g. Java 17/21) to pass all checks. We don't plan to support many Java version in the foreseeable future.

pitagoras3 (approved):

faderskd (approved):

PR #1678, opened 2023-06-12
bump spring boot to 2.7.7

moscicky:
Bump spring boot to 2.7.7.

Additionally, move checking whether `groups` path is created out of `StorageConfiguration` - there was circular dependency error after upgrade. See: https://github.com/spring-projects/spring-framework/issues/27876 for details.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperGroupRepository.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperGroupRepository.java
@@ -12,6 +12,7 @@ import pl.allegro.tech.hermes.domain.group.GroupNotEmptyException;
 import pl.allegro.tech.hermes.domain.group.GroupNotExistsException;
 import pl.allegro.tech.hermes.domain.group.GroupRepository;
 
+import javax.annotation.PostConstruct;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -107,5 +108,12 @@ public class ZookeeperGroupRepository extends ZookeeperBasedRepository implement
 
         String path = paths.groupPath(groupName);
         return readFrom(path, Group.class, quiet);
+    }
+
+    @PostConstruct
+    public void init() {
+        logger.info("Before ensuring init path exists");
+        ensurePathExists(paths.groupsPath());
+        logger.info("After ensuring init path exists");
     }
 }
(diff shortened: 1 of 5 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperGroupRepository.java
```diff
@@ -108,4 +110,11 @@ private Optional<Group> getGroupDetails(String groupName, boolean quiet) {
         String path = paths.groupPath(groupName);
         return readFrom(path, Group.class, quiet);
     }
+
+    @PostConstruct
+    public void init() {
+        logger.info("Before ensuring init path exists");
+        ensurePathExists(paths.groupsPath());
```
pitagoras3, line 117: **issue:** `ensurePathExists` which is used here is _"different"_ from the original one.
You're using method from `ZookeeperBasedRepository`, where previous code was using `ZookeeperClient` class method with same name.

I quickly looked on both implementations and they look similar:

***ZookeeperClient (previous)***
```java
public void ensurePathExists(String path) {
    try {
        if (curatorFramework.checkExists().forPath(path) == null) {
            logger.info("Creating path: {} in Zookeeper: {}", path, datacenterName);
            curatorFramework.create().creatingParentsIfNeeded().forPath(path);
        }
    } catch (Exception e) {
        logger.error("Error when creating path: {} in Zookeeper: {}", path, datacenterName, e);
        throw new InternalProcessingException("Could not ensure existence of path: " + path);
    }
}
```

***ZookeeperBasedRepository (current)***
```java
protected void ensurePathExists(String path) {
    ensureConnected();
    if (!pathExists(path)) {
        try {
            zookeeper.create().creatingParentsIfNeeded().forPath(path);
        } catch (Exception e) {
            throw new InternalProcessingException(e);
        }
    }
}
```

**suggestion:** Ensure that changing this method won't have negative effect, and if so - remove `ZookeeperClient.ensurePathExists` method, as it seems not used nowhere else.

pitagoras3 (changes requested):

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperGroupRepository.java
moscicky: Underneath they do the same thing: check if path exists using curator framework and create the path with parents if needed. 

If we were to comment out this check e2e tests would fail immediately:

```
Caused by: org.apache.zookeeper.KeeperException$NoNodeException: KeeperErrorCode = NoNode for /hermes/groups
```

Removed the `ZookeeperClient` version since its no longer used.

druminski (approved):

PR #1695, opened 2023-07-17
add prometheus support to frontend module

moscicky:
This PR adds support for prometheus metrics in 
- `hermes-frontend`

### Prometheus support status
- [x] hermes-api (no metrics)
- [x] hermes-benchmark (no metrics)
- [x] hermes-client (already supported)
- [ ] hermes-common 
- [ ] hermes-consumers
- [x] hermes-frontend (this PR)
- [ ] hermes-management
- [ ] hermers-metrics
- [x] hermes-mock (no metrics)
- [x] hermes-schema (no metrics)
- [x] hermes-test-helper (no metrics)
- [x] hermes-tracker 
- [x] hermes-tracker-elasticsearch


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/MetricsFacade.java
```diff
@@ -26,5 +30,13 @@ public SubscriptionMetrics subscriptionMetrics() {
     public ConsumerMetrics consumers() {
         return consumerMetrics;
     }
+
+    public PersistentBufferMetrics persistentBufferMetrics() {
```
piotrrzysko, line 34: Let's agree on a convention for naming these methods. Currently, some of them are named `somethingMetrics`, while others are simply named `something`. For the sake of conciseness, I'd vote for the latter pattern.

moscicky: Agree, lets stick to names without `Metrics` suffix

faderskd (dismissed):

hermes-metrics/src/main/java/pl/allegro/tech/hermes/metrics/HermesCounterWithRate.java
```diff
@@ -0,0 +1,4 @@
+package pl.allegro.tech.hermes.metrics;
+
+public interface HermesCounterWithRate extends HermesCounter, HermesRateMeter {
```
piotrrzysko, line 3: Do we really need to have a separate interface? AFAIK, Dropwizard's Meter also provides `count`. The only place where we benefit from having this distinction is the `published` counter, as we don't produce m1_rate for it. So, perhaps we can provide a dedicated implementation for this case (even as an anonymous class).

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/ProducerMetrics.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private static final String KAFKA_PRODUCER = "kafka-producer.";
+    private static final String ACK_LEADER = "ack-leader.";
+    private static final String ACK_ALL = "ack-all.";
+
+    private static final String ACK_ALL_BUFFER_TOTAL_BYTES = KAFKA_PRODUCER + ACK_ALL + "buffer-total-bytes";
+    private static final String ACK_ALL_BUFFER_AVAILABLE_BYTES = KAFKA_PRODUCER + ACK_ALL + "buffer-available-bytes";
+    private static final String ACK_ALL_METADATA_AGE = KAFKA_PRODUCER + ACK_ALL + "metadata-age";
+    private static final String ACK_ALL_RECORD_QUEUE_TIME_MAX = KAFKA_PRODUCER + ACK_ALL + "record-queue-time-max";
```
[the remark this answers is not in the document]
moscicky: They are, see: https://github.com/allegro/hermes/pull/1695/files/dd3c91fb565e3a1fe2bc44408d6b57a668d6e3f7#diff-e9fd10aab35b3d7b1bedf1d9f0249c2fe5a1f0cc874c4849041f5c522159b180R62

hermes-metrics/src/main/java/pl/allegro/tech/hermes/metrics/counters/DefaultHermesCounterWithRate.java
```diff
@@ -0,0 +1,30 @@
...
+    protected final io.micrometer.core.instrument.Counter micrometerCounter;
+    protected final com.codahale.metrics.Meter graphiteMeter;
+
+    protected DefaultHermesCounterWithRate(io.micrometer.core.instrument.Counter micrometerCounter,
+                                    com.codahale.metrics.Meter graphiteMeter) {
+        this.micrometerCounter = micrometerCounter;
+        this.graphiteMeter = graphiteMeter;
+    }
+
+    public static DefaultHermesCounterWithRate from(io.micrometer.core.instrument.Counter micrometerCounter,
```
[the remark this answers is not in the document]
moscicky: removed

hermes-metrics/src/main/java/pl/allegro/tech/hermes/metrics/HermesCounterWithRate.java
moscicky: We can drop the interface by directly exposing concrete implementations of `HermesCounters`: `DefaultHermesCounter`, `MeterBackedHermesCounter` from `HermesCounter` - please see if thats what you expected: https://github.com/allegro/hermes/pull/1695/commits/ce935d08596e1a14768134f44534a6a5a08a87ec I don't think its possible to provide the implementation as anonymous class, can you perhaps write a snippet of what you had in mind?

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/ProducerMetrics.java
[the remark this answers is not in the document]
piotrrzysko: Oh, right, I overlooked it. However, I noticed that sometimes you use `milliseconds` as a unit, which is not in line with what the [documentation on best practices](https://prometheus.io/docs/practices/naming/) says:

> ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers).

Yesterday, I discovered this API: 
```java
meterRegistry.more().timeGauge("consumers-rate.max-rate.coordinator.duration", List.of(), obj, TimeUnit.MILLISECONDS, f);
```

I used it in consumers, but I haven't tested it yet. I'm going to do so today.

hermes-metrics/src/main/java/pl/allegro/tech/hermes/metrics/HermesCounterWithRate.java
piotrrzysko: My idea was as follows: 

Let's have the `HermesCounter` interface: 

```java
public interface HermesCounter {
   
    void increment(long size);

    default void increment() {
        increment(1L);
    }
    
    double getOneMinuteRate();
}
```

and two classes implementing it (forget about my suggestion regarding the anonymous class):

```java
public class DefaultHermesCounterWithRate implements HermesCounter {
    private final io.micrometer.core.instrument.Counter micrometerCounter;
    private final com.codahale.metrics.Meter graphiteMeter;

    DefaultHermesCounterWithRate(io.micrometer.core.instrument.Counter micrometerCounter,
                                    com.codahale.metrics.Meter graphiteMeter) {
        this.micrometerCounter = micrometerCounter;
        this.graphiteMeter = graphiteMeter;
    }
    
    ...

    public void increment(long size) {
        this.micrometerCounter.increment((double) size);
        this.graphiteMeter.mark(size);
    }

    @Override
    public double getOneMinuteRate() {
        return graphiteMeter.getOneMinuteRate();
    }
}
```

```java
public class DefaultHermesCounter implements HermesCounter {
    private final io.micrometer.core.instrument.Counter micrometerCounter;
    private final com.codahale.metrics.Counter graphiteCounter;

    DefaultHermesCounter(io.micrometer.core.instrument.Counter micrometerCounter,
                                   com.codahale.metrics.Counter graphiteCounter) {
        this.micrometerCounter = micrometerCounter;
        this.graphiteCounter = graphiteCounter;
    }
   
    ...

    public void increment(long size) {
        this.micrometerCounter.increment((double) size);
        this.graphiteCounter.inc(size);
    }

    @Override
    public double getOneMinuteRate() {
        // I forgot about this method :(. Can we get the rate from the micrometer counter? We will need to do that in the future either way.
    }
}
```

As you see, my idea has one flaw (see the comment in the `DefaultHermesCounter#getOneMinuteRate` method). Considering that we will get rid of `HermesCounterWithRate` after the migration (is my assumption correct?), there is no point in providing a fake implementation of `getOneMinuteRate` in `DefaultHermesCounter`.

hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/ProducerMetrics.java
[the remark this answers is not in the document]
moscicky: I thought this best practice does not apply to Gauges (because we would need to scale them accordingly), but the API you mentioned is exactly what we needed! [Docs](https://micrometer.io/docs/concepts#_timegauge) say that the metric will automatically be converted to seconds and that the name suffix will be added, very cool

hermes-metrics/src/main/java/pl/allegro/tech/hermes/metrics/HermesCounterWithRate.java
moscicky: I don't think there is an easy way to extract `m1_rate` client side from micrometer (I couldn't find a built-in option to do that). So lets assume that calculating `oneMinuteRate` based on `micrometer#Counter` requires querying metric backend or some fancy client side aggregation and is out of scope of this PR. 

What in that case? _Implementing_ `getOneMinuteRate` in `DefaultHermesCounter` with something like 

```java
@Override
    public double getOneMinuteRate() {
        throw UnimplementedException()
    }
```

seems dangerous as we loose compile-time safety and someone _could_ use it by mistake. That's why I separated `HermesCounter` and `HermesCounterWithRate` originally. What do you think?

piotrrzysko: By `fake implementation` I meant something like: 

```java
    @Override
    public double getOneMinuteRate() {
        return 0d;
    }
```

or as you suggested: 
```java
    @Override
    public double getOneMinuteRate() {
        throw new UnimplementedException()
    }
```

> I don't think there is an easy way to extract m1_rate client side from micrometer 

So, how are we going to implement `getOneMinuteRate` after removing dropwizard?

moscicky: > So, how are we going to implement getOneMinuteRate after removing dropwizard? 

Perhaps @faderskd had an idea as he originally introduced implementations of `HermesRateMeter` which were solely based on dropwizard.  I'd say the options are:
1. querying the metric backend periodically
2. implementing rate metering on application side

Anyway, I thought this was purposefully left to deal with later on so I didn't want to do it in this PR

piotrrzysko: OK, let's leave it as it is now.

piotrrzysko (approved):

PR #1703, opened 2023-08-10
Hermes console vue stats

moscicky:


hermes-console-vue/json-server/server.js
```diff
@@ -0,0 +1,27 @@
+const topics = require('./topics.json');
```
Ejden, line 1: Change this file to typescript file. No js allowed :D

hermes-console-vue/json-server/server.js
```diff
@@ -0,0 +1,27 @@
...
+const subscriptions = require('./subscriptions.json');
```
Ejden, line 2: When you will migrate to typescript use imports instead of 'require' function

hermes-console-vue/json-server/server.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+server.post('/query/subscriptions', (req, res) => {
+  res.jsonp(subscriptions);
+});
+
+const router = jsonServer.router('json-server/db.json');
+server.use(router);
+
+server.listen(3000, () => {
```
Ejden, line 25: Have you checked that all endpoints from db.json works like we expect them to work after defining server on our own?

hermes-console-vue/json-server/server.js
```diff
@@ -0,0 +1,27 @@
...
+  console.log('JSON Server is running');
+});
```
Ejden, line 27: As i'm looking with Mateusz at this endpoint, I think it would be best to simply add a new endpoint in hermes-management to return necessary data, instead of returning huge json with all topics, and counting them on frontend. What do you think? Just two of those endpoints that returns topics and subscriptions are returning 12MB (sic!) of data.

hermes-console-vue/src/api/hermes-client/index.ts
```diff
@@ -60,3 +61,17 @@ export function fetchOfflineClientsSource(
     `/topics/${topicName}/offline-clients-source`,
   );
 }
+
+export function queryTopics(queryJSON: object): ResponsePromise<Topic[]> {
+  return axios.post<Topic[]>(`/query/topics`, queryJSON, {
+    headers: { 'Content-Type': 'application/json' },
```
Ejden, line 67: I think we can extract this header and place it at the top of this file. Something like this:
```typescript
const contentTypeHeader = 'Content-Type'
const jsonMediaType = 'application/json'
```

hermes-console-vue/src/composables/use-stats/useStats.ts
```diff
@@ -0,0 +1,184 @@
...
+  fetchTopics: Error | null;
+  fetchAckAllTopics: Error | null;
+  fetchTrackingEnabledTopics: Error | null;
+  fetchAvroTopics: Error | null;
+  fetchSubscriptions: Error | null;
+  fetchTrackingEnabledSubscriptions: Error | null;
+  fetchAvroSubscriptions: Error | null;
+}
+
+export function useStats(): UseStats {
```
Ejden, line 32: Looks pretty good, probably could be simplified after adding dedicated endpoint for statistics in management api

hermes-console-vue/src/views/admin/stats/StatsView.vue
```diff
@@ -0,0 +1,111 @@
...
+  } = useStats();
+</script>
+
+<template>
+  <v-container>
+    <v-row class="home__logo" justify="center" cols="12">
+      <v-col md="12">
+        <loading-spinner v-if="loading" />
+        <console-alert
+          v-if="error"
```
Ejden, line 31: Hmm, i'm not sure how this works, but error is a structure of errors so checking if error is not null could lead to returning always true (?) as error structure is not null. I've checked, and it seems like it's always true when all the errors are nulls. Code below logs 'Error occurrence'
```typescript
interface SomeErrors {
    fetchError: Error | null;
    updateError: Error | null;
}

const errors = {
    fetchError: null,
    updateError: null,
} as SomeErrors;

if (errors) {
    console.log('Error occurrence');
} else {
    console.log('No error detected');
}
```

hermes-console-vue/src/views/admin/stats/StatsView.vue
```diff
@@ -0,0 +1,111 @@
...
+          <h2>{{ subscriptionCount }}</h2>
+        </v-row>
+      </v-col>
+      <v-col md="3">
+        <v-row>
+          <h2>{{ $t('stats.trackingEnabled') }}</h2>
+        </v-row>
+        <v-row>
+          <h2>
+            {{ trackingEnabledSubscriptionCount }} ({{
```
Ejden, line 95: When displaying numbers use our util function formatNumber()

hermes-console-vue/json-server/server.js
moscicky: yes, they do

hermes-console-vue/json-server/server.js
moscicky: Yeah I agree, I was also surprised by how its implemented right now

hermes-console-vue/src/views/admin/stats/StatsView.vue
moscicky: done

hermes-console-vue/src/views/admin/stats/StatsView.vue
moscicky: changed to single error

hermes-console-vue/src/api/hermes-client/index.ts
moscicky: not relevant anymore

hermes-console-vue/json-server/server.js
moscicky: server.js was removed

hermes-console-vue/json-server/server.js
moscicky: server.js was removed

hermes-console-vue/src/composables/use-stats/useStats.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+import axios from 'axios';
+import type { Mocked } from 'vitest';
+
+vitest.mock('axios');
+const mockedAxios = axios as Mocked<typeof axios>;
+
+describe('useStats', () => {
+  it('should hit expected Hermes API endpoint', async () => {
+    // given
+    mockedAxios.get.mockResolvedValueOnce({ data: statsResponse });
```
Ejden, line 14: Instead of mocking axios, use mocked server from msw library. U can look at useTopic.spec.ts how it should be done

hermes-console-vue/src/composables/use-stats/useStats.ts
```diff
@@ -0,0 +1,92 @@
...
+  trackingEnabledSubscriptionShare: Ref<number | undefined>;
+  avroSubscriptionCount: Ref<number | undefined>;
+  avroSubscriptionShare: Ref<number | undefined>;
+  loading: Ref<boolean>;
+  error: Ref<boolean>;
+}
+
+export function useStats(): UseStats {
+  const loading = ref(false);
+  const topicCount = ref<number>();
```
Ejden, line 24: You can define an interface that will contain all those statistics. This would reduce all those refs to just one ref named statistics, that will be returned from this composable

hermes-console-vue/src/router/index.ts
```diff
@@ -49,6 +49,11 @@ const router = createRouter({
       component: () =>
         import('@/views/admin/consumer-groups/ConsumerGroupsView.vue'),
     },
+    {
+      path: '/stats/',
+      name: 'stats',
+      component: () => import('../views/stats/StatsView.vue'),
+    },
```
Ejden, line 56: Use @ sign instead of ..

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Stats.java
```diff
@@ -0,0 +1,46 @@
+package pl.allegro.tech.hermes.api;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+public class Stats {
+    private final TopicStats topicStats;
```
Ejden, line 9: What do you think about using @Nonnull property? Do you use it? Maybe you (as a team) should consider using it

Ejden (changes requested):

hermes-console-vue/src/composables/use-stats/useStats.spec.ts
moscicky: fixed

hermes-console-vue/src/composables/use-stats/useStats.ts
```diff
@@ -0,0 +1,92 @@
...
+  trackingEnabledTopicShare: Ref<number | undefined>;
+  avroTopicCount: Ref<number | undefined>;
+  avroTopicShare: Ref<number | undefined>;
+  subscriptionCount: Ref<number | undefined>;
+  trackingEnabledSubscriptionCount: Ref<number | undefined>;
+  trackingEnabledSubscriptionShare: Ref<number | undefined>;
+  avroSubscriptionCount: Ref<number | undefined>;
+  avroSubscriptionShare: Ref<number | undefined>;
+  loading: Ref<boolean>;
+  error: Ref<boolean>;
```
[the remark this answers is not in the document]
moscicky: fixed

hermes-console-vue/src/composables/use-stats/useStats.ts
moscicky: fixed

hermes-console-vue/src/router/index.ts
moscicky: fixed

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Stats.java
moscicky: There is no such convention right now so I'd rather not change that in this PR. However this something to consider in the future

hermes-console-vue/src/views/stats/StatsView.vue
```diff
@@ -0,0 +1,115 @@
...
+      </v-col>
+    </v-row>
+    <div v-if="stats">
+      <v-row class="mt-16">
+        <h1>{{ $t('stats.topics') }}</h1>
+      </v-row>
+      <v-row>
+        <v-col md="3">
+          <v-row>
+            <h2>{{ $t('stats.total') }}</h2>
```
Ejden, line 33: One last little thing. We can use headings classes from vuetify typography: https://vuetifyjs.com/en/styles/text-and-typography/. This way we would have consistent headings accros our app. Other than that: Looks pretty good! Congrats on your first vue 3 PR :D

moscicky: fixed! and thanks :)

Ejden (approved):

PR #1714, opened 2023-08-24
extract notification store

moscicky:


hermes-console-vue/src/store/app-notifications/useAppNotifications.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useNotificationsStore = defineStore('notifications', {
+  state: (): NotificationsState => {
+    return {
+      notifications: [],
+    };
+  },
+  actions: {
+    async dispatchNotification({ title, text, type }: NotificationConfig) {
+      const notification: Notification = {
+        id: generateUUID(),
```
szczygiel-m, line 18: Minor: what about using window.crypto.randomUUID() instead? no need for another dependency

szczygiel-m (approved):

hermes-console-vue/src/store/app-notifications/useAppNotifications.ts
Ejden: But then notification store would not work on http domains

Ejden (approved):

PR #1718, opened 2023-08-30
Feature/favorite entities

szczygiel-m:


hermes-console-vue/src/store/favorites/useFavorites.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+    async removeSubscriptionByQualifiedName(subscriptionQualifiedName: string) {
+      this.subscriptions = this.subscriptions.filter(
+        (subscriptionName) => subscriptionName !== subscriptionQualifiedName,
+      );
+    },
+  },
+  getters: {
+    getTopics(state: FavoritesState): string[] {
+      return state.topics?.sort((a, b) => a.localeCompare(b));
```
moscicky, line 36: suggestion: if the order in which favourites were added is not important (they are sorted lexicographically here), we can use `Set` instead of array. This would also prevent duplicated record if there were some weird edge cases

hermes-console-vue/src/views/favorite/topics/topic-listing/FavoriteTopicsListing.vue
```diff
@@ -0,0 +1,87 @@
...
+          <th></th>
+        </tr>
+      </thead>
+      <tbody v-if="filteredTopics.length > 0">
+        <tr
+          v-for="(topic, index) in filteredTopics"
+          :key="topic"
+          class="topics-table__row"
+        >
+          <td class="text-medium-emphasis" @click="onTopicClick(topic)">
```
moscicky, line 47: question: is there a way to make the whole row trigger `onTopicClick` except the area of `favourites icon` which should trigger `favorites. removeTopic` (so that we don't have to use `onTopicClick` for each cell)?

hermes-console-vue/src/views/subscription/subscription-metadata/SubscriptionMetadata.vue
```diff
@@ -20,6 +25,14 @@
     suspend: [];
     activate: [];
   }>();
+
+  const copyToClipboard = (content: string) => {
```
moscicky, line 29: suggestion: perhaps we can move all of this to `copy-utils` as well?

moscicky (approved):
Much needed feature, great work 👍🏻

hermes-console-vue/src/views/favorite/topics/topic-listing/FavoriteTopicsListing.vue
szczygiel-m: I tried to do it for the whole row (I mean redirect on click), but I failed to overwrite this action on the button

PR #1727, opened 2023-09-15
Fix handling of interrupts

moscicky:
When consumer process is deemed unhealthy it is interrupted. There were places in code when `InterruptedException` was catched but interrupted status was not restored. Because of that `ConsumerProcess#run` was unable to detect that the thread was interrupted:

```java
while (running && !Thread.currentThread().isInterrupted()) {
```

consumer process was still sending messages to subscribers but was marked as unhealthy and its messages were not commited. This situation would last until consumer was restarted. 

In this PR `interrupted` statuses are restored, enabling unhealthy consumer process to exit.


piotrrzysko (approved):

druminski (approved):
I can imagine it was not trivial to detect, congratulation, awesome bug fix 🥇

moscicky:
> I can imagine it was not trivial to detect, congratulation, awesome bug fix 🥇

@druminski all credits go to @piotrrzysko 👏🏻

piotrrzysko:
Don't be so modest. Without your pertinent questions it would take way longer to pinpoint the bug.

PR #1730, opened 2023-09-20
Kafka SASL enhancements

artlomako:
Now it's not possible to connect to Kafka with SASL other than Plain, i.e. SASL/SCRAM which is default for AWS Managed Kafka.
I added the `sasl.loginModule` kafka config property to allow change of the login module for SCRAM authentication.
I also allowed to override entire jaas config with `jaasConfig` property.


hermes-common/src/main/java/pl/allegro/tech/hermes/common/config/KafkaAuthorizationProperties.java
```diff
@@ -47,4 +49,20 @@ public String getPassword() {
     public void setPassword(String password) {
         this.password = password;
     }
+
```
faderskd, line 52: I know your code does not change name of the class, but this class should be actually named accordingly to the `authentication` not `authorization`, e.g. `KafkaAuthenticationProperties`. The same holds for all places where the properties are used.

artlomako: I cleaned up a bit things related to Kafka auth, added fallback setters to properties classes to ensure backward compatibility. Please check last two commits.

faderskd:
I like your idea about introducing jaas config. It gives us one more opportunity: removing username and password which is not required for all SASL authentication methods. The last place where we use the username/password is [here](https://github.com/allegro/hermes/blob/ff1614becd228229c062e7f80f5d56ff53336ab5/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaTopicMetadataFetcherFactory.java#L38). If we replace that with `kafkaParameters.getJaasConfig()` we can remove username/password from `KafkaParameters`

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/kafka/KafkaMessageReceiverFactory.java
```diff
@@ -168,14 +168,10 @@ private KafkaConsumer<byte[], byte[]> createKafkaConsumer(Topic topic, Subscript
...
     private void addKafkaAuthorizationParameters(Properties props) {
-        if (kafkaAuthorizationParameters.isEnabled()) {
-            props.put(SASL_MECHANISM, kafkaAuthorizationParameters.getMechanism());
-            props.put(SECURITY_PROTOCOL_CONFIG, kafkaAuthorizationParameters.getProtocol());
-            props.put(SASL_JAAS_CONFIG,
-                    "org.apache.kafka.common.security.plain.PlainLoginModule required\n"
-                            + "username=\"" + kafkaAuthorizationParameters.getUsername() + "\"\n"
-                            + "password=\"" + kafkaAuthorizationParameters.getPassword() + "\";"
-            );
+        if (kafkaParameters.isEnabled()) {
```
faderskd, line 171: It should be named like `isAuthenticationEnabled()`.  It is just the interface method so can be changed freely without worry about backward compatibility

artlomako:
@faderskd good point with jaas usage. In my last commit I removed `getUsername()` and `getPassword()` methods from interface and replaced usages with `getJaasConfig()`. I also used `KafkaParameters` interface in `management` module settings

faderskd (approved):

PR #1755, opened 2023-10-18
resolve #1645 | stop subscription immediately even when there are pen…

### LINKED ISSUE #1645: Batch subscription should be stoppable despite the pending message retries
If you have a batch subscription sending messages to the receiver returning 5xx status codes, the messages are retried until TTL is passed. Sending with retries blocks `BatchConsumer` main loop, thus blocking processing consumer signals. If the subscription was stopped at the moment of retries, it will wait until retries are exhausted. This is not a desired behaviour. The subscription should be stopped immediately even if the are pending retries.

pkasperowicz:
…ding retries


szczygiel-m (dismissed):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/supervisor/process/ConsumerProcess.java
```diff
@@ -116,6 +116,7 @@ private void process(Signal signal) {
                 case STOP:
                     logger.info("Stopping main loop for consumer {}. {}", signal.getTarget(), signal.getLogWithIdAndType());
                     this.running = false;
+                    stop();
```
faderskd, line 119: @pkasperowicz are you sure we need this ? The `stop()` method is actually called in a `finally` block in the `run()` method. The main `try` block in the `run()` method is stopped when `running.flag` is set to false.

pkasperowicz: When the stop signal is received, we need to inform the consumer to stop consuming events. 
`stop()` invokes `consumer.tearDown()` and it sets `BatchConsumer.consuming = false`. 
It informs `Retryer` to stop retrying messages. 

Alternative approach might be extending the API of BatchConsumer to recive stop signal and mark it as some boolean e.g. `stopSignalReceived`.

faderskd: ok but I'm wondering if it will work anyway because to call the `stop()` method we have to handle STOP signal. To handle STOP signal we have to process signals queue and run `signalsInterrupt.run()`. The retrier actually can be busy of sending messages and never handle signals. Have you tried to test it ? I'll try to replay the error on the docker environment or in integration tests.

faderskd: I think the case is now that consumer is not handling STOP singal because otherwise the running flag would be read -> and the while loop would stop -> and the `tearDown` method would be called 🤔

faderskd: ok I see it will work, because the signals queue is processed with every retry:
```
            retryer.call(() -> {
                loadRecorder.recordSingleOperation();
                signalsInterrupt.run();
                return sender.send(
```
Can we avoid duplicated calling `stop()` method ? For example we can write method like this:
```
    private void stopWithExceptionHandling() {
        if (!running) {
            return;
        }
        try {
            stop();
        } catch (Exception exceptionWhileStopping) {
            logger.error("An error occurred while stopping consumer process of subscription {}",
                    getSubscriptionName(), exceptionWhileStopping);
        } finally {
            onConsumerStopped.accept(getSubscriptionName());
            Thread.currentThread().setName("consumer-released-thread");
        }
    }
``` and use it it both in `finally` block of run method and in process method:
```                
...
case STOP:
                    logger.info("Stopping main loop for consumer {}. {}", signal.getTarget(), signal.getLogWithIdAndType());
                    stopWithExceptionHandling();
                    this.running = false;
                    break;
```

pkasperowicz: @faderskd ok changed it

faderskd: @pkasperowicz thank you, +1

faderskd (approved):

szczygiel-m (approved):

faderskd:
Moved to another [PR](https://github.com/allegro/hermes/pull/1974)

PR #1774, opened 2023-10-31
Case insensitive filter in group topics view

kszapsza:
- Filter in `GroupTopicsListing` view is case insensitive, so that we can find e.g. `MyFavoriteTopic` by typing `myfavoritetopic`. Same for `GroupListing` (though generally group names are lowercase either way).
- Fix `Unresolved function or method toBeInTheDocument()` warnings in `hermes-console-vue` tests (add `jest` to globals in tsconfigs).


szczygiel-m (changes requested):
Can You also apply this change to other listings? E.g subscription search in topicView

hermes-console-vue/src/api/inconsistent-group.ts
```diff
@@ -17,5 +17,5 @@ export interface InconsistentSubscription {
 
 export interface InconsistentMedata {
   datacenter: string;
-  content: string | undefined;
+  content?: string;
```
kszapsza, line 20: When we have `content: string | undefined`, we can have `undefined` value, but the key `content` has to be present in the object, so in stubs we have to write: `content: undefined` (otherwise we have a type error).

In case of `content?: string`, we can both have object with key `content` with value `undefined` provided explicitly, but we also can omit this key completely (e.g. in test data, the value of absent key will still be `undefined` in JS).

hermes-console-vue/src/views/topic/subscriptions-list/SubscriptionsList.vue
```diff
@@ -25,27 +25,30 @@
 
   const subscriptionItems = computed(() =>
     props.subscriptions
+      ?.filter((subscription) =>
```
kszapsza, line 28: Previous approach in which we mapped either to object (name/color/statusText/href) or to `null` and then filtered with `.filter(Boolean)` was problematic for TS type system. Though the operation filtered out null objects, IntelliJ still complained in Vue template below, that `subscriptionItems` elements can be possibly `null`.

We could introduce an interface:
```typescript
interface SubscriptionListItem {
  name: string;
  color: string;
  statusText: State;
  href: string;
} 
```
and then change `.filter(Boolean)` filter to the so-called TypeScript type predicate:
```typescript
.filter((subscription): subscription is SubscriptionListItem => subscription !== null)
```
then TS type system would know, that null values possibly introduced in `map` above were filtered out.

But IMHO it is cleaner to simply perform subscription filtering before `map`, as I did here

szczygiel-m (approved):

piotrrzysko (approved):

PR #1780, opened 2023-11-13
Use ephemeral ports in integration tests

piotrrzysko:


szczygiel-m (approved):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/HermesServer.java
```diff
@@ -151,4 +159,23 @@ private HttpHandler handlers() {
     private boolean isFrontendRequestDumperEnabled() {
         return hermesServerParameters.isRequestDumperEnabled();
     }
+
+    @VisibleForTesting
+    @SuppressWarnings("unchecked")
+    public int getPort() {
+        Field channelsField = FieldUtils.getDeclaredField(Undertow.class, "channels", true);
```
druminski, line 166: Do we need to use reflection to read Undertow port? According to a [thread](https://stackoverflow.com/questions/55678642/unable-to-fetch-random-port-when-server-port-is-0) on Stackoverflow something like this should work:

```
InetSocketAddress socketAddress = (InetSocketAddress) undertow.getListenerInfo().stream().findFirst().orElseThrow(() -> new IllegalStateException("No port available yet.")).getAddress();
return socketAddress.getPort();
```

druminski (approved):
+1 but please take a look at my comment related to reflection

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/HermesServer.java
```diff
@@ -151,4 +159,23 @@ private HttpHandler handlers() {
     private boolean isFrontendRequestDumperEnabled() {
         return hermesServerParameters.isRequestDumperEnabled();
     }
+
+    @VisibleForTesting
```
Diewa, line 163: If I understand correctly, this annotation makes the method visible in tests. So do we need this annotation when the method is public, or do we need the public access modifier?

druminski: > If I understand correctly, this annotation makes the method visible in tests. 

Nope, this annotation is from Guava and doesn't change anything, it only states that the method was created only for tests. But your question is still valid whether we need it or not. When comes to me, it's ok to use it as it states explicite that the method was created only for test purposes and maybe you shouldn't use it in production.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/HermesServer.java
piotrrzysko: Good point. I applied your suggestion. Thanks.

PR #1790, opened 2023-11-28
Rewriting tests part 2

szczygiel-m:
1. JmsConsumingTest
2. FilteringJsonTest
3. FilteringHeadersTest


moscicky (changes requested):
One minor issue to fix and then +1 from me

integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/FilteringJsonTest.java
```diff
@@ -0,0 +1,96 @@
...
+                .build());
+
+        // when
+        hermes.api().publishUntilSuccess(topic.getQualifiedName(), ALICE.asJson());
+        hermes.api().publishUntilSuccess(topic.getQualifiedName(), ALICE_GREY.asJson());
+        hermes.api().publishUntilSuccess(topic.getQualifiedName(), BOB.asJson());
+        hermes.api().publishUntilSuccess(topic.getQualifiedName(), BOB_GREY.asJson());
+
+        // then
+        subscriber.waitUntilReceived(BOB_GREY.asJson());
```
moscicky, line 74: I'd say this check is insufficient because it only checks whether `BOB_GREY` has been received and does not fail when receiving other messages. Can we add an assertion that only this message was received? This also applies to other filtering tests

moscicky (approved):

PR #1796, opened 2023-12-11
rewrite e2e tests

moscicky:
Rewrite tests which require custom frontend setup:
- buffer loading test
- graceful shutdown test
- https client publishing test
- publishing on avro topic without schema test
- attaching keep alive header test

Additionally reuse kafka, zk and schema registry between those tests


integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/AsyncPublisherAutoRetryTest.java
```diff
@@ -0,0 +1,74 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+
+public class AsyncPublisherAutoRetryTest {
+    private static final String TOPICS_ROOT = "/topics/";
+    private static final String EXISTING_TOPIC = "existing";
+    private static final String EXISTING_TOPIC_URL = TOPICS_ROOT + EXISTING_TOPIC;
+    private static final String NOT_EXISTING_TOPIC = "not-existing";
+    private static final String NOT_EXISTING_TOPIC_URL = TOPICS_ROOT + NOT_EXISTING_TOPIC;
+    private static final int frontendPort = Ports.nextAvailable();
```
piotrrzysko, line 27: I wouldn't use `Ports.nextAvailable()`. Take a look at how it is implemented. It opens and closes a socket, potentially leaving it in the `TIME_WAIT` state. This, in turn, sometimes leads to "Address already in use" errors. Instead, I'd use ephemeral ports - just pass 0 as a port.

integration-tests/src/slowIntegrationTest/java/pl/allegro/tech/hermes/integrationtests/HermesServerGracefulShutdownTest.java
```diff
@@ -0,0 +1,65 @@
...
+    public static InfrastructureExtension infra = new InfrastructureExtension();
+
+    private HermesFrontendTestApp frontend;
+    private HermesServer hermesServer;
+    FrontendTestClient frontendClient;
+
+    @BeforeEach
+    public void beforeEach() {
+        frontend = new HermesFrontendTestApp(infra.hermesZookeeper(), infra.kafka(), infra.schemaRegistry());
+        frontend.withProperty(FRONTEND_GRACEFUL_SHUTDOWN_ENABLED, false);
```
piotrrzysko, line 32: If I'm not mistaken, this parameter doesn't affect the tests in this class in any way. Yet it is a bit misleading to set it to true while we are testing that the graceful shutdown is enabled. Perhaps, we can skip setting it.

integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/AsyncPublisherAutoRetryTest.java
```diff
@@ -0,0 +1,74 @@
...
+import pl.allegro.tech.hermes.test.helper.util.Ports;
+
+import java.net.URI;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+
+public class AsyncPublisherAutoRetryTest {
```
piotrrzysko, line 21: What do you think about moving this test to the hermes-client module? Take a look at `pl.allegro.tech.hermes.client.ReactiveHermesSenderTest`. This one can also be considered an integration test but is placed in the hermes-client module. 

The other way around approach is also possible, of course, but I lean towards having only tests in the integration-tests module that span multiple hermes modules.

moscicky: Looking more into this I wonder if this test is needed at all. In `hermes-client` module we already have tests covering cases tested here. I'd vote for removing this test class completely

integration-tests/src/slowIntegrationTest/java/pl/allegro/tech/hermes/integrationtests/HermesServerGracefulShutdownTest.java
moscicky: agree, I've set it by default as it was in old e2e tests and removed it from this test

integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/AsyncPublisherAutoRetryTest.java
moscicky: good catch, applied the suggestion in another test

piotrrzysko (approved):

PR #1809, opened 2023-12-28
fix date parser in OfflineRetransmissionRequest class

Diewa:
**context:** The requests creating a task for a new retransmission, coming from the UI, have the following date format: 'yyyy-MM-dd'T'HH:mm:ss'Z'' or 'yyyy-MM-dd'T'HH:mm'Z''. If we set the date with minute precision from the UI, the request will not contain seconds at all. For example, 30.09.2023, 22:00:00 will be send as 30.09.2023, 22:00.

**problem description:**
We noticed that if the transmitted date has second precision (e.g., 30.09.2023, 22:00:11), it will not be handled correctly, and the user will receive an error: 'endTimestamp/startTimestamp must not be null,' which is related to how we handle requests.

This pull request adds support for a date format that includes seconds.


hermes-api/src/main/java/pl/allegro/tech/hermes/api/OfflineRetransmissionRequest.java
```diff
@@ -37,25 +41,25 @@ public OfflineRetransmissionRequest(
...
+        this.startTimestamp = initializeTimestamps(startTimestamp);
+        this.endTimestamp = initializeTimestamps(endTimestamp);
     }
 
-    private void initializeTimestamps(String startTimestamp, String endTimestamp) {
-        if (startTimestamp == null || endTimestamp == null) {
-            this.startTimestamp = null;
-            this.endTimestamp = null;
-            return;
+    private Instant initializeTimestamps(String timestamp) {
```
szczygiel-m, line 48: nit: can be renamed to `initializeTimestamp` since there is only one function parameter

szczygiel-m (approved):

szczygiel-m (approved):

pitagoras3 (approved):

PR #1820, opened 2024-01-25
Minor tests stability improvements

szczygiel-m:


integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/PublishingAndConsumingTest.java
```diff
@@ -222,6 +224,13 @@ public void shouldNotConsumeMessagesWhenSubscriptionIsSuspended() {
         hermes.api().waitUntilSubscriptionActivated(topic.getQualifiedName(), subscriptionName);
         hermes.api().suspendSubscription(topic, subscriptionName);
         hermes.api().waitUntilSubscriptionSuspended(topic.getQualifiedName(), subscriptionName);
+        // wait until consumer group removed
+        waitAtMost(30, TimeUnit.SECONDS).until(() ->
```
druminski, line 228: Shouldn't this check be a part of "waitUntilSubscription{Suspended,Removed}()" method? Motivation, next time when a developer will write a test which calls waitUntilSubscription{Suspended,Removed}() and do some assertions after it, then without the assertion which you wrote here, the developer can receive some false-positives information. What do you think about it?

druminski (dismissed):
Added one non blocking comment, +1

integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/PublishingAndConsumingTest.java
szczygiel-m: Nice catch! fixed 😄

druminski (approved):

faderskd (approved):

PR #1823, opened 2024-02-22
add multi dc broker producer

moscicky:


hermes-api/src/main/java/pl/allegro/tech/hermes/api/ErrorCode.java
```diff
@@ -63,7 +63,8 @@ public enum ErrorCode {
     UNKNOWN_MIGRATION(NOT_FOUND),
     INVALID_QUERY(BAD_REQUEST),
     IMPLEMENTATION_ABSENT(NOT_FOUND),
-    MOVING_SUBSCRIPTION_OFFSETS_VALIDATION_ERROR(BAD_REQUEST);
+    MOVING_SUBSCRIPTION_OFFSETS_VALIDATION_ERROR(BAD_REQUEST),
+    SENDING_TO_KAFKA_TIMEOUT(GATEWAY_TIMEOUT);
```
piotrrzysko, line 67: I'm on the fence with this status code. On one hand, I like it because it is more specialized than, for example, 503. However, it suggests that Hermes acts just as a proxy to something (which is kinda true ;)), but our clients don't necessarily think about it that way. So, they might assume that a load balancer between Hermes and them has returned it. Additionally, within the total request duration, we perform some processing (of course not too much, and usually Kafka is the source of publish requests latency), so it's not like we just pass requests to an external dependency. 

Do you think 503 is too ambiguous in our context?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/Producers.java
```diff
@@ -1,40 +1,45 @@
...
         return topic.isReplicationConfirmRequired() ? ackAll : ackLeader;
     }
 
+    public List<KafkaProducer<byte[], byte[]>> getRemote(Topic topic) {
+        return topic.isReplicationConfirmRequired() ? remoteAckLeader : remoteAckAll;
+    }
+
+
+    //TODO: should probably be split per DC
     public void registerGauges(MetricsFacade metricsFacade) {
```
piotrrzysko, line 43: Since `pl.allegro.tech.hermes.frontend.producer.kafka.KafkaProducer` is now responsible for instrumentation and is aware of its datacenter, wouldn't it be better to move this method there?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaProducer.java
```diff
@@ -0,0 +1,166 @@
...
+import pl.allegro.tech.hermes.metrics.HermesTimerContext;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.Future;
+import java.util.function.Supplier;
+
+// TODO: is it a good idea to create abstraction on top of Producer? Currently it provides DC and instrumentation
```
piotrrzysko, line 28: I'm fine with wrapping an instance of `org.apache.kafka.clients.producer.Producer` in something that adds instrumentation and datacenter awareness, but implementing the entire interface seems overkill. Maybe, let's just add a class that exposes only necessary methods (`send`, `close`, `partitionsFor`) + adds the mentioned features in the comment. Additionally, I'd refrain from calling the wrapper KafkaProducer. In my opinion, this name is "reserved" for the Kafka producer provided by the Kafka client library. I guess something like `KafkaMessageSender` is good enough.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/buffer/SimpleExecutionCallback.java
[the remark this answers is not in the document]
moscicky: removed

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/BufferAwareBrokerMessageProducer.java
```diff
@@ -0,0 +1,31 @@
+package pl.allegro.tech.hermes.frontend.producer.kafka;
+
+import pl.allegro.tech.hermes.frontend.metric.CachedTopic;
+import pl.allegro.tech.hermes.frontend.producer.BrokerMessageProducer;
+import pl.allegro.tech.hermes.frontend.publishing.PublishingCallback;
+import pl.allegro.tech.hermes.frontend.publishing.message.Message;
+
+public class BufferAwareBrokerMessageProducer implements BrokerMessageProducer {
```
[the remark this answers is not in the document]
moscicky: yes, this nam was temporary. I have changed the names to the ones you suggested

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/RemoteProducerProvider.java
```diff
@@ -0,0 +1,9 @@
+package pl.allegro.tech.hermes.frontend.producer.kafka;
+
+import pl.allegro.tech.hermes.frontend.metric.CachedTopic;
+
+import java.util.Optional;
+
+public interface RemoteProducerProvider {
```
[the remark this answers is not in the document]
moscicky: good idea, I've removed this interface

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/config/HandlersChainProperties.java
```diff
@@ -12,6 +12,8 @@ public class HandlersChainProperties implements HandlersChainParameters {
 
     private Duration longIdleTimeout = Duration.ofMillis(400);
 
+    private Duration globalAsyncTimeout = Duration.ofMillis(500);
```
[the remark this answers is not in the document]
moscicky: I agree. This name was temporarily chosen based on our initial discussions. I have renamed the parameter and added a documentation for those parameters in `HandlersChainParameters`. After incubation this parameter should also be added to rtd

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaProducer.java
moscicky: > I'm fine with wrapping an instance of org.apache.kafka.clients.producer.Producer in something that adds instrumentation and datacenter awareness, but implementing the entire interface seems overkill. Maybe, let's just add a class that exposes only necessary methods (send, close, partitionsFor) + adds the mentioned features in the comment

You are right, there is no point in delegating all the methods when only some of them are used. It also very confusing that there would be 3 available send methods. Left only those which are needed.

I've also change the name to `KafkaMessageSender` but this resulted in `Producers` being changed to `KafkaMessageSenders`  for consistency

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/Producers.java
moscicky: Good idea, I've moved metrics to `KafkaMessageSender`. `KafkaMessageSenders` is now only a thin wrapper

hermes-api/src/main/java/pl/allegro/tech/hermes/api/ErrorCode.java
moscicky: > Additionally, within the total request duration, we perform some processing (of course not too much, and usually Kafka is the source of publish requests latency), so it's not like we just pass requests to an external dependency.

This error is right now returned when we are pretty much sure that Kafka was the source of latency:

```java
public boolean setTimeoutSendingToKafka() {
        return state.compareAndSet(SENDING_TO_KAFKA_PRODUCER_QUEUE, TIMEOUT_SENDING_TO_KAFKA) || state.compareAndSet(SENDING_TO_KAFKA, TIMEOUT_SENDING_TO_KAFKA);
    }
```

My initial idea was that if clients are monitoring Hermes responses they would be able to tell the difference between 500 and 504. 500 would mean "server side, unexpected error" while 504 would mean "server side, sort-of expected error"? But maybe you are right that 504 should be used exclusively by proxies (say nginx), I would say that 503 is sufficient

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/MultiDatacenterMessageProducer.java
```diff
@@ -0,0 +1,166 @@
...
+                      Message message) {
+        String datacenter = sender.getDatacenter();
+        try {
+            sender.send(producerRecord, cachedTopic, message, new DCAwareCallback(
+                    message,
+                    cachedTopic,
+                    datacenter,
+                    callback));
+        } catch (Exception e) {
+            // message didn't get to internal producer buffer and it will not be send to a broker
```
faderskd, line 79: typo: send -> sent

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/MultiDatacenterMessageProducer.java
```diff
@@ -0,0 +1,166 @@
...
+                    callback.onUnpublished(message, cachedTopic, datacenter, exception);
+                }
+            }
+        }
+
+    private static class SendCallback {
+
+        private final PublishingCallback callback;
+        private final AtomicBoolean sent = new AtomicBoolean(false);
+        private final AtomicInteger tries;
```
faderskd, line 108: Just curious: this is atomic because the `onUnpublished` (where the variable is decremented) can be called concurrently be a local and remote producer ?

faderskd (dismissed):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/MultiDatacenterMessageProducer.java
piotrrzysko: fixed

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/MultiDatacenterMessageProducer.java
piotrrzysko: Yes, that's the reason why it is atomic.

faderskd (approved):

piotrrzysko (approved):

PR #1826, opened 2024-02-29
Propagate readiness status to all datacenters

piotrrzysko:
### Intro
This is part of the feature of falling back to remote datacenters when the local one is unhealthy during publishing. Considering that hermes-frontend has access only to ZooKeeper in its local datacenter, it needs to have information there on the readiness status of all datacenters to select only those that are ready. Before this change, only the readiness status of the local datacenter was stored in the local ZooKeeper.

### Migration
* To ensure backward compatibility with previous versions, this PR introduces a new ZooKeeper path (`/datacenter-readiness`) to store readiness status, instead of using the old path (`/frontend/readiness`). Thus, rollbacks to older versions are possible without any additional changes.
* The new path (`/datacenter-readiness`) is populated on the first and subsequent readiness status changes. Versions including this change do not respect values set in the previous path (`/frontend/readiness`). If the new path does not exist or is empty, Hermes assumes that all datacenters are ready.
* The old path is not automatically removed. Therefore, we recommend manually deleting it after migration.


moscicky (approved):
very smooth change. I would only suggest to mention the migration process in the PR descriptions. As far as I understand defaults would always return ready so we can just mention that that true values will be persisted in ZK on first save

piotrrzysko:
>  I would only suggest to mention the migration process (or lack of it) and whether the change is safe to rollback in the PR description.

Good point. Added.

szczygiel-m (approved):

PR #1832, opened 2024-03-04
extract checking topic availability from BrokerMessageProducer

piotrrzysko:
Addresses https://github.com/allegro/hermes/blob/e27aef8dfc0dfc56a35f53c4ebbda607da76e3f4/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/MultiDCKafkaBrokerMessageProducer.java#L156


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingRunner.java
```diff
@@ -1,81 +1,154 @@
...
-        return allResults;
+    private boolean refreshMetadataForDatacenters(List<String> datacenters) {
+        try {
+            long start = System.currentTimeMillis();
+            logger.info("Loading topic metadata for datacenters: {}", datacenters);
+            List<CachedTopic> topics = topicsCache.getTopics();
+            List<MetadataLoadingResult> allResults = loadMetadataForTopics(topics, datacenters);
+            logger.info("Finished loading topic metadata in {}ms", System.currentTimeMillis() - start);
+            logResultInfo(allResults);
+            return allResults.stream().anyMatch(MetadataLoadingResult::isFailure);
```
moscicky, line 74: `allResults.stream().anyMatch(MetadataLoadingResult::isFailure)` means that we return `true` if any job fails. Shouldn't we negate the result?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingJob.java
```diff
@@ -13,8 +11,6 @@
 
 public class TopicMetadataLoadingJob implements Runnable {
```
moscicky, line 12: not really related to this PR but I think it would be helpful to at least add a comment what is this the purpose of this class - i.e. to keep producers warm. Without that it is hard to tell, it refreshes topics in the background and discards the result - what for?

hermes-frontend/src/test/groovy/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingRunnerTest.groovy
```diff
@@ -33,60 +33,60 @@ class TopicMetadataLoadingRunnerTest extends Specification {
...
     def "should leave retry loop when reached max retries and failed to load metadata"() {
         given:
-        BrokerMessageProducer producer = Mock()
-        def hook = new TopicMetadataLoadingRunner(producer, topicsCache, 2, Duration.ofSeconds(10), 4)
+        BrokerTopicMetadataFetcher checker = Mock()
+        def hook = new TopicMetadataLoadingRunner(checker, "dc1", ["dc1", "dc2"], topicsCache, 2, Duration.ofSeconds(10), 4)
 
         when:
-        hook.refreshMetadata()
+        hook.refreshMetadataForLocalDatacenter()
```
moscicky, line 72: how about checking the result of this method?

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerTopicMetadataFetcher.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.apache.kafka.common.PartitionInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import pl.allegro.tech.hermes.frontend.metric.CachedTopic;
+import pl.allegro.tech.hermes.frontend.producer.BrokerTopicMetadataFetcher;
+
+import java.util.List;
+
+public class KafkaBrokerTopicMetadataFetcher implements BrokerTopicMetadataFetcher {
```
moscicky, line 11: I am wondering whether it makes sense to separate `KafkaBrokerTopicMetadataFetcher ` from `KafkaTopicMetadataFetcher`:
1. they names are really similar which makes it confusing
2. KafkaTopicMetadataFetcher is only used by KafkaBrokerTopicMetadataFetcher

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerTopicMetadataFetcher.java
```diff
@@ -0,0 +1,63 @@
...
+            return false;
+        }
+
+        logger.warn("No information about partitions for topic {}", kafkaTopicName);
+        return false;
+    }
+
+    @Override
+    public boolean tryFetchFromDatacenter(CachedTopic topic, String datacenter) {
+        // TODO: To fetch metadata for a selected datacenter https://github.com/allegro/hermes/pull/1823 is required.
```
moscicky, line 51: In the final implementation we have to be extra careful: this class uses `Producers` and `KafkaTopicMetadataFetcher`. `KafkaTopicMetadataFetcher` in turn uses `AdminClient`. So we are essentially using two different Kafka clients (KafkaProducer and AdminClient) here and we need to make sure in the future that both of them refer to the same datacenter

moscicky (changes requested):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/server/TopicMetadataLoadingRunner.java
moscicky: `KafkaReadinessCheckTest` is probably failing because of this

piotrrzysko: Correct. Fixed.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerTopicMetadataFetcher.java
piotrrzysko: Good point. Merged.

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/producer/kafka/KafkaBrokerTopicMetadataFetcher.java
piotrrzysko: Perhaps, an option worth considering is to introduce datacenter awareness in this PR since its scope is relatively narrow and it is easier to spot any bugs.

piotrrzysko: Let me play around with this a bit. Perhaps remodeling this class slightly will address this comment and https://github.com/allegro/hermes/pull/1832#discussion_r1512650819.

moscicky (approved):

PR #1836, opened 2024-03-10
Add basic HermesMock listener for kotest

deejay1:
# What
Add basic implementation of a kotest extension based on the Wiremock one.

# Why
To provide a ready to go implementation for kotest users.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,3 +1,6 @@
+plugins {
+    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.5.0'
+}
 rootProject.name='hermes'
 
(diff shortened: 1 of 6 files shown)


settings.gradle
```diff
@@ -1,3 +1,6 @@
+plugins {
+    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.5.0'
```
bgalek, line 2:
```suggestion
    id 'org.gradle.toolchains.foojay-resolver-convention' version '1.0.0'
```

bgalek: And you can convert this file to KTS while at it :)

PR #1844, opened 2024-04-10
switch back to regular wiremock but with jetty12 support

bgalek:
When using `hermes-mock` with `wiremock-jetty12` tests in such project will have problem choosing property wiremock version (since wiremock-standalone will be on their classpath).

This should make builds faster and more accessible for people to use :)


szczygiel-m:
Hey, thanks for the contribution! 🎉  It seems that after your change the unit tests are not passing, could you take a look at it?

bgalek:
yeah, I'll look into that :)

PR #1855, opened 2024-04-29
provide more visibility into offline retransmission task creation

moscicky:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/OfflineRetransmissionEndpoint.java
```diff
@@ -91,9 +91,9 @@ private void ensurePermissionsToBothTopics(OfflineRetransmissionRequest request,
     private static class OfflineRetransmissionAuditor {
         private static final Logger logger = LoggerFactory.getLogger(OfflineRetransmissionAuditor.class);
 
-        public void auditRetransmissionCreation(OfflineRetransmissionRequest request, ContainerRequestContext requestContext) {
+        public void auditRetransmissionCreation(OfflineRetransmissionRequest request, ContainerRequestContext requestContext, OfflineRetransmissionTask task) {
```
szczygiel-m, line 94: nit: what do You think about passing only taskId instead of whole Task object?

szczygiel-m (approved):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/api/OfflineRetransmissionEndpoint.java
moscicky: I'd say that `OfflineRetransmissionTask` being return is more descriptive than `String` and fields are all immutable anyway. I've also seen (e.g. in [spring](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html)) that `save` methods to return the whole saved object

moscicky: I'd rather go for `OfflineRetransmissionTask` since it is the task being audited and it has other fields. We only care for id for now tough. But I'd stick to `OfflineRetransmissionTask`

szczygiel-m (approved):

faderskd (approved):

PR #1884, opened 2024-07-30
Faster topic removal in MultiDC setup

moscicky:
Fixes a race condition due to which topic/group removal in remote DC was taking up to 150s. Exact bug is described in `pl.allegro.tech.hermes.infrastructure.zookeeper.ZookeeperTopicRepository#removeTopic`. After this change group/topic removal takes tens of ms in remote DC.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
@@ -156,6 +156,16 @@ public abstract class ZookeeperBasedRepository {
                 .commit();
     }
 
+    protected void deleteInTransaction(String path, String childPath) throws Exception {
+        ensureConnected();
+        zookeeper.inTransaction()
+                .delete().forPath(childPath)
+                .and()
+                .delete().forPath(path)
+                .and()
+                .commit();
+    }
+
     protected void create(String path, Object value) throws Exception {
         ensureConnected();
         zookeeper.create().forPath(path, mapper.writeValueAsBytes(value));
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
@@ -183,13 +183,8 @@ public abstract class ZookeeperBasedRepository {
     }
 
     protected void remove(String path) throws Exception {
-        logger.info("Removing path {} from ZK", path);
-        long start = System.currentTimeMillis();
         ensureConnected();
-        logger.info("Ensured connected to ZK in {} ms", System.currentTimeMillis() - start);
-        start = System.currentTimeMillis();
         zookeeper.delete().guaranteed().deletingChildrenIfNeeded().forPath(path);
-        logger.info("Deleted path {} from ZK in {} ms", path, System.currentTimeMillis() - start);
     }
 
     private interface ThrowingReader<T> {
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
@@ -5,6 +5,8 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.api.transaction.CuratorTransactionFinal;
+import org.apache.curator.utils.ZKPaths;
 import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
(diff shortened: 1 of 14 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperTopicRepository.java
```diff
@@ -77,12 +78,67 @@ public void createTopic(Topic topic) {
...
+        if (pathExists(topicPreviewPath)) {
+            pathsForRemoval.add(topicPreviewPath);
+        }
+
+        pathsForRemoval.add(paths.subscriptionsPath(topicName));
+        pathsForRemoval.add(paths.topicPath(topicName));
+
         try {
-            remove(paths.topicPath(topicName));
+            deleteInTransaction(pathsForRemoval);
```
faderskd, line 141: should we also use this method in other places ? because now we have two different methods for removing paths🤔 for paths without children it will be no-brainer (just use a simple remove),  but for those with children we should consider both

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/dc/MultiDatacenterRepositoryCommandExecutor.java
```diff
@@ -58,7 +61,7 @@ private <T> void execute(RepositoryCommand<T> command, boolean isRollbackEnabled
                     throw ExceptionWrapper.wrapInInternalProcessingExceptionIfNeeded(e, command.toString(), repoHolder.getDatacenterName());
                 }
             } catch (Exception e) {
-                logger.warn("Execute failed with an error", e);
+                logger.warn("Failed to execute repository command: {} in ZK dc: {}", command, System.currentTimeMillis() - start, e);
```
faderskd, line 64: `Failed to execute repository command: {} in ZK dc: {} in: {} ms", repoHolder.getDatacenterName(), System.currentTimeMillis() - start`

moscicky: fixed

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperTopicRepository.java
moscicky: Transactional removal should only be used when needed - it has a higher overhead + it is required to supply all the paths. Most of the time we remove things for prefix,  e.g. delete all nodes for topic and this requires `deleteChildrenIfNeeded` which is not available in transactions. So imo these two methods have different usage scenarios

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
```diff
@@ -156,6 +166,20 @@ protected void createInTransaction(String path, Object value, String childPath)
                 .commit();
     }
 
+    protected void deleteInTransaction(List<String> paths) throws Exception {
+        if (paths.isEmpty()) {
+            throw new InternalProcessingException("Attempting to remove empty set of paths from ZK");
+        }
+        ensureConnected();
+        CuratorTransactionFinal transaction = zookeeper.inTransaction().delete().forPath(paths.get(0)).and();
```
pitagoras3, line 174: **issue:** ZooKeeper API mentions that `inTransaction` method is deprecated. 

```java
/**
 * Start a transaction builder
 *
 * @return builder object
 * @deprecated use {@link #transaction()} instead
 */
public CuratorTransaction inTransaction();
```

Can we use `transaction()` method here?

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
```diff
@@ -156,6 +166,20 @@ protected void createInTransaction(String path, Object value, String childPath)
...
+
+        for (int i = 1; i < paths.size(); i++) {
+            transaction = transaction.delete().forPath(paths.get(i)).and();
+        }
+
+        transaction.commit();
+    }
```
pitagoras3, line 181: **note:** Here is my proposition for refactor:

```java
protected void deleteInTransaction(List<String> paths) throws Exception {
    if (paths.isEmpty()) {
        throw new InternalProcessingException("Attempting to remove empty set of paths from ZK");
    }
    ensureConnected();
    zookeeper.transaction().forOperations(
            paths.stream().map(x -> {
                try {
                    return zookeeper.transactionOp().delete().forPath(x);
                } catch (Exception e) {
                    throw new RuntimeException(e);
                }
            }).collect(Collectors.toList())
    );
}
```

Feel free to tweak with it :)

pitagoras3 (changes requested):

pitagoras3 (approved):

hermes-common/src/main/java/pl/allegro/tech/hermes/infrastructure/zookeeper/ZookeeperBasedRepository.java
pitagoras3: Will be resolved here: https://github.com/allegro/hermes/issues/1886

faderskd (approved):

PR #1887, opened 2024-08-05
do not remove subscription if it already existed before subscription creation

moscicky:
This PR changes rollbacks of multi dc commands to be exception aware and changes `CreateSubscriptionRepositoryCommand` so that subscription is not removed if it already existed before. 

That could happen if two subscription creation requests are made concurrently so that both requests are [marked as valid](https://github.com/allegro/hermes/blob/master/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/subscription/validator/SubscriptionValidator.java#L54), one creates subscription in storage and the other fails and performs rollback which causes the subscription to be removed. 

In multi DC setup, this would result in inconsistent state:

dc 1 -> null
dc 2 -> susbscription

### DIFF
--- a/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/commands/CreateTopicRepositoryCommand.java
+++ b/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/commands/CreateTopicRepositoryCommand.java
@@ -22,7 +22,7 @@ public class CreateTopicRepositoryCommand extends RepositoryCommand<TopicReposit
     }
 
     @Override
-    public void rollback(DatacenterBoundRepositoryHolder<TopicRepository> holder) {
+    public void rollback(DatacenterBoundRepositoryHolder<TopicRepository> holder, Exception exception) {
         /*
         We don't want to do a rollback due to possible race conditions with creating a topic on Kafka.
         It increases the probability of discrepancies between Kafka and Zookeeper: topic exists in Kafka,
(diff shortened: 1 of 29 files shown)


hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/commands/CreateTopicRepositoryCommand.java
```diff
@@ -22,7 +22,7 @@ public void execute(DatacenterBoundRepositoryHolder<TopicRepository> holder) {
     }
 
     @Override
-    public void rollback(DatacenterBoundRepositoryHolder<TopicRepository> holder) {
```
pitagoras3, line 25: **question:** Can we add similar logic for `TopicAlreadyExistsException`?

moscicky: rollbacks for `CreateTopicRepositoryCommand` are already disabled:
https://github.com/allegro/hermes/blob/a3cf6fd7ed7750815a260a69fd37a0193375511d/hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/topic/commands/CreateTopicRepositoryCommand.java#L25

I believe we can revisit topic rollbacks in the future with the current knolwedge

pitagoras3 (approved):

faderskd (approved):

pitagoras3 (approved):

PR #1891, opened 2024-08-14
Fixed bug leading to semaphore leak when filtering messages

szczygiel-m:


faderskd (approved):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/offset/PendingOffsets.java
```diff
@@ -72,6 +76,8 @@ public Map<SubscriptionPartitionOffset, MessageState> getOffsetsSnapshotAndRelea
             if (entry.getValue() == MessageState.PROCESSED) {
                 slots.remove(entry.getKey());
                 permitsReleased++;
+            } else if (entry.getValue() == MessageState.FILTERED) {
```
faderskd, line 79: Maybe it would be convenient to have a comment that we only increase semaphore for sending messages, and omitting increasing for filtered, but in any case we put a slot for both of the messages. I just had some problems with understanding how there can be slots in the queue but we didn't increased semaphore for them, thus the comment would help.

szczygiel-m: I have changed the approach and now filtered messages acquires semaphore, please take a look now 😄

integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/MetricsTest.java
```diff
@@ -257,6 +257,69 @@ public void shouldReportMetricForFilteredSubscription() {
...
+                        .expectBody(String.class)
+                        .value((body) -> assertThatMetrics(body)
+                                .contains("hermes_consumers_subscription_inflight")
+                                .withLabels(
+                                        "group", topic.getName().getGroupName(),
+                                        "subscription", subscription.getName(),
+                                        "topic", topic.getName().getName()
+                                )
+                                .withValue(1.0)
+                        )
```
moscicky, line 319: why 1.0 instead of 2.0?

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Message.java
```diff
@@ -53,6 +53,8 @@ public class Message implements FilterableMessage {
 
     private long currentMessageBackoff = -1;
 
+    private boolean isFiltered = false;
```
moscicky, line 56: we must ensure that the this value is visible to all threads (see the comment on top of this class)

moscicky (changes requested):
Previously the semantics of `Optional.empty` from `MessageReceiver#next` were that the message either the message was not present in kafka or it was filtered. Please add a java doc in `MessageReceiver` explaining what are the meanings of `Optional.empty` and `Messages#isFiltered` now

integration-tests/src/integrationTest/java/pl/allegro/tech/hermes/integrationtests/MetricsTest.java
szczygiel-m: Because previously we set `subscriptionPolicy().withInflightSize(1)` and we want to be sure, that this value wouldn't be exceeded when there are more messages and there were messages filtered previously.

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/Message.java
szczygiel-m: I have checked it and this is always used within a single thread.

moscicky (approved):

hermes-consumers/src/main/java/pl/allegro/tech/hermes/consumers/consumer/receiver/MessageReceiver.java
```diff
@@ -10,6 +10,21 @@
 
 public interface MessageReceiver {
 
+    /**
+     * Retrieves the next available message from the queue.
+     *
```
faderskd, line 15: ❤️

faderskd (approved):

Issue #1904, opened 2024-09-30
Add navigation-drawer in hermes-console


pitagoras3:
## What?

In UI (hermes-console) we would like to have possibility to use [navigation drawer](https://vuetifyjs.com/en/components/navigation-drawers/).

## Why?

To make it easier to navigate through hermes-console.

maknik:
I really want to take care of this issue!

PR #1914, opened 2024-10-16
Navigation side bar (#1904)

maknik:
Screenshot 2024-10-16 at 21 02 06


hermes-console/src/components/navigation-drawer/NavigationDrawer.vue
```diff
@@ -0,0 +1,130 @@
...
+  const roles = useRoles(null, null)?.roles;
+  const rail = ref(false);
+</script>
+
+<template>
+  <v-navigation-drawer :rail="rail" permanent @click="rail = false">
+    <template v-if="!isLoggedIn">
+      <v-list>
+        <v-list-item
+          :prepend-avatar="`https://alleavatar.allegrogroup.com/${userData.user_name}.jpg`"
```
szczygiel-m, line 26: Since it is opensourced repository we can't hardcode such internal links. We should remove it or make it a configurable string, but the second option would require more changes also on the backend site of hermes-management.

szczygiel-m (commented):
Nice PR! 😄

hermes-console/src/components/navigation-drawer/NavigationDrawer.vue
pitagoras3: I agree with Mateusz, can you remove this?

WojciechSova:
Moved to https://github.com/allegro/hermes/pull/1968

PR #1921, opened 2024-11-08
Delete Kafka consumer group on Hermes subscription deletion

MarcinBobinski:


hermes-management/src/main/java/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManager.java
```diff
@@ -80,4 +86,31 @@ public void createConsumerGroup(Topic topic, Subscription subscription) {
           e);
     }
   }
+
+  @Override
+  public void deleteConsumerGroup(Topic topic, Subscription subscription) {
```
szczygiel-m, line 91: Is there a need to pass the whole `Topic` and `Subscription` as a parameters?

hermes-management/src/test/groovy/pl/allegro/tech/hermes/management/infrastructure/kafka/service/KafkaConsumerGroupManagerSpec.groovy
```diff
@@ -151,6 +157,68 @@ class KafkaConsumerGroupManagerSpec extends Specification {
         kafkaContainer.dockerClient.unpauseContainerCmd(containerId).exec()
     }
 
+    def "Should delete specified consumer group and retain others"() {
```
szczygiel-m, line 160: Nice. Can we also add some integration test for that mechanism? I mean the one which is creating topic and subscription (with validation of creation of consumer-group), deleting it and then checking whether consumer group was deleted too. What do You think about it?

MarcinBobinski: sure, good idea

moscicky (dismissed):
Thanks for the PR. Please add a short summary in the description on how consumer group removal worked before this change and why the changes was introduced.

MarcinBobinski:
I’ll have to close this PR and redo it because I didn’t realize that to remove a consumer group, all the consumers need to be disconnected first. 
This changes how things need to be handled, so I’ll start fresh with the correct approach.

PR #1922, opened 2024-11-12
SKYEDEN-3234 detect unused topics

questras:
- Add a cron job to detect inactive topics and notify them
- notification thresholds, cron etc are configured via properties
- Notifier implementation is not provided, but the feature can work without it
- the job runs on one instance of hermes management only, based on provided ZK datacenter for leadership election (in properties)


moscicky (approved):
Nice, really clean solution and tests

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/detection/InactiveTopic.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+
+public record InactiveTopic(
+    @JsonProperty String qualifiedTopicName,
+    @JsonProperty long lastPublishedMessageTimestampMs,
+    @JsonProperty List<Long> notificationTimestampsMs,
```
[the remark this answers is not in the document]
questras: It was my first thought but I decided to keep a list as it also tells you how many times you've already notified the topic, which might be useful.
What do you think about providing a configurable limit (via properties) of the list size? this way you still keep the history but it doesn't grow indefinitely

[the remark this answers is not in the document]
moscicky: Sure, we can go with bounded list

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/detection/InactiveTopic.java
```diff
@@ -18,6 +18,19 @@ InactiveTopic notificationSent(Instant timestamp) {
...
         this.lastPublishedMessageTimestampMs,
         newNotificationTimestampsMs,
-        whitelisted);
+        this.whitelisted);
+  }
+
+  InactiveTopic limitNotificationsHistory(int limit) {
+    List<Long> newNotificationTimestampsMs =
+        notificationTimestampsMs.stream()
+            .sorted((a, b) -> Long.compare(b, a))
```
moscicky, line 27: nit: do we really need to sort this list? Timestamps should be order first so we can just take last `limit` elements?

moscicky (approved):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/detection/InactiveTopic.java
questras: I don't want to rely on the current implementation that the timestamps will be sorted. If we ever change anything that changes the order of timestamps then we might get an unexpected behavior and this sorting makes sure we are always safe.

moscicky (approved):

hermes-management/src/main/java/pl/allegro/tech/hermes/management/domain/detection/InactiveTopicsDetectionJob.java
```diff
@@ -0,0 +1,149 @@
+package pl.allegro.tech.hermes.management.domain.detection;
+
+import static java.util.stream.Collectors.groupingBy;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import java.time.Clock;
+import java.time.Instant;
+import java.util.*;
```
szczygiel-m, line 9: nit: * import

szczygiel-m (approved):
Great job!

szczygiel-m (approved):

PR #1942, opened 2025-01-08
Hermes console redesign

Ejden:
This PR introduces a slightly new design in Hermes Console. To be more specific:

* Introduced navigation drawer at the left of the screen that is visible on every page making navigation easier
* (In progress) Changed type of all cards to flat
* (In progress) Introduced new home page design


Ejden:
Split PR into two seperate PRs, so I'm closing this one

PR #1973, opened 2025-08-13
Fix bug related to returning null for createdAt/modifiedAt when anonymizing subscription

faderskd:
When subscription contains oauth policy or endpoint with password it is anonymized before returning it for GUI. The anonymization process didn't set proper createdAt/modifiedAt fields so the GUI displayed "Invalid format" for the fields. This PR fixes this issue.


Diewa (dismissed):

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -171,7 +175,9 @@ public static Subscription createSerialSubscription(
         profilingThresholdMs,
         subscriptionIdentityHeadersEnabled,
         autoDeleteWithTopicEnabled,
-        metricsConfig);
+        metricsConfig,
+        null,
```
pitagoras3, line 179: **question:** As this method `createSerialSubscription` is used only by `public Subscription build()` method, may we add createdAt and modifiedAt params to this method signature, and pass them here explicitly?

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -216,7 +222,9 @@ public static Subscription createBatchSubscription(
         0,
         subscriptionIdentityHeadersEnabled,
         autoDeleteWithTopicEnabled,
-        SubscriptionMetricsConfig.DISABLED);
+        SubscriptionMetricsConfig.DISABLED,
+        null,
```
pitagoras3, line 226: **question:** same as above

hermes-api/src/main/java/pl/allegro/tech/hermes/api/Subscription.java
```diff
@@ -284,7 +292,9 @@ public static Subscription create(
         profilingThresholdMs,
         subscriptionIdentityHeadersEnabled,
         autoDeleteWithTopicEnabled,
-        metricsConfig);
+        metricsConfig,
+        null,
+        null);
   }
 
   @Override
```
pitagoras3, line 300: **question:** Can we add missing fields to equals and hashCode methods?

PR #2035, opened 2026-05-05
Asynchronous sending in RestClientHermesSender

michaldabrowski:
This PR fixes the `RestClientHermesSender` to make it truly async


faderskd (approved):

szczygiel-m (approved):
nice PR! 😄 let me ask one question here, because github is having some trouble
Image

michaldabrowski:
> nice PR! 😄 let me ask one question here, because github is having some trouble Image

For network blocking I/O work seems reasonable to have default like that, usually this kind of pools should be n larger than available cores.

PR #2036, opened 2026-05-15
Include cause details in Avro validation error responses

szczygiel-m:
When publishing a message that fails Avro validation (e.g. missing required field), Hermes returns a generic error like "Invalid message: Failed to convert JSON to Avro" without any details about the root cause.

This happens because the json2avro library wraps AvroRuntimeException into AvroConversionException without including the cause message. This PR fixes it on the Hermes side by extracting the cause details in MessageCreateHandler and appending them to the error response.

Before: {"message":"Invalid message: Failed to convert JSON to Avro","code":"VALIDATION_ERROR"} 

After: {"message":"Invalid message: Failed to convert JSON to Avro: Field age type:ARRAY pos:15 not set and has no default value","code":"VALIDATION_ERROR"}

### DIFF
--- a/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageCreateHandler.java
+++ b/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageCreateHandler.java
@@ -95,7 +95,7 @@ class MessageCreateHandler implements HttpHandler {
   private static String buildDetailedMessage(Exception exception) {
     String message = exception.getMessage();
     Throwable cause = exception.getCause();
-    if (message == null) {
+    if (message == null || message.isEmpty()) {
       return cause != null && cause.getMessage() != null ? cause.getMessage() : "";
     }
     if (cause != null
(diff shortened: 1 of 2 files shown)


hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageCreateHandler.java
```diff
@@ -91,4 +91,19 @@ public void handleRequest(HttpServerExchange exchange) throws Exception {
           exception);
     }
   }
+
+  private static String buildDetailedMessage(Exception exception) {
+    String message = exception.getMessage();
+    Throwable cause = exception.getCause();
+    if (message == null) {
```
michaldabrowski: don't we need `|| message.isEmpty()`

michaldabrowski (dismissed):

hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/publishing/handlers/MessageCreateHandler.java
szczygiel-m: Nice catch! fixed :)

michaldabrowski (approved):

faderskd (approved):

PR #2042, opened 2026-07-08
Convert broker latency to non-histogram timer

faderskd:
This pull request makes a minor adjustment to the way broker latency metrics are recorded. Specifically, it removes the configuration for publishing percentile histograms and the maximum expected value from the metric registration.
This change was introduced due to the huge number of metrics produced for histogram purposes. This will reduce the number of produced broker latency metrics by a factor of ~50x.

### DIFF
--- a/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/BrokerMetrics.java
+++ b/hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/BrokerMetrics.java
@@ -18,8 +18,6 @@ public class BrokerMetrics {
         .tag("broker", broker)
         .tag("broker_dc", brokerDc)
         .tag("ack", ack.name())
-        .publishPercentileHistogram()
-        .maximumExpectedValue(Duration.ofSeconds(5))
         .register(meterRegistry)
         .record(duration);
   }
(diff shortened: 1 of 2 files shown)


hermes-common/src/main/java/pl/allegro/tech/hermes/common/metric/BrokerMetrics.java
```diff
@@ -18,8 +18,6 @@ public void recordBrokerLatency(
         .tag("broker", broker)
         .tag("broker_dc", brokerDc)
         .tag("ack", ack.name())
```
michaldabrowski, line 20: in the docs/docs/deployment/metrics.md:5 we have information that

>Latencies are measured as: 50, 75, 95, 99 and 99.9 percentiles.

is it still up to date?

faderskd: It depends on the micrometer `MeterRegistry` configuration for prometheus. By default we report these 
https://github.com/allegro/hermes/blob/92e41085234ba95ad1938a1249395e8e4c88d62b/hermes-frontend/src/main/java/pl/allegro/tech/hermes/frontend/config/MicrometerRegistryProperties.java#L11

michaldabrowski (approved):

szczygiel-m (approved):

